Issue #5, opened 2017-03-01
Daily reminder


wolkenschieber:
The app looks great and fulfills its task quite perfectly. But I'm having trouble doing continuous measurements. There a configurable daily reminder through notification would be a real enhancement for me.

oliexdev:
Hi wolkenschieber

thank you for your suggestion. If I have a bit more time I will implement in the next version an option to enable/disable a customizable notification. I give a quick look to search for information how to implement this on Android. There is a good developer page under https://developer.android.com/training/scheduling/alarms.html
If you are a developer yourself, I would be happy if you can create a pull request with a first implementation.
Anyhow which text should be displayed in the notification message?

Best regards
Oli

wolkenschieber:
Time is also rare for me, but if I find some I'll give it a go.

> Anyhow which text should be displayed in the notification message?

I'd go with:
*Daily reminder*
*Tab to enter weight*

oliexdev:
I implemented a first weight reminder version, see commit https://github.com/oliexdev/openScale/commit/6eeb5f962f7187d6b55464978d89b5ebf4d27fbd
I made the notification text editable (default text is "Time to weight").
But this version need to be tested. Currently I am not sure if there is still a bug because it was more complicated as I first thought. Maybe somebody can review it.

oliexdev:
I think I found a bug, see commit https://github.com/oliexdev/openScale/commit/1645ce4e989d0f88c690a03a0fc36945a0bed1a3

PR #11, opened 2017-03-05
Interationalize weekday names, Update libraries, some constants

wolkenschieber:
Hi,

I imported the project into Android Studio to test Issue #5. In an emulator the alarms are triggered, as long the app isn't removed from the recent apps list. We'll have to see how this works on a real device.

I did some small refactorings in the refactorings in the reminder code. I'm not use if you really want to integrate this one, because I also changed some *iml* files :-/
But maybe you can take a look pick the parts you like.

Best regards
wolkenschieber

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,8 @@ proguard/
 
 # Log Files
 *.log
+
+# Android Studio
+*.iml
+/.idea/workspace.xml
+/.idea/libraries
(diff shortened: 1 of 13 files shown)


oliexdev:
Hi wolkenschieber,

thank you for your pull request, looks great. 
I merged everything into the master branch (only comment out the debug info).
It looks like that the _iml_ files can be ignored. I will remove them from the repository.

Best regards
oli

Issue #10, opened 2017-03-05
GUI changes in settings are not instantly applied


oliexdev:
Currently GUI changes in settings are not instantly applied. You have to restart the app or switch several time between the three sub fragments to update a GUI fragment.

nastynabz:
Hello,
I want to try to fix this issue.
I have an idea to fix it.

oliexdev:
Hi nastynabz,

that would be great. Please create a pull request with your code changes.

Best regards
oli

oliexdev:
thanks to nastynabz the bug is fixed, see pull request #19.

PR #19, opened 2017-03-28
Gui changes

nastynabz:
When you modify settings, the fragment are now refreshed when you press "Back".


oliexdev:
Hi nastynabz,

thanks for your code contribution, with your help we can close issue #10 now. 
I modified your code a bit so that is not necessary to have a public static member. I recreated the fragments on the activity result.

Best regards
Oli

PR #23, opened 2017-04-06
Alarm rescheduling and notification cancelation

wolkenschieber:
Hi,

within this pull request the alarm scheduling was reworked. 
With this refactorings alarms are now rescheduled for the current day, if data is entered before the alarm timestamp. Furthermore the reminder notification is removed, if data is entered while it's still present.

Best regards
wolkenschieber


oliexdev:
Hi wolkenschieber,

thanks for your awesome code refactoring and further more for the alarm feature improvement! :+1: 
I just refactored your code a bit so that we don't need an extra database listener and util class.
Unfortunately I didn't commit my changes of your code before I restructured the source code files into sub packages, so my commit https://github.com/oliexdev/openScale/commit/3003b0a75ffd2a6f12bcd948fc6af9513ba53832 is a little bit confusing

Best regards
oli

wolkenschieber:
You are very welcome.

If  you decide to do further actions on table entry changes, you can always put such listeners back in.

PR #32, opened 2017-05-02
Add support for Sanitas SBF-70 scale

jflesch:
Support for:
- weight
- fat
- water
- muscle

No history support.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/OpenScale.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/OpenScale.java
@@ -27,6 +27,7 @@ import com.health.openscale.core.alarm.AlarmHandler;
 import com.health.openscale.core.bluetooth.BluetoothCommunication;
 import com.health.openscale.core.bluetooth.BluetoothCustomOpenScale;
 import com.health.openscale.core.bluetooth.BluetoothMiScale;
+import com.health.openscale.core.bluetooth.BluetoothSanitasSbf70;
 import com.health.openscale.core.database.ScaleDatabase;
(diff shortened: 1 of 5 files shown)


oliexdev:
Hi jflesch,

thanks for your awesome PR :+1:  I merged it into the master tree.
Does the Sanitas SBF-70 scale support a history function in general?

Best regards
oli

jflesch:
Kinda. It can memorize some measures (about 10 I would guess), and the application can fetch them later.

oliexdev:
ok interesting maybe the history function is the same like the [Xiaomi Mi Scale](https://github.com/oliexdev/openScale/wiki/Xiaomi-Bluetooth-Mi-Scale). The Custom Characteristic   `0000FFE5-0000-1000-8000-00805F9B34FB` of the Sanitas SBF-70 scale seems interesting for the history function because it has also an indication flag and you can write a value in. But I can't test it because I don't own that scale ;)

jflesch:
I'll try to have a look at some point (I'm currently seriously lacking free time, so I can't say when exactly).
BTW, I forgot to thank you too for your wiki page : https://github.com/oliexdev/openScale/wiki/Xiaomi-Bluetooth-Mi-Scale . It was really helpful to reverse-engineer the Sanitas SBF-70 :)

jflesch:
Also, I forgot to specify one thing:
There is a limitation with my code. The scale has to be already initialized and configured using the proprietary app provided :/

oliexdev:
no problem it's great that we have another scale reversed engineered.
Maybe I place somewhere an information about the limitations and functions of the scales otherwise some users will confused why it isn't working.

PR #34, opened 2017-05-03
Graphics: reverse top chart and bottom chart

jflesch:
Graphics: reverse top chart and bottom chart. This way, from top to bottom, char are years->months->days instead of years->days->months.

It makes it also clearer that you can press on the months' graphs to change the days' graph.


oliexdev:
thanks I merged it into the master tree but I will place the enable/disable month button on bottom right of the screen.

PR #53, opened 2017-07-01
Add new HW address for Medisana BS444; MagicNumber is current time, fixes wrong timestamp;

mapero:
Add another HW address for Medisana Smartscale


oliexdev:
@mapero thank you for your help

Issue #4, opened 2017-03-01
feature request. more measurements


nutpantz:
Just found this on fdroid. Looks great.
Only thing I can see it I might need is more measurements.
Not that these measurements need to be tracked, just compared to the last or first measurements.
Measurements like
Biceps, forearms, chest, neck, thigh and calf maybe even left and right sides.
Great for weight loss to see what is going missing
And good for body sculpting to see where muscle is being built.

Great work.

oliexdev:
Hi nutzpantz

thanks for your suggestion. I am not quiet sure how is the best way to implement more measurements, so that the app looks not overloaded and confusing. Do you have any design ideas?
The next question is how many measurements are needed and which? Where to start and where to end? For example should the neck circumference included or not. Some people will need it, but most of them not. An idea is to add measurement fields dynamically where the user can freely give them a title, a unit and a valid range. But this is not easy to implement and again the app should be easy to use. Maybe you or somebody else have another idea?
I am sorry at the moment the best way is to use the comment field to note any further measurements. 

Best regards
Oli

nutpantz:
Hello
Personally I like a button that switches to a page just about measurements.
One side would show the body and approximately where to measure for each
It should show the last measurement taken next to the new measurement input spot
You could have the body in the middle and measurements on each side for left and right.
Maybe one or two on the bottom for custom entries.
Maybe a show/hide list in settings for people who just want to enter some measurements and not clutter the screen unnecessarily or remind them that they are not measuring everything.

As for what measurements are needed that's a personal thing and more for the developer as to who
He is aiming his app at.

Weight watchers benefit from the primary ones
(Thigh, waist, hips, chest, neck, biceps) as they change a lot and can encourage people when they can't see just the waist changing.

Body builders would need more 
(Calf, forearm, maybe wrist and ankles(I'm not sure myself))
The more options the more out will appeal to a broader range of users

Larry0ua:
Maybe you just need some custom user-provided field names, it would fit nicely into Measurements settings section

oliexdev:
@Larry0ua 
as I already mention above this would be a solution but also a complex one. Everything graph, table, export/import function, and the database itself and so on need to be dynamically programmable.
I make a small step in this direction because now the measurement views can be dynamically added... I guess this will take more time. 

Best regards
oli

martinler:
Hello, any idea about integration with Google Fit? Beurer stock app does not provide this feature, so i'm looking for some replacement. Thanks!

[...]

PR #57, opened 2017-07-25
Add absolute body fat and muscle

MilenaMMay:
Add absolute body fat and muscle on DataEntry, Overview and Table so that the users can also monitor whether they lose or gain fat or muscle when their weight changes.

@oliexdev I understand from this issue https://github.com/oliexdev/openScale/issues/4 that you wouldn't like to clutter the app with too much measurements. I generally agree but I still think that absolute values of muscles and fat would be of value for the user (at least for me).

I made them configurable so the user can disable this feature. For discoverability I initially turned them on.


MilenaMMay:
There's one

MilenaMMay:
Closing this so you can merge the pull request from my branch https://github.com/oliexdev/openScale/pull/59

PR #59, opened 2017-07-25
Add absolute body fat and muscle

MilenaMMay:
Add absolute body fat and muscle on DataEntry, Overview, Table and Chart so that the users can also monitor whether they lose or gain fat or muscle when their weight changes.

@oliexdev I understand from this issue #4 that you wouldn't like to clutter the app with too much measurements. I generally agree but I still think that absolute values of muscles and fat would be of value for the user (at least for me).

I made them configurable so the user can disable this feature. For discoverability I initially turned them on.


oliexdev:
Thanks milena for your pull request but somehow I don't like the idea to add an extra field for absolute fat and muscle. We have already a measurement for fat and muscle. The only think that is changes is the displayed value like the unit value... so it would be better to change only the displaying value instead of adding a new measurement. But currently I am not quite sure what is the best way to do it (make a extra option or display both values at the same time?!)
I want to show the non-editable measurements in the DataEntryActivity because you see all values in detail (and hopefully better) if you click on one measurement.
The refactoring of the GraphFragment is a good step forward but the goal is someday to have just the measurement view classes that have to be implement and all other classes like GraphFragment, TableFragment grab all information that they needed from the measruement view classes so you don't have to touch these classes anymore. Maybe in this way we can dynamically easily add new measurements by a user.
Anyway I have to think about it and thanks again for your effort.

MilenaMMay:
Hi Oli,

1) Regarding the non-editable measurements in the DataEntryActivity: ok, I will revert that.

2) Regarding the refactoring of the GraphFragment: if that's a step toward more flexibility, I can extract this commit in a separate pull request.

3) Regarding the absolute fat and muscle feature: please let me explain why I think it is useful.

**Example 1: someone who wants to lose body fat:**
Data 1: 100 kilo weight, 30% muscle, 30% fat => means 30 kilo muscle, 30 kilo fat.
Data 2: 80 kilo weight, 35% muscle, 30% fat => means 28 kilo muscle, 24 kilo fat.
=> even if that persons relative body fat did not change, 6 kilo of fat were lost, but that person could not monitor that goal. By the relative numbers, it looks as if that person gained muscle, but the absolute muscle mass did in fact decrease by 2 kilo.

**Example 2: someone who wants to gain muscle:**
Data 1: 80 kilo weight, 30% muscle, 30% fat => means 24 kilo muscle, 24 kilo fat.
Data 2: 100 kilo weight, 35% muscle, 30% fat => means 35 kilo muscle, 30 kilo fat.
=> that person gained muscle BUT even if that persons relative body fat did not change, 6 kilo of fat were gained additionally, which was not the goal.

Being one of these examples, I currently track my absolute numbers in a separate sheet and would love to see it included in openScale. What if the absolute values were initially disabled?

Cheers,
Milena

oliexdev:
you don't have to create an extra PR because I can do ´cherry-pick´ commit.

I understand the problem but what if we add an option to calculate the absolute values or the percentage? But still it is not a perfect solution...

DestyNova:
Since the absolute/% values are of course related, why not add both to the form, either via a drop-down or radio button, or simply include two fields and trigger an update of one when the other changes?

As it is, it's hard for me to use the app because my scale reports body fat in % but muscle mass in kg. So I have to open a calculator and convert one into %, remembering the values after the scale turns off.

woj-tek:
@DestyNova - I have the same problem with Mi Scale - it shows muscle mass in kg which is weird.

Maybe there could be a toggle in the preferences for each field and it could be selected whether you want to input [%] or [kg] and then in the entry form you would have input allowing you to enter one of the types of measurements (percent or kilograms) and the calculated entry would be displayed below (non editable, greyed out).

Alternatively both could be displayed/linked and then changing one would cause recalculating of the other, but IMHO if the scale gives the measurements in single value then having item in preferences would be better (and we could avoid cluttering data entry form)

oliexdev:
I implemented an option to select if the body fat, body water or muscle are displayed in % or in their selected user unit with commit https://github.com/oliexdev/openScale/commit/7fce43e8eab36ac2d4583cfcabcffbfa204d2cbb

Issue #69, opened 2017-09-21
Linear regression for measurements


beothorn:
When trying to achieve a goal, it is nice to see to which direction the measurements are leading to, specially the weight.
I've used other weight measurements software that showed me how when I would reach my goal if I kept my current pace and it was really usefull.
Maybe as an option?
Is that something you would find usefull?

oliexdev:
Hi beothorn,

I think as an option it would be useful to have a linear regression line for the weight but for all measurements I think that would be too much...   Additionally I can add an optional "goal" line. 

Best regards
oli

oliexdev:
I implemented a goal line and a linear regression line for weight values with commit https://github.com/oliexdev/openScale/commit/8761ef5f0e0f74183567d47c1066df6cff0a4df0

But I am not really sure anymore if it really helps the linear regression line because depending on the data it is only rough outline of your predicted weight process!?
I tried also a quadratic regression line but is that more useful? Anyway I implemented as an option so anybody can enable or disable the two lines in the graph.

oliexdev:
I added an option to set the polynomical order for the regression weight line.

beothorn:
Wow, thanks, that was fast :)

I'm using it to lose weight and I feel more motivated when I see my weight loss trend compared to my goal. 
It is not the best model to predict weight loss but because my goal is set to next year, it helps me to know if I am losing too much weight (I don't think it is healthy to lose too much in a too short time) or too little.

Quadratic is really nice, but at least for me I will keep Regression order at 1.

I saw that the goal line wasn't rendering because:
```
valuesGoalLine.push(new PointValue(0, goalWeight));
valuesGoalLine.push(new PointValue(31, goalWeight));
```
I've added some code at https://github.com/beothorn/openScale/commit/089a871375000e39ef9cfd014fffd1585ad20fe6
just so the goal line shows up.

It seems that the goal line and the prediction line aren't yet correct for the monthly view. I'll take a look.

Thanks for the awesome app

beothorn:
Ops, prediction line is correct (I've added entries only for one month) , just my code for the goal line it's not.

beothorn:
https://github.com/oliexdev/openScale/pull/70

oliexdev:
The goal line was displayed but I implemented it in a way that is not automatically zoomed out to the goal line because it is possible that the goal weight and the weight values have a big gap. In these situation the x-axis would be to big. So please zoom out the diagram to see the goal line.

[...]

PR #70, opened 2017-09-23
Using goal weight to draw goal line

beothorn:
Showing red goal line going from last measurement to goal date.


oliexdev:
thanks for your PR but I will decline it and keep my implementation, see issue #69.

PR #72, opened 2017-09-23
Brazilian portuguese

beothorn:
Brazilian portuguese localized strings.


oliexdev:
thanks for your translation @beothorn :smile:

PR #73, opened 2017-09-25
Add support for Digoo DG-S038H

Murgeye:


oliexdev:
thanks @Murgeye for your contribution. :+1: 
Is it possible to take a photo of the scale in the same angle like in the [openScale wiki](https://github.com/oliexdev/openScale/wiki), so I can use it there? I don't want to use a public image of the scale due to copyright reasons.

Murgeye:
Would this picture work for you?

digoo

I would have added the measurements myself, but I wasn't quite sure where to add them as I am horrible at designing UIs and wouldn't want to mess with that :smile:

oliexdev:
thanks I added the image to the wiki.

well to implement a new measurement view is quite easy, just extend the MeasurementView class. Unfortunately afterwards  you have to touch for the diagram and the import/export function and storing into the database each class :/
Its a old issue #4 but its complicated to implement it so it more convenient to add a new measurement.

PR #76, opened 2017-09-30
Added slovak translation

MatoDroid:


oliexdev:
thanks for your translation @MatoDroid

PR #78, opened 2017-10-08
Create strings.xml for Polish language

roqeer:
GUI polish translation


oliexdev:
thanks @roqeer for your Polish translation.

Issue #71, opened 2017-09-23
Reverse engineering Yunmai mini scale


oliexdev:
I am trying to reverse engineer the yunmai mini scale.

I already found out the weight values but I can't figure out the body fat percentage, muscle percentage and so on. Any help or idea is welcome.

So far what I got:

- No weight values are stored on the scale itself. You have to be connected to the yunmai app to receive any data
- Weights are continuous send to UUID: Unknown (0xffe4) (if you stay on the scale)
- Sent values are like `0d:1f:0b:01:59:c6:15:2f:26:95:03`
  * byte 0-6: bytes are constant
  * byte 7: first nibble is constant, second nibble is changing
  * byte 8-9: contains the weight value (here `0x2695` = 9877 = 98,77 btw the value is rounded in the yunmai app as 98,8 kg)
  * byte 10: is changing
- last measurement from the scale of the continuous sent send data looks like `0d:1f:14:02:00:59:c6:15:2f:23:d0:b6:32:26:95:00:a5:08:f2:37`
  * byte 0-5: bytes are constant
  * byte 6: first nibble is constant, second nibble is changing
  * byte 7-8: bytes are changing
  * byte 9-12: bytes are constant
  * byte 13-14: contains the weight value again
  * byte 15: bytes are constant
  * byte 16-19: bytes are changing
- after that the yunmai app send some values to the scale. In response they send some values back to the scale like the following `0d:1f:12:06:01:23:d0:b6:32:af:01:19:55:5a:8d:26:01:6f`
-  I guess that the values displayed in the yunmai app are rounded values like as the weight value
- Since so many bytes are constant and only few bytes are changing it could be possible that the information of the body measurements (like fat, muscle, water percentage etc.) are compressed or somehow encrypted. (for example a fat percentage of 22,9 % is somehow in the above values encoded)

Here are my Bluetooth log files with the corresponding body measurements which we have to find (open it with wireshark):

**23.9.2017 07:36**
[btsnoop_hci_979.log](https://github.com/oliexdev/openScale/files/1326417/btsnoop_hci_979.log)
* Weight 97,9 kg
* Body fat 22,9 %
* Muscle 51,7 % 
* Water 56,0 % 
* Protein 16,2 %
* Bone mass 4,0 %

**23.9.2017 10:02** 
[btsnoop_hci_988.log](https://github.com/oliexdev/openScale/files/1326418/btsnoop_hci_988.log)
* Weight 98,8 kg
* Body fat 22,9%
[...]

PR #87, opened 2017-10-26
Initial working support for Mi Scale v2

ashimokawa:
Notes:
- This is basically a copy of BluetoothMiScale.java with some data offset adjustments
- The old realtime measurement characteristic is absent
- We only get impedance measurements, no direct body fat calculation is done on the firmware
- ~~The scale is not yet switched to kg, it is still operated in catty and there is no HW switch!~~ **DONE**


ashimokawa:
Sorry, I think this is not ready to merge - or needs at least more testing.

I had trouble getting data today - after tinkering with Gadgetbrige and then connecting with openScale it started to work again - think I might need to send user info (height, gender), will do further tests.

I suggest to leave this open, as I might add more commits.

oliexdev:
@ashimokawa thanks for your effort. After a long time I received my mi scale v2. After a first glance I see that only the weight and the impedance are send via Bluetooth. The calculation of water percentage muscle percentage and so on is done by the mi fit app or server, like the yunmai mini see issue #71.

oliexdev:
I updated source code so that is first received the history data and then weight on the scale is received.
Also on each time the scale is initialized with the time and the scale unit.
the "realtime" measurement is now send to `0x2a9c` UUID.

ashimokawa:
@oliexdev 
Cool, with check that out tomorrow.

PR #88, opened 2017-10-31
Target api 26

wolkenschieber:
This pull request aims to leverage api level to 26. To do so it mainly introduces runtime permissions. The permissions granting process is delegated to the library dexter.

The changes also include a gradle version upgrade. This means that Android Studio 3 is required to work on the application!

Please decide whether you want to accept the pull request. Also some additional testing won't harm ;-)


oliexdev:
thanks for your PR but what are currently the benefits of using a higher api level?
Currently I don't use any higher api and to increase the level I get some drawbacks like to need a permission for coarse location and reducing the number of supportable android versions.
But maybe I don't understand the android api schema. :/

wolkenschieber:
The android api schema is easy: An new android os version introduces an new api version.

The api is subject of change. Features are being introduced and bugs are fixed. Not all bugfixes are backported into the older support libraries. So in oder to benefit from those, a switch to the new api is necessary.

However an new version may require changes to the code. Let's take the runtime permissions. Since Marshmallow (API 23) dangerous permissions must be requested at runtime. Declaring those in the manifest doesn't do it anymore. 

But since then the user has more control over his data. A flashlight app may request contact permissions, but it can be denied.

The coarse location permission was also added for [data protection reasons](https://developer.android.com/about/versions/marshmallow/android-6.0-changes.html#behavior-hardware-id).

The minimum api level remains at 18. So there will be as much supported devices as before.

In the end it's up to you, whether you want to switch or not. Some things might break, some things might be different. If there's no need for new android features, the update is not mandatory.

oliexdev:
thanks @wolkenschieber for your explanation.
I think I will update it if I need a special feature that require a higher api level. I see often comments in another apps why the coarse location is needed for a Bluetooth connection. Most people are angry about it because they don't know the reason.
Therefore I will keep this PR open until I need it but I am grateful for your PR @wolkenschieber. :+1:

wolkenschieber:
You're very welcome. Thanks for sharing this app!

oliexdev:
I set the target api to 27 and implemented with commit https://github.com/oliexdev/openScale/commit/97a75762dcc5fd36bc35a5159f5bc58a973a1d86 a permission handler.

But thanks again for your PR @wolkenschieber

PR #95, opened 2017-11-09
Unify date and time handling

erijo:
* Present date in the "medium" (default) format (e.g. Nov 9, 2017)
  instead of dd.MM.YYYY (e.g. 09.11.2017). Besides that it is easier
  to read and wraps better in the table view, it is also localized and
  avoids the uncertainty if day or month comes first.

* Similar, time is now presented in a localized format (i.e. 12 or 24
  hour clock is taken from system settings).

* The time pickers are also configured to show a 12 or 24 hour clock
  depending on the system setting.

* Handling of birthday and goal date is now fully done in the user
  settings activity and OpenScale.(add|update)ScaleUser no longer has
  to know which format the date string was in.

* The date and time measurement view now opens the date/time picker
  with the initial date/time set to the measurement's value.


oliexdev:
@erijo thanks again for improving openScale :+1:

PR #100, opened 2017-11-20
Upgrade gradle and use default build tools version

erijo:
As suggested by Android Studio 3.0.1.


oliexdev:
thanks @erijo I updated my Android Studio also to version 3.0.1

PR #101, opened 2017-11-20
Make the linear gauge easier to read

erijo:
By drawing the value text under the linear gauge it doesn't collide with the limit text and is easier to read. Also made the min and max value dynamic to put the normal zone in the middle and give a better resolution for the values that the user cares about.

Also some centering of text and minor fixes.


oliexdev:
thanks @erijo for your improvement and bug fixes. :+1:

erijo:
Thank you @oliexdev for making such a great program open source!

FYI I have a few more changes queued in this area. Will create a pull request when I get some time.

I also noticed that there are mixed tabs/spaces in some source files. Would you accept a pull request to convert tabs to spaces or do you consider it mostly noise?

oliexdev:
sure but please create for the mixed tabs/spaces clean up an extra PR. I am looking forward for your next PR :smiley:

PR #105, opened 2017-11-26
Hide user selector when there is only one user

erijo:


oliexdev:
Is the user selector really disturbing?

erijo:
Perhaps not disturbing, but for someone like me with only one user it adds no value and takes up precious vertical space that can be used to show more data without having to scroll.

oliexdev:
ok convinced :wink:

PR #106, opened 2017-11-26
Update diff value and dependent values when measurement change

erijo:
If e.g. the weight is changed, also update the BMI value.

Also some minor bug fixes to not show evaluator row when there is no value to evaluate, and to make it possible to click the evaluator row to hide it.


oliexdev:
thanks for your patch @erijo 
Only the "don't show evaluator row when there is no value to evaluate" maybe I will revert it later because even if no value are evaluated, the user is at least informed that this is not possible.

erijo:
Perhaps a toast (if that is what they are called) can be shown when the user clicks a row that can't be evaluated? To at least give some feedback.

oliexdev:
yeah would be also an idea.

PR #107, opened 2017-11-28
Add swedish translation

erijo:


oliexdev:
@erijo thanks for your Swedish translation :+1:

PR #112, opened 2017-12-03
Add support for non-binary gender

sdukhovni:
The `ScaleUser.gender` variable can now be set to 2 to indicate a non-binary gender, and a corresponding radio button has been added to the UI.  For evaluation sheets, we evaluate a non-binary user as `NORMAL` if their body metrics are within the combined `NORMAL` ranges for male and female metrics.  For body metrics that depend on biological sex, we return the average of the male and female values for a non-binary user.  For communicating with bluetooth scales, we largely stick to the existing code pattern, designating all non-male users as female.

This is kind of terrible, but I don't have any better ideas for how to make this app as gender-inclusive as possible while relying on data from non-inclusive research papers and interfacing with non-inclusive external devices.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothSanitasSbf70.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothSanitasSbf70.java
@@ -184,7 +184,7 @@ public class BluetoothSanitasSbf70 extends BluetoothCommunication {
                             (byte) selectedUser.birthday.getMonth(),
                             (byte) selectedUser.birthday.getDate(),
                             (byte) selectedUser.body_height,
-                            (byte) (((1 - selectedUser.gender) << 7) | activity)
+                            (byte) (((selectedUser.isMale() ? 1 : 0) << 7) | activity)
                     });
                 } else {
                     // Get existing user information
--- a/android_app/app/src/main/java/com/health/openscale/core/bodymetric/BFDeurenberg.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/bodymetric/BFDeurenberg.java
@@ -26,10 +26,22 @@ public class BFDeurenberg extends EstimatedFatMetric {
 
     @Override
     public float getFat(ScaleUser user, ScaleData data) {
+        float genderVal = 0.0f;
+        switch (user.gender) {
+            case ScaleUser.MALE:
+                genderVal = 1.0f;
+                break;
+            case ScaleUser.FEMALE:
+                genderVal = 0.0f;
+                break;
+            default:
+                genderVal = 0.5;
+                break;
+        }
         if (user.getAge(data.getDateTime()) >= 16) {
-            return (1.2f * data.getBMI(user.body_height)) + (0.23f*user.getAge(data.getDateTime())) - (10.8f*(1-user.gender)) - 5.4f;
+            return (1.2f * data.getBMI(user.body_height)) + (0.23f*user.getAge(data.getDateTime())) - (10.8f*genderVal) - 5.4f;
         }
 
-        return (1.294f * data.getBMI(user.body_height)) + (0.20f*user.getAge(data.getDateTime())) - (11.4f*(1-user.gender)) - 8.0f;
+        return (1.294f * data.getBMI(user.body_height)) + (0.20f*user.getAge(data.getDateTime())) - (11.4f*genderVal) - 8.0f;
     }
 }
--- a/android_app/app/src/main/java/com/health/openscale/core/bodymetric/BFDeurenbergII.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/bodymetric/BFDeurenbergII.java
@@ -26,10 +26,9 @@ public class BFDeurenbergII extends EstimatedFatMetric {
 
     @Override
     public float getFat(ScaleUser user, ScaleData data) {
-        if (user.isMale()) {
-            return (data.getBMI(user.body_height) * 1.2f) + (user.getAge(data.getDateTime()) * 0.23f) - 16.2f;
-        }
+        float maleFat = (data.getBMI(user.body_height) * 1.2f) + (user.getAge(data.getDateTime()) * 0.23f) - 16.2f;
+        float femaleFat = (data.getBMI(user.body_height) * 1.2f) + (user.getAge(data.getDateTime()) * 0.23f) - 5.4f;
 
-        return (data.getBMI(user.body_height) * 1.2f) + (user.getAge(data.getDateTime()) * 0.23f) - 5.4f;
+        return Utils.genderizeMetric(user, maleFat, femaleFat);
     }
 }
--- a/android_app/app/src/main/java/com/health/openscale/core/bodymetric/BFEddy.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/bodymetric/BFEddy.java
@@ -26,10 +26,9 @@ public class BFEddy extends EstimatedFatMetric {
 
     @Override
     public float getFat(ScaleUser user, ScaleData data) {
-        if (user.isMale()) {
-            return (1.281f* data.getBMI(user.body_height)) - 10.13f;
-        }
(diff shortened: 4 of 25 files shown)


oliexdev:
Hi @sdukhovni 

thanks for your PR but I think its dangerous to average the values between male and female for non-binary genders. You would create expectations which are not based on valid researched foundation. People would progress a diet or gain weight to fit into the wrong evaluation results. I am not a medical doctor but I would assume that even if they identify as non-binary it has no impact into the measuring if they use their biological gender (or they switch the gender in the app if they biological gender becomes "more" the opposite gender). I might be wrong but that should be medical research topic.

sdukhovni:
(Sorry, this comment got kind of long)

Thanks for your feedback. I agree with you that I probably didn't think things through very well with regard to the body metrics, and I definitely want users to not make important health decisions based on questionable data.

I'd like to step back and discuss what my goal is with this PR, and we can figure out the best way to proceed.

- For users whose gender doesn't match their biological sex (see footnote), requiring them to input their biological sex is unpleasant and uncomfortable, and should be avoided unless the app is completely unusable without this data.

- For users who are taking crossgender hormones, intersex users, and others, their physiologies will not necessarily align perfectly with "typical" physiologies for cisgender men or women, so in the absence of specific medical research on body metrics of people other than cisgender men and women, I don't think these formulas can ever be trusted to give reliable data, no matter what binary sex is plugged into the formulas. More generally, once we look beyond cisgender men and women, there's enough physiological diversity that there may not exist a small set of formulas that provide useful approximations for body fat, water, etc. for all cases.

- Given those points, the natural conclusion seems to be that we shouldn't display these metrics at all for users who aren't cisgender men or women. I didn't do that with my PR because it felt exclusionary in unfortunate ways, and providing ballpark estimates seemed like a reasonable option. However, if in your judgment it's better to provide no data rather than inaccurate data, I'm ok with that solution.

So, my new proposal: The user creation UI should have a set of radio buttons labelled

Sex:
- Male
- Female
- Intersex/Other/Prefer not to say

and if the third option is chosen, the body metrics view is replaced with a message along the lines of "we're sorry, but due to a lack of available medical research, we're unable to calculate body metrics for the sex you have selected." For these users, the app would still be usable as a weight tracker that lets you log your weight, display graphs of your weight over time, and connect with bluetooth scales; this completely covers my personal intended use case for the app, and I don't feel I really need the body metrics.

Would you be willing to accept a pull request to implement that?


### Footnote (optional reading):

I didn't want to launch into a discussion of gender theory in the middle of describing proposed features, but I think this is worth considering: This app currently asks about "gender" in the UI and uses "gender" and "sex" fairly interchangeably for variable names in the code; while I don't intend to push hard on this if you'd rather not deal with it, I'd like to clarify the distinction a bit better.

"Gender" refers to how a person conceives of themself and presents themself to others, and is a social/cultural characteristic that doesn't necessarily have any connection to biology.

"Sex" or "biological sex" is a somewhat ill-defined term that can refer to any of chromosomes, genital configuration, and/or physiology, even though all three of these can vary independently given various medical treatments or medical conditions. Some people, known as "intersex", are born with bodies that do not fit within typical definitions for "male" or "female", and may have ambiguous genitals and/or chromosomal configurations other than XX or XY.

It seems like the body metrics in this app depend on sex rather than gender, but with regard to the evaluation sheets, gender may be more important for users who are trying to attain a physiology typical for a sex other than their assigned-at-birth sex. However, if we're deciding to not provide evaluations for such users, then it seems reasonable to only ask users their sex and not worry about gender. With that in mind, I'd like to replace all references to "gender" in the codebase with "sex" if that's something you're ok with; again, this is a lower priority for me than just not requiring users to select a binary sex, so if you'd rather not make that code change I can leave it out.

oliexdev:
> For users whose gender doesn't match their biological sex (see footnote), requiring them to input their biological sex is unpleasant and uncomfortable, and should be avoided unless the app is completely unusable without this data.

So your goal is to have an option which are not female and male because it feels you uncomfortable and unpleasant to select one of these two? Honestly I never thought about that problem but I can understand that in a way.

> For users who are taking crossgender hormones, intersex users, and others, their physiologies will not necessarily align perfectly with "typical" physiologies for cisgender men or women, so in the absence of specific medical research on body metrics of people other than cisgender men and women, I don't think these formulas can ever be trusted to give reliable data, no matter what binary sex is plugged into the formulas. More generally, once we look beyond cisgender men and women, there's enough physiological diversity that there may not exist a small set of formulas that provide useful approximations for body fat, water, etc. for all cases.

Well its a complex topic not only for this topic its goes even further: As I know all scale companies the calculation of their body metrics are not transparent and are not based on any published scientific publication. Even most body metrics are mathematical derived from the fat and weight and so on. I tried at least to make the [body metric estimations](https://github.com/oliexdev/openScale/wiki/Body-metric-estimations) transparent as much as I can. But if the calculation is done on the scale itself I really don't know how the calculation is done for the body metrics. So in the end we don't really know, even for cisgender, if the calculation are based on reliable data. Furthermore if we think bigger I found formulas in public research papers that have customized formulas for ethics groups (e.g. asian vs. non-asian formulas) or for disabled people (e.g. who lost a leg). And we only speaking of the calculation of the body metrics not the evaluation of the results. As you see if you want to support all of the variations it becomes quite complex. Furthermore the [Bioelectrical impedance analysis](https://en.wikipedia.org/wiki/Bioelectrical_impedance_analysis) that are typical used in home scales are not very precise. 

> Given those points, the natural conclusion seems to be that we shouldn't display these metrics at all for users who aren't cisgender men or women. I didn't do that with my PR because it felt exclusionary in unfortunate ways, and providing ballpark estimates seemed like a reasonable option. However, if in your judgment it's better to provide no data rather than inaccurate data, I'm ok with that solution.

Well at least it should be transparent for the user. We can't really provide no data because some Bluetooth scales need to specify a sex at the initialisation process otherwise the scale won't work.
So yeah we could not evaluate any data (btw. which I already do if you are too young or too old).

> and if the third option is chosen, the body metrics view is replaced with a message along the lines of "we're sorry, but due to a lack of available medical research, we're unable to calculate body metrics for the sex you have selected." For these users, the app would still be usable as a weight tracker that lets you log your weight, display graphs of your weight over time, and connect with bluetooth scales; this completely covers my personal intended use case for the app, and I don't feel I really need the body metrics.

To be honest I don't really want to make a huge code change or to be more precise the openScale design should be simple and light and the source code should be clear and maintainable. So my idea would be if we add a third option "non-binary" or "other". If we select this sex we create a popup message with an information with one of these two variants:
a ) information that we don't evaluate any body metric for reasons xyz but we set your sex to male for necessary settings. (too disable the evaluation we could set internally their age to 0)
b) information that we automatically set your sex internally to male for reasons xyz and evaluate all body metrics as male but these are not accurate
So we don't have a lot of code change but it would be transparent but I don't know which variant would be better!? What do you think of this idea?

> I didn't want to launch into a discussion of gender theory in the middle of describing proposed features, but I think this is worth considering: This app currently asks about "gender" in the UI and uses "gender" and "sex" fairly interchangeably for variable names in the code; while I don't intend to push hard on this if you'd rather not deal with it, I'd like to clarify the distinction a bit better.

Ok I was not aware of these subtle differences. Well if you think the word sex perfects better for us then I am ok with that.

sdukhovni:
Thanks for being responsive to my concerns, I really appreciate it :)

I think not providing metrics/evaluations at all is better than providing male metrics/evaluations. I also think that internally representing non-cisgender-male-or-female users as "sex: other" is cleaner and easier for future maintainers to understand than representing them as "sex: male, age: 0"; if we aren't providing metrics then the required code change gets a lot smaller. I'd get rid of my edits to the body metrics, and the only edit to the evaluation sheets would be to always return `UNDEFINED` for "sex:other" users. We'd mostly just be changing ScaleUser.java, fixing the small number of places in the bluetooth communication code that assume `ScaleUser.gender` is always 0 or 1, and adding a couple of new messages to the GUI. I'll push that code onto this PR sometime tonight or tomorrow and you can decide whether it seems reasonable.

oliexdev:
I will close this PR. Please open a new PR for the another solution.

Issue #111, opened 2017-12-02
Need Help: reverse engineering Beurer BF 700 bt protocol


linuxlurak:
Hello! I have a Beurer BF 700 and the bulky "HealthManager App" from Beurer for Android. I decoded the Android App in a few steps and got the source code. But I don't know what to look for. 

I found the fcc docs from beurer here: https://fccid.io/VGR-BF700.

Publishing the app's source code would be an infringement; it is closed source. Please contact me privately for more info about the source code.

Have a peek in the files here:
![image](https://user-images.githubusercontent.com/3813355/33514017-aced46e4-d74c-11e7-8ea4-9bcf106fb66c.png)

and here:
![image](https://user-images.githubusercontent.com/3813355/33514036-f7e8f742-d74c-11e7-90a8-718bac4cd019.png)

The BF700 device is generically handled as a BF800.

oliexdev:
The FCC documents are not relevant. Decompiled source code is hard to understand.  Please read the [wiki page](https://github.com/oliexdev/openScale/wiki/How-to-reverse-engineer-a-Bluetooth-4.x-scale) for how to reverse engineer the Bluetooth protocol. Btw I can't contact you because you don't have any email address published on your github page.

Dododappere:
The Beurer BF700/800 scales use basically the same protocol as the Sanitas SBF70. Sanitas is actually a sub-brand of Beurer...

Where Sanitas commands start with 0xe7, Beurer commands start with 0xf7. The byte that follows after that is usually the same for both the Sanitas and Beurer scales, apart from some exceptions.

linuxlurak:
I assumed this too! How did you get proof of it? How do you know the start command? It compared detail photos of the devices, but had no access to the internal electronic or even a firmware dump...

I'll check it with the start command, thanks for letting us know!

linuxlurak:
Some days before I forked openscale and created a Beurer BF700 clone based on the sanitas here: https://github.com/linuxlurak/openScale/blob/master/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothBeurerBF700.java

I'll compile the app and post the results here if it works with your starting byte.

linuxlurak:
connecting to the beurer bf700 works according to the app:
![image](https://user-images.githubusercontent.com/3813355/34035231-ace60c46-e181-11e7-9316-f44c2143dc14.png)

No data gets synchronized at the moment. I assume the app requests the data on every connection with the scale?

[...]

PR #113, opened 2017-12-15
openscale support of Beurer BF700 & BF800

linuxlurak:
see https://github.com/oliexdev/openScale/issues/111


oliexdev:
thanks @linuxlurak I merged with commit https://github.com/oliexdev/openScale/commit/4f7675e268352996063f6794caed4529119efa43 the two files BF700 and BF800 into one BF700_800 file.

PR #114, opened 2017-12-22
Misc fixes

erijo:
A couple of minor fixes/improvements. See individual commits for more info.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/gui/activities/DataEntryActivity.java
+++ b/android_app/app/src/main/java/com/health/openscale/gui/activities/DataEntryActivity.java
@@ -181,6 +181,8 @@ public class DataEntryActivity extends Activity {
             id = getIntent().getExtras().getLong("id");
         }
 
+        ScaleData scaleData;
+
         if (id > 0) {
             // keep edit mode state if we are moving to left or right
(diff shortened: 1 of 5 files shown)


oliexdev:
thanks for your PR.

commit d8b6458 , 62c54f9 and b00fb1e were accepted.

For 9b7121f I don't really understand the commit you reset the diff value but we update it again anyway?

For a8bd2cc I only changed that we use the isVisible method but I didn't refactor the method itself because its more clear to understand the method.

erijo:
If diffValue is set when updateValue is called, it will trigger a call to updateDiffValue. But this is unnecessary as we know that updateDiffValue will be called directly after updateValue anyway.

The reason for this attempt to optimize things is that I noticed that it took some time to switch between entries in the dataentry view and I wanted to speed that up.

oliexdev:
I guess its so slow because every time you switch between the data a sql access is executed to get a tuple of data (previous, selected and next one) because I need the correct id which is not continues.

erijo:
I thought so as well initially @oliexdev. But after doing some very rudimentary measurements it looked like reading from the database was fairly quick and that most of the time was spent updating the measurement views. But this was in the emulator and my computer is too slow so it's almost painful to use...

Could you try with and without 9b7121f and see if you can see any difference?

Issue #48, opened 2017-06-03
Bug: Wrong dates in overview diagram


momo42:
I noticed that some dates below the overview diagram are incorrect. This happens with openScale-1.5.2-MultiUserFix and two different users on two different phones. The corresponding list measurement entries are correct. I didn't test the regular 1.5.2 version for this bug yet.

Example ("03.06.17, 03.06.17, 03.06.17" but should be "02.06.17, 02.06.17, 03.06.17":
![screenshot_20170603-001322](https://cloud.githubusercontent.com/assets/28578876/26749504/fdc6583c-480c-11e7-81b3-7e06565f15d7.png)
![screenshot_20170603-001328](https://cloud.githubusercontent.com/assets/28578876/26749508/0737f4ac-480d-11e7-8ac4-1ade24af2ca8.png)

oliexdev:
should be fixed with commit https://github.com/oliexdev/openScale/commit/d36cc0ed1b72ea72ba6693243908bfc0da0c7479
The problem was that the calculation took place between the days in seconds and that was too precise. :grin:

PR #116, opened 2017-12-22
Show days since current date in overview

erijo:
I find it more logical to see how long ago a measurement was done,
instead of how long ago since the last measurement was a measurement
done.

Fix a bug where the days since a measurement was calculated
incorrectly if the measurement was done in the previous year.

Convert the days label to plurals so that "day" is used when the
measurement was done yesterday.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,4 +23,4 @@ after_success:
  - chmod +x ./upload_apk.sh
  - ./upload_apk.sh
 
-script: cd android_app && ./gradlew -i testDebugUnitTest assembleDebug
+script: cd android_app && ./gradlew testDebug assembleDebug
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,4 +23,4 @@ after_success:
  - chmod +x ./upload_apk.sh
  - ./upload_apk.sh
 
-script: cd android_app && ./gradlew assembleDebug
+script: cd android_app && ./gradlew -i testDebugUnitTest assembleDebug
--- a/android_app/app/build.gradle
+++ b/android_app/app/build.gradle
@@ -31,4 +31,6 @@ dependencies {
     compile "com.android.support:support-v4:${supportLibVersion}"
     compile "com.android.support:appcompat-v7:${supportLibVersion}"
     compile 'com.github.lecho:hellocharts-library:1.5.8@aar'
+
+    testCompile 'junit:junit:4.12'
 }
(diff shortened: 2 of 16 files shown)


oliexdev:
Thanks again for your PR. I really appreciate it. :)

Does we go with your calculation into the problem reported in issue #48 because its too precise?

erijo:
It will not print the date as you removed that in that commit. But it will print 0 days for more than the last entry.  If current time is 00:10 and you made a measurement 00:01 and one the day before at 23:50 then this code will say that both measurements were done 0 days ago, while the previous one would have reported 0 days and 1 day ago respectively. 

I'm not so sure that the previous behaviour were more correct. But if you think so I could try to rework the code to be more like it was before, but without the bug that entries on two sides of the end of a year wasn't handled correctly.

oliexdev:
yeah I think it will be better to handle it like before because nobody calculate exact the time. They will assume yesterday I weight so my last measurement was 1 day ago and not 0 day (and 8 hours ago). The hours and minutes are not shown which is also not necessary.

But it would be great if you could eliminate the bug over the year.

erijo:
This new variant should work as the current version on master, minus the bug that it doesn't work when the year has changed. Also, now with unit tests to prove that it works as expected.

oliexdev:
@erijo thanks a lot, I just moved the junit test into an own package name

erijo:
Ok. I just used Android studio to create the unit test and that's the package name it suggested. Personally I have almost no experience with Java and junit so I don't know how it's "normally" done. Does Android studio still find the test when it is in another package?

Btw, have you disabled the Travis builds?

erijo:
@oliexdev: http://junit.org/junit4/faq.html#organize_1 argues that tests should be in the same package as the code as that makes it possible to test methods that's only visible inside the package as well. 

I'd be interested in hearing your view on why it should be in different package. Maybe I can learn something :)

oliexdev:
I would like to have all test classes in one spot. I don't want to mix up program source code and test source code like in the first example in the FAQ.
Currently I also don't want to implement a parallel structure. It blows up the data structure and it makes for me no sense if we only have one or few test cases. I would accept in this structure if I have really for each class a test case.

Yes currently I disabled the Travis builds because of [this problem](https://github.com/travis-ci/travis-ci/issues/8622). Sadly I didn't get a reply.

erijo:
Ok, thanks for the explanation @oliexdev!

PR #118, opened 2017-12-25
Add support for Silvercrest SBF75 scale

matushorvath:
Hello

I added support for SilverCrest SBF75 scale, which looks like a rebranded Sanitas SBF70 with a different device name. As far as I can tell, the scale works fine with these changes.

Thank you

Matus


oliexdev:
thanks @matushorvath please report if something doesn't work.

atmaniak:
Using the current dev build, the app is crashing when the scale is mesuring others data (not weight).
Weight is not saved.

matushorvath:
@atmaniak I will install a current build and try it out. Are you using a SBF75 or SBF70 scale?

atmaniak:
The sticker says : 
Type: SBF 75

PR #121, opened 2018-01-01
Fix travis automatic build and deploy

erijo:
- Upgrade dependency to android-26 to make it build with recent changes.
- Make the build from travis include the commit hash in the version.
- Make deploy reuse the same release and don't create dummy tags.

I've been testing this on my branch and almost everything works. The only thing that doesn't work is that the time for the release isn't updated and remain at the time for the initial release. But I've added the build time to the release body to make it clear when it was built (that is updated on every build).


oliexdev:
awesome @erijo but what was the key to success? How do you manage it that they overwrite the existing pre-release? Must be the tag name the same to be overwritten?

erijo:
The key was to update the tag both locally and on the server. Between `before_deploy` and `deploy`, travis updates all tags by fetching them from github, so if the server isn't updated in `before_deploy`, the tag is overwritten and the release update doesn't work as required. One other key was setting target_commitish to the current commit as otherwise github got confused about which commit the release was for.

PR #124, opened 2018-01-02
Make it easier to install the dev app

erijo:
Change the application id so that the released app and the dev app can co-exist. This makes it much easier to actually test the dev app.

Also use the same key for all dev builds so that the user can easily upgrade between dev versions without having to uninstall the app first.


erijo:
I've rebased this to include your import/export changes @oliexdev, so it should apply cleanly now.

oliexdev:
thanks for your improvement :+1:

PR #131, opened 2018-01-05
SanitasSBF70 problem with time. Catalan/Valencian language added.

comcamot:
1) SanitasSBF70 was saving the date-time at which data was recovered  by the app and tnot the date-time of the measurement. Solved.

2) Catalan/Valencian translation included.


oliexdev:
thanks for your PR :+1:

PR #137, opened 2018-01-11
Refactor measurement views

erijo:
The biggest part of this PR involves refactoring the inheritance hierarchy for MeasurementViews by adding a new FloatMeasurementView that all measurement views that holds a float inherits from. This way the base class MeasurementView can be cleaned up as much code is moved to FloatMeasurementView where it can deal with the value in a more explicit way as it is always float there.

The second part of this involves changing the DataEntryActivity to not save measurements to the database when navigating between them unless they have been modified. This improves the performance when switching between measurements. I have some more things I would like to do in this area, such as changing how the active user is handled, but that is for later.

DataEntryActivity now holds on to a clone of the ScaleMeasurement it is displaying, which makes it easier to update when something changes and cascade the changes to the other views.

Other performance improvements includes not updating the views unless the value has changed and using SpannableStringBuilder instead of Html.fromHtml to style text (only in MeasurementView, not yet in TableFragment).

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/OpenScale.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/OpenScale.java
@@ -23,6 +23,7 @@ import android.os.AsyncTask;
 import android.os.Handler;
 import android.preference.PreferenceManager;
 import android.support.v4.app.Fragment;
+import android.text.format.DateFormat;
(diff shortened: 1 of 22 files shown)


oliexdev:
thanks again @erijo for your awesome improvements of the source code. :+1:

PR #142, opened 2018-01-14
Support importing old CSV exports

erijo:
Add support for importing CSV exports done with all versions back to version 1.0. Also fixed a bug where the user id wasn't set before inserting imports in database.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/OpenScale.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/OpenScale.java
@@ -40,10 +40,12 @@ import com.health.openscale.core.database.ScaleUserDAO;
 import com.health.openscale.core.database.ScaleUserDatabase;
 import com.health.openscale.core.datatypes.ScaleMeasurement;
 import com.health.openscale.core.datatypes.ScaleUser;
+import com.health.openscale.core.utils.CsvHelper;
 import com.health.openscale.gui.fragments.FragmentUpdateListener;
-import com.j256.simplecsv.processor.CsvProcessor;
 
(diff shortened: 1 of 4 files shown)


oliexdev:
Thanks! 94.1 % of all users use the openScale version > 1.6, but the downward compatibility up to version 1.0 is of course even better. :wink:

erijo:
I initially mostly wanted to add support for the 1.6 format, but it got a bit out of hand :)

erijo:
Btw, I noticed a problem with the index in the database. The unique constraint should be on user id + timestamp and not just on timestamp.

oliexdev:
How you have noticed the problem?
It could be theoretically a problem if we have two users with the exact same measurement time otherwise I see no problem yet.

erijo:
I noticed it when testing my changes here. First nothing showed up after import. Then I noticed that user id wasn't set when importing so I fixed that. But no measurements showed up after importing anyway. I tried deleting all measurements but it still didn't work. It turned out to be due to this constraint and all measurements being in the database with user id -1 after the first import. I had to delete the whole database to get it working.

oliexdev:
How did you delete the measurements after your fix with the user id? Because the normal delete button doesn't delete really an entry in the database only mark as deleted. That would mean that all times were still in the database. If you delete all measurements via the settings, you will really delete all measurements in the database.

erijo:
I did the delete from the settings. But I assume it only deletes measurements from the active user, and the measurements in the database had user id -1.

oliexdev:
yes you are right it only deletes all entries of the selected user.

PR #144, opened 2018-01-14
Stacktrace

erijo:
Show a custom activity when a crash happens, enabling the user to copy the stacktrace to the clipboard and points him/her to github for reporting issues.


oliexdev:
thanks, I also looked for a crash report tool yesterday. I ended up with https://sentry.io/ but yours is more lightweight. Nevertheless I hope we won't get so much crash reports. :sweat_smile:

PR #148, opened 2018-01-15
Move test to the location where gradle expects to find them

erijo:
I was trying to add another android test but I couldn't figure out how to actually run it. But after moving the test to androidTest it was really simple to run from android studio.

I also looked at the build log from Travis and as far as I can tell the unit tests aren't running there as no test sources could be found. By moving them to where gradle expects them the tests are now run.

I know you didn't want the tests here, but it makes it much simpler to actually run them so I hope you can accept it.

### DIFF
--- a/android_app/app/build.gradle
+++ b/android_app/app/build.gradle
@@ -5,6 +5,7 @@ android {
 
     defaultConfig {
         applicationId "com.health.openscale"
+        testApplicationId "com.health.openscale.test"
         minSdkVersion 18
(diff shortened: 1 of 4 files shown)


erijo:
I can add that I spent a lot of time trying to get the android tests running on Travis, but it takes too long time to start the emulator for it to work.

oliexdev:
ok although I'm not a big fan of moving the test cases into another subdirectory.

PR #149, opened 2018-01-15
Added possibility to switch unit on Sanitas/SilverCrest/Beurer scales

Dododappere:


oliexdev:
thanks, the problem is that some scales are receiving their weight in a different unit. So I have to check the others scales if they are using `setWeight(...)` or `setConvertedWeight(...)` but I am too tired for today. :sleeping:

Dododappere:
Indeed, we need to make sure that the 'weight' property always contains the correct value in Kg...so if a scale receives LB/ST the setConvertedWeight method should be used....otherwise it won't work consistently

oliexdev:
I checked the other scales with commit https://github.com/oliexdev/openScale/commit/74d954dc382a0d9d4cab4d9620a86d0deff8b8e7
But I am not always sure if they get always a base unit or the converted weight because I can't test it without the scale. So sometimes I have to guess.

Dododappere:
Thanks!  Yes, it is difficult without having all the scales for testing.....

PR #152, opened 2018-01-16
Introduce enum for weight unit and gender

erijo:
Plus a statistics bug fix and a fix for a sometimes (rare) incorrectly calculated age.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/OpenScale.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/OpenScale.java
@@ -39,6 +39,7 @@ import com.health.openscale.core.database.ScaleUserDAO;
 import com.health.openscale.core.database.ScaleUserDatabase;
 import com.health.openscale.core.datatypes.ScaleMeasurement;
 import com.health.openscale.core.datatypes.ScaleUser;
+import com.health.openscale.core.utils.Converters;
 import com.health.openscale.core.utils.CsvHelper;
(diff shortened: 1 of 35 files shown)


oliexdev:
@erijo can you resolve the conflict please? Thanks in advance.

erijo:
@oliexdev: I have rebased the changes on the latest master now.

oliexdev:
thanks, the source code getting better and better. :smile:

oliexdev:
@erijo somehow the Travis build is failing because the license is not accepted. Before it works with the license bug fix workaround. Do you have an idea how to solve it?
It is also a bit strange that Travis downloaded build-tool version 26.0.2 and not 27.0.2....

erijo:
That's strange. The same commit built fine both on my travis instance and in the pull request. The build-tools version 26.0.2 was installed successfully then.

The simplest way to fix it is probably to add the same workaround for build-tools;26.0.2 as for platforms;android-27.

Btw, have you seen this: https://f-droid.org/en/docs/FAQ_-_App_Developers/#how-do-i-change-the-description-and-add-meta-information-like-screenshots

I noticed that you added some screenshot thing, and if I understand things correctly, fastlane can generate screenshots as well and upload all data to the play store and f-droid.

oliexdev:
Yeah I have seen fastlane but with the build in screenshot function it is also works good, beside I I had to remove the UUID suffixes.

PR #155, opened 2018-01-20
Update dependency type to match what the dependency is used for

erijo:
Fixes some of the build problems I've been having when rebuilding tests and saves about 1.5MB when comparing app-debug.apk between master and this branch.


oliexdev:
thanks. Travis seems to work again.

PR #157, opened 2018-01-21
Some more misc changes

erijo:
Fixes a few minor issues that I've come across. Please see individual commits for more info.


oliexdev:
Thanks could you change the behaviour of the left/right buttons that they are not disappear but for instead that they are disabled (e.g. gray).

erijo:
@oliexdev: better now?

oliexdev:
Thanks, I implemented a combined approach. :wink:  So I will disable all navigations buttons if we don't need them. Otherwise the navigations buttons are disabled at the end like you implemented it.
Additionally, I fixed a conversion bug in the evaluation sheet.

erijo:
Good combination!

Good catch with the evaluation sheet conversion bug!

PR #161, opened 2018-01-24
Database upgrade and cleanup

erijo:
Changes the constraint for datetime to also include userId so that two users can have measurement for the same time. Also added a foreign key constraint so that there must be an user to be able to add a measurement. Added migration + test to verify that the upgrade works.

Tested import of a backup database with version 1. To have the migration steps run after import I changed the code to reopen the database after the import is done.

Also introduced caching of the selected scale user so that it doesn't have to be read from the database all the time.

Also some cleanups, such as the removal of unused methods from the old database classes.

### DIFF
--- a/android_app/app/schemas/com.health.openscale.core.database.AppDatabase/2.json
+++ b/android_app/app/schemas/com.health.openscale.core.database.AppDatabase/2.json
@@ -1,0 +1,195 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 2,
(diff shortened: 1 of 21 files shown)


oliexdev:
top :+1:

oliexdev:
With your foreign user id key I get now the following warning:

> Warning:(37, 8) userId column references a foreign key but it is not part of an index. This may trigger full table scans whenever parent table is modified so you are highly advised to create an index that covers this column.

But userId is in the index. Do you have any idea?

PR #163, opened 2018-01-26
Activity state

erijo:
Make data entry activity and graph and table fragment remember the state so that when the user e.g. rotates the screen, the state isn't lost.

Also made switching year in graph view move from e.g. January 2018 to December 2017 instead of to January 2017.


oliexdev:
Can you please change the "static private" to the correct order "private static".
Also please remove the NonNull annotations. Currently I don't want to use any code inspections inside the source code. Thanks.

erijo:
Sure thing, I'll fix it.

erijo:
@oliexdev: fixed now

oliexdev:
thanks now it looks good. 👍

PR #165, opened 2018-01-27
Readd NonNull annotation on ScaleUser gender and scaleUnit

erijo:
Removing the annotations requires adding a database migration as the scheme changes.

### DIFF
--- a/android_app/app/schemas/com.health.openscale.core.database.AppDatabase/2.json
+++ b/android_app/app/schemas/com.health.openscale.core.database.AppDatabase/2.json
@@ -2,7 +2,7 @@
   "formatVersion": 1,
   "database": {
     "version": 2,
-    "identityHash": "6e74b969dde2ca76a6fe8ace46a51f8c",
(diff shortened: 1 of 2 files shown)


oliexdev:
yes but it is in the development version of openScale, so we don't need any migration strategy. People who using the development version should be aware of bugs or experimental source code changes.

erijo:
But this is a change between 1.7 and the dev version. If you run the DatabaseMigration test you'll see that it fails.

erijo:
I added the NonNull annotation when I changed gender and scaleUnit from ints to enums as when they were ints, room automatically made them non-null, and to avoid changing the database scheme I added the NonNull annotation which makes room generate the scheme as before the change.

oliexdev:
Ok I see, then we have to change the migration method or we use the nonnull method again but I think its more confusing for people who will look into the source code, why only these variables have a nonnull annotation.

erijo:
If you merge this @oliexdev, I can work on removing the annotations and adding the migration.

oliexdev:
ok thx

PR #166, opened 2018-01-27
Change order of columns in index

erijo:
By putting userId first we avoid a compiler warnings an most likely
make better use of the index.


oliexdev:
strange that the order of the index is relevant. :confused: 
But big thanks to find the solution.

PR #172, opened 2018-01-31
Add options menu to data entry activity

erijo:
- Remove buttons at the bottom and instead use the back button to go
  back or cancel an add or edit. As for adding a measurement or saving
  an edit, a text button is now shown in the toolbar when adding or
  editing a measurement.
- The delete button is always hidden in the options menu.
- Hide left/right arrows while in edit mode to make it clear that the
  entry must be saved before one can navigate away.
- Show evaluator view by default the first time the activity is shown
  to let new users see it.
- Replace bitmap icons with vector variants. Add indication in
  CREDITS where the icons were downloaded from. The color was changed
  to #cccccc before download and then the icons were imported in
  Android Studio as vector assets.

Test it and let me know what you think @oliexdev.


oliexdev:
thanks for your constantly efforts and ideas @erijo 

But this time I think your changes doesn't fit well with openScale. 

> Remove buttons at the bottom and instead use the back button to go
> back or cancel an add or edit. As for adding a measurement or saving
> an edit, a text button is now shown in the toolbar when adding or
> editing a measurement.

I think it's more clear if they are buttons. Furthermore, for me the "colourful" buttons looks better.

> The delete button is always hidden in the options menu.

Why to hide it? If you need them you will find it better with the icon and again it is a bit more colourful and you need one less click. 

> Hide left/right arrows while in edit mode to make it clear that the
> entry must be saved before one can navigate away.

Again then you have one more click. In the current version it is automatically saved when you go to left or right. Maybe we could show a toast message that a value has been saved.

> Show evaluator view by default the first time the activity is shown
> to let new users see it.

I think the users will discover the evaluator very quickly.

> Replace bitmap icons with vector variants. Add indication in
> CREDITS where the icons were downloaded from. The color was changed
> to #cccccc before download and then the icons were imported in
> Android Studio as vector assets.

Well for bitmaps I don't need a special program to edit it. Beside I am not good in editing in svg graphics. :smile:

erijo:
You have a point with the colorful buttons at the bottom. They could be reintroduced.

As for the delete option I see it as something you don't do that often so it doesn't have to take up space. And putting it in the menu is common in other apps. Plus I've always found the delete icon to be a bit odd: the other two has a background color to indicate when they are active, but the delete icon has another color that doesn't indicate that it's active.

I changed the way edits are done so that when you click edit you enter a special mode that makes it very clear that you are editing a measurement. Then when you are done you click save (or back to cancel) and are returned to the view mode where you can navigate to the next measurement to edit. When I tested this I found it very intuitive.

As for the vector graphics I have no strong opinion. It was just easier to have a single icon instead of multiple.

Have you tested the changes here? If not, please do.

oliexdev:
sure I tested your changes. :smile: 

I like that the delete button is red because it is a warning colour. And sometimes you want (at least for me) to delete many measurements. openScale even have an option to disable the security check co I can delete faster. :wink: 

If you click on the edit button, you will see that you are in the edit mode clearly in my opinion. On the one hand by the green colour and on the other hand by the edit symbol on every editable measurements.

erijo:
If I were to redo this so that it looks more like it does now, but with the following changes:
- addition of back button in the top left corner
- changing the edit mode to don't be a persistent mode but instead be more like I did here where it is a mode entered to do a change and is left by clicking the cancel (or back) or the ok button. In my opinion this would be much clearer to the end user (or at least to me as I've found the current behavior slightly confusing at times).

Is this something that you could consider merging?

oliexdev:
ok then lets do a redesign of all button layouts in openScale  as follow:

1. top title background in the data/user entry activity should be the same blue colour.
2. remove abort and ok buttons with your back button top left and a save/ok symbol on top right everywhere.
3. replace the add measurement button with a floating button in the bottom right
4. move import/export/share buttons to the action menu or maybe use a share symbol button with menu
5. redesign user data entry like the data entry. 
6. move in diagram fragment the month/year button to the action menu as a button (but only visible if the diagram fragment is active)
7. move user out of the settings to ?! maybe to the nav drawer menu?!

Of course, you don't have to implement everything. 😄

erijo:
I think that what you wrote sounds mostly fine and I would like to help out where I can. 

Regarding the save button in the data entry view I've actually changed my mind and think that having a big button at the bottom is more convenient. 

But I'll start with refreshing this PR and then work from there.

oliexdev:
hmm I test implemented point 3 in the list but then the floating button is covering some measurements values. :/
I could add some invisible measurements so the scroll view is longer and the values can be scrolled out!? What do you think?

erijo:
What about adding a Space view at the bottom of the scroll list?

oliexdev:
ok now it looks like that https://github.com/oliexdev/openScale/commit/2a27c3801ddce40ef0344b709c03eb20d9fbb499

oliexdev:
or we move the button to the action menu?

erijo:
I think what you did looks pretty good. But I also think it should be in the menu as it would be good to be able to access from the other views as well. Perhaps in both places?

oliexdev:
I think now its good with commit https://github.com/oliexdev/openScale/commit/a8eb24eb373e5d8fc1d759fb14656ca23b202936

Additionally I remember which fragment was last used and reopen it on start up.

erijo:
I've now rebase the changes in this PR to latest master and made the following changes:
- no SVG icons, instead the current icons are used but tinted white (as otherwise they aren't seen on the black background)
- the delete icon is always visible (ifRoom)

Does that sound alright or have I missed something?

PR #174, opened 2018-02-01
Table view performance improvements

erijo:
The bulk of the changes here (i.e. the last two commits) speeds up loading of the table view by not creating child views that are never displayed and using spanned strings directly instead of via Html.fromHtml. Now it loads almost instant.

Also two clean up commits and a small visual improvement.


erijo:
With this one could increase number of measurements / page in table view to e.g. 200.

erijo:
After going to bed I realized that 8af0aea9a was probably not the right way to do it, so please hold on merging this, or merge everything except that last commit.

erijo:
There, I redid that last commit to something a little less intrusive. Also increased the number of entries per page in the table view to 50.

oliexdev:
great speed up. I guess using spannable strings instead of using html formatter is much faster.

erijo:
Yes, fromHtml was a big part in the profiling result. But so was all float to string that was done but wasn't used.

PR #176, opened 2018-02-05
Color

erijo:
Change app color and theme to material blue 500.

Was it something like this you meant with bullet 1 in #172?

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/gui/activities/DataEntryActivity.java
+++ b/android_app/app/src/main/java/com/health/openscale/gui/activities/DataEntryActivity.java
@@ -15,7 +15,6 @@
 */
 package com.health.openscale.gui.activities;
 
-import android.app.Activity;
 import android.app.AlertDialog;
 import android.content.Context;
 import android.content.DialogInterface;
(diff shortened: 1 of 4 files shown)


oliexdev:
Sorry, no that was not my intention. Can you please remove the colour change from your PR so I can do a clean merge into the master repo?

I guess my points in #172 are obsolete because some changes I have to see how it looks like and how the handling are. 

But the next point is to get rid of the month/yearly switch button in the graph fragment. Do you have any idea? Maybe to migrate as an option in the settings? But then you have no direct access any more... or to open a menu when you long press in the graph!?

erijo:
I removed the color change, but kept the theme change. Not that I know if it made any difference (think not), but I guess the material theme is the new way to do it so might as well switch to that.

As for the graph, perhaps there should be a filter icon (e.g. https://materialdesignicons.com/icon/filter-variant) in the toolbar that could pop up a dialog that could be used to switch mode and even set a start and end range. And potentially more.

oliexdev:
thx. I am not sure about the filter icon... maybe a mockup would help to see if it looks good and works flawlessly.

PR #177, opened 2018-02-06
Data entry v2

erijo:
A retry on PR #172.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/gui/activities/DataEntryActivity.java
+++ b/android_app/app/src/main/java/com/health/openscale/gui/activities/DataEntryActivity.java
@@ -19,12 +19,16 @@ import android.app.AlertDialog;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.content.SharedPreferences;
-import android.content.res.ColorStateList;
 import android.graphics.Color;
+import android.graphics.drawable.Drawable;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
-import android.support.design.widget.FloatingActionButton;
+import android.support.v4.graphics.drawable.DrawableCompat;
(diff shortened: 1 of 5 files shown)


oliexdev:
Some thoughts about that PR:
1. Usage of a save icon for the text "add" or "save" e.g. [save icon](https://www.flaticon.com/free-icon/save_149111#term=save&page=1&position=2)
(I uploaded my scripts to convert svg icons into png [here](https://gist.github.com/oliexdev/e08caeb3fa0ba90fdb5857661512656c))
2. Then the user edit dialog have to be also changed like the edit data entry.
3. Maybe tinting the icon colours like before and not white!?
4, It is not really clear what you have to do if you want to abort your edit mode (the back button is not really enough in my opinion)
5. I am still not sure if its a good idea that when you are in the edit mode that you can't go to the left or right anymore!?

erijo:
I've rebase the PR on top of latest master and re-added the cancel and ok/add buttons at the bottom.

1. I removed the "add" and "save" text icons and replaced them with the button at the bottom instead. One could potentially have both (i.e. both an icon at the top and the button at the bottom). What do you think?
1. They look rather different already, why do you think the user edit dialog has to be changed as well?
1. Do you mean to use a different background (circle?) or just another color? If so, which one?
1. Re-added the cancel button.
1. This boils down to what we expect the common use case to be. For me personally I never edit any measurement once it's done. But perhaps someone with a bluetooth scale has a different use case. Is it common, in your view, to edit multiple measurements at once?

oliexdev:
see latest commit that was my idea but I am unsure if the colour tints icons looks good?

erijo:
Looks good @oliexdev, nice work!

I have a few changes, but I'll post that as a separate PR.

PR #178, opened 2018-02-06
Travis

erijo:
Fix build and make travis build android test as well (but not execute it).

Also fixed the problem with the screenshots starting from the last open fragment instead of the overview.


oliexdev:
thx

PR #188, opened 2018-02-13
Tableview

erijo:
A bit more optimization of the tableview: it now only renders data that is actually seen.

Also, changed the way share is done as the previous way didn't work for me (got some error). It now works without any special permission. I also converted the import function to use intent so it's now possible to import from e.g. a file stored on google drive.

I did an attempt at making export use an intent as well, but the problem there was that the export always created a new file which seemed like something that might not be what one wants when using that. Anyway, it is possible to use share to export the data to e.g. google drive.

Also fixed a crash when trying to import an empty csv file.

### DIFF
--- a/android_app/app/src/main/AndroidManifest.xml
+++ b/android_app/app/src/main/AndroidManifest.xml
@@ -34,6 +34,16 @@
                 <action android:name="android.intent.action.BOOT_COMPLETED"/>
             </intent-filter>
         </receiver>
+
+        <provider
+            android:name="android.support.v4.content.FileProvider"
+            android:authorities="${applicationId}.fileprovider"
+            android:grantUriPermissions="true"
+            android:exported="false">
+            <meta-data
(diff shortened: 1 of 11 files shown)


oliexdev:
What do you mean with the export is always created a new file?

erijo:
I tried using the ACTION_CREATE_DOCUMENT intent for export as then the read/write storage permission wouldn't be needed (assuming the DB backup was changed in the same way). But the problem with that was that it always created a new file as one couldn't select an existing one to overwrite.

But writing this I realize that we could remember the selected export path and ask the user if that should be used and only use the intent if not. That way it would be straightforward to also add an option to automatically perform the export when a measurement is added/modified (i.e. automatic backup).

What do you think, should I attempt to implement such a thing?

oliexdev:
yeah sure good idea :+1: 
for the automatic backup my idea was to generate weekly or monthly (which is configurable) an automatic backup.

PR #195, opened 2018-02-18
add 7 days weight mean to the graph

mmangard:
Hi,
I sketched a visualisation of the 7 day weight mean which is calculated in the statisic page. I think such a feature could be quite handy if you want to estimate how much weight you lost on average in one week.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/gui/fragments/GraphFragment.java
+++ b/android_app/app/src/main/java/com/health/openscale/gui/fragments/GraphFragment.java
@@ -66,6 +66,7 @@ import lecho.lib.hellocharts.model.PointValue;
 import lecho.lib.hellocharts.model.SelectedValue;
 import lecho.lib.hellocharts.model.SubcolumnValue;
 import lecho.lib.hellocharts.model.Viewport;
+import lecho.lib.hellocharts.model.ValueShape;
 import lecho.lib.hellocharts.util.ChartUtils;
 import lecho.lib.hellocharts.view.ColumnChartView;
 import lecho.lib.hellocharts.view.LineChartView;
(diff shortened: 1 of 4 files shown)


oliexdev:
good idea but I don't like how it was implemented. It should be more implemented like the "showMonth" diagram (just you the weekly base as a base unit). Then not only the weight is averaged, but all measurements are averaged.

oliexdev:
implemented with commit https://github.com/oliexdev/openScale/commit/dd6c9b20981d00d85b3bac621d52b38350c62b6f

Please try it out.

mmangard:
I just tried your implementation and want to mention some aspects.
I am not sure if the calculation of the mean is appropriate in your implementation.

Let's take the use-case where one measurement is taken on every day:
In this case the weekly mean of your implementation can be described with the following formula:

mean= 1/64 * mon + 1/64 * tue + 1/32 * wed + 1/16 * thu + 1/8 * fri + 1/4*sat + 1/2 * sun

In order to compare the mean values of different weeks I think the assumtion that the measurement values of the sundays are 32 times more important than the values of the mondays and tuesdays can not be made. 

I think that the implementation where the mean values are implemented as "overlay"  like the regression line make some interesting visualisations possible.  

1) The rightmost value corresponds to the value of the statistic page 
2) The measurement values in the background give an idea of the variance which I think  is  a good combination for the mean values. 

screenshot_20180224-211550

oliexdev:
you are right the calculation was wrong I fixed it with commit https://github.com/oliexdev/openScale/commit/298254366956aa5e87976371b0dab288573793ab
Please test it extensively, because I am not sure if there is still an error in this source code.

mmangard:
The calculation looks good.  One thing/idea which came into my mind is if it makes sense to display "average"-specific information  (e.g. number of measurements, startdate, enddate and values used for the calculation ) if the user tabs on the label.
currently only the value of the measurement on monday is displayed.

PR #197, opened 2018-02-19
Measurements

erijo:
Recreate activity after settings have changed. This makes both the app theme take effect immediately and updates measurement views correctly after adding/removing measurements.

Add option so that date/time is once again placed last on the data entry activity. The views don't jump around as much then when switching between edit/view. 

Some other minor changes/improvements.


oliexdev:
great :)

Issue #164, opened 2018-01-27
Change data entry


natrius:
Hello, 

thanks for the app. I'm using **Version:** 1.7 (23) via the f-droid store.
**Android:** 7.1.2
Just some workflow changes to make manual entries easier.

- Choose in what order the data has to be entried.
- Add option to automatically open the next entrie field.

**Explanation:**
My scale does show the data pretty fast and has now bluetooth. So i have to look for the data, press the "Add" button, entry the data, press "Ok", search the next "Add" button, and so on.
It would be much easier if i could just "Start Entry" - write data - "OK" - write next data - "OK" - finished

(Additional question: How can i help with translations?)

Greetings

erijo:
I had the same idea in #81 and I plan on implementing it eventually.

erijo:
@natrius: if you haven't found it already, translations can now be done here: https://hosted.weblate.org/projects/openscale/

natrius:
@erijo
Thanks, maybe you want to add this information into the readme to the "Contribute" section.

erijo:
Good idea, perhaps @oliexdev can add something?

Issue #81, opened 2017-10-11
Improve measurements entry


erijo:
On my scale, after measurements have been taken, the values for body fat, muscles etc. loop by twice after which they disappears. To make measurements entry in the app smoother it would be nice if you could configure the order in which the measurements appear during entry to be the same as on the scale. Then after entering the weight and pressing ok (or next) it would display the next entry in the configured order. It would also be nice if there where a + and - button on the entry screen to quickly increase or decrease a value by e.g. 0.1 kg.

oliexdev:
Hi erijo,

I understand your problem and yeah your idea sounds good but to configure the order is not simple to implement yet. But I will keep that in mind. Thanks.
The + and - buttons is much easier to implement. If I have a bit more time I will implement it.

oliexdev:
added + and - buttons with commit https://github.com/oliexdev/openScale/commit/22d812a2ccd97bfbd407997959fbb4ed20532804

PR #198, opened 2018-02-20
Make it possible to change order of measurment views

erijo:
By long pressing a view in the data entry activity, the view can be dragged up or down to change the order. The order is then saved and reflected in overview, graph, table and data entry fragments/activity.

This fixes part of #81 (and #164).

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/gui/MainActivity.java
+++ b/android_app/app/src/main/java/com/health/openscale/gui/MainActivity.java
@@ -59,7 +59,6 @@ import com.health.openscale.gui.fragments.OverviewFragment;
 import com.health.openscale.gui.fragments.StatisticsFragment;
 import com.health.openscale.gui.fragments.TableFragment;
 import com.health.openscale.gui.utils.PermissionHelper;
-import com.health.openscale.gui.views.MeasurementView;
 
 import java.lang.reflect.Field;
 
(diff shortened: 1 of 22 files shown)


oliexdev:
Great. Just one question is it possible to easily migrate this approach into the preference menu? Because I think it makes more sense to set the order into the preference menu not on the activity.
Like to have a new subpage in Preference->Measurements->Measurement order->Subpage where  you can order the measurements.

erijo:
Yes, I think that is doable. Will look into it.

erijo:
How does this look and feel @oliexdev?

oliexdev:
yeah its awesome big thx @erijo

erijo:
I somehow managed to miss a crash that I introduced. Should be fixed in #200.

oliexdev:
@erijo I noticed in the dark theme the subscreen measurement order preference texts doesn't have the correct white colour. I already tried to change that but I couldn't find the problem yet. Maybe you have an idea?

erijo:
Maybe MeasurementOrderPreference needs to call super with some suitable value for defStyle*?

oliexdev:
fixed with commit https://github.com/oliexdev/openScale/commit/d4a968d6e4991525e7b1ebdf7cc4f2532b841b79

PR #199, opened 2018-02-21
Add next button to input dialog for moving directly to next measurement

erijo:
This fixes the last part of #81 (and #164) (together with #198).


oliexdev:
Good idea. But currently the dialog is always closed and opened again which results into a flickering. But anyway it works. :+1:

erijo:
Yes, it could perhaps be improved. At least it's a start :)

PR #200, opened 2018-02-22
Add missing listener unregister

erijo:
Also don't call getContext() directly as that requries newer API version.

Fixes a crash after my last PR. Don't know how I could miss that...

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/gui/preferences/MeasurementPreferences.java
+++ b/android_app/app/src/main/java/com/health/openscale/gui/preferences/MeasurementPreferences.java
@@ -214,11 +214,17 @@ public class MeasurementPreferences extends PreferenceFragment implements Shared
     }
 
     @Override
+    public void onPause() {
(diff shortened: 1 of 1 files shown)


oliexdev:
no problem :wink: 
I will release the next version of openScale on the weekend. Your measurement order is really a awesome improvement that distinguish to other health apps.

oliexdev:
I don't want to push people. If they want to translate something they will probably come again.
With commit https://github.com/oliexdev/openScale/commit/f1a9f0a71d1bd3adab18ca4e26755943bc8265b1 (wrong comment description though) I Improved a bit better the weekly x-axis view.

erijo:
I noticed now that to use the intent for the export function as well, minSdkVersion needs to be set to 19 as ACTION_CREATE_DOCUMENT was added in that version. How do you feel about that?

erijo:
According to these numbers, going to api level 19 wouldn't affect that many: https://developer.android.com/about/dashboards/index.html

oliexdev:
api level 19 as a minimum is ok.

Issue #169, opened 2018-01-30
[Suggestion] Import/Export from Google Account


Skyrid3r:
Is there anyway that there can be a feature where the data can be imported and exported (optional) from a google account? This would be helpful to easily share the data between multiple devices running the same google account. Also for backup reasons which may be useful.

erijo:
The backup reason might be a reason to increase API level to >= 23 as that gives apps the auto backup feature. That one the other hand has the downside of requiring location for Bluetooth to work.

Another idea I've been having is to use the overflow menu on the table view to hold export and import to CSV options. This could then also have options to "share" the CSV file to other applications such as Gmail, Google drive, etc. as is seen in many other apps.

oliexdev:
@Skyrid3r openScale won't have any integrated Gmail support in openScale, read the FAQ.

But @erijo second idea could be an option or you could also export the backup files into a Google-Drive or Dropbox folder which is automatically sync. 

An automatic local backup of the database (e.g. once a month) would be also a good idea for openScale.

oliexdev:
I implemented a share button for the CSV export. But I removed the abort button for that because it looks strange with three buttons or does anybody have a better idea where to place the share button?

erijo:
I think it would fit nicely in an options menu (i.e. the three dots menu in the top right corner). Personally I think export and import should go there as well to leave more room for measurements in the table view.

erijo:
As of #188 it's possible to both import from Google drive and export (by using the share option).

So I would say that this can be closed.

erijo:
@Skyrid3r: can you test the latest development version (https://github.com/oliexdev/openScale/releases/tag/travis-dev-build) and see if it does what you want. If yes, please close this issue, if no please let us know what's missing.

Skyrid3r:
Mind sharing where this export function is located? Latest development build is from Jan, 2. I can't find it anywhere. Not even in the 3 dot/stripe menu.

erijo:
I should also mention that for export to e.g. google drive you have to select "Share" in the menu.

Skyrid3r:
I don't have any share option in the menu though. Using 1.7 dev. Is it because I'm using a different language?

erijo:
In the Table view there is an icon with two arrows, one pointing up and the other down. If you click this, do you mean that you only see Import and Export, but not Share?

Skyrid3r:
[...]

PR #201, opened 2018-02-23
Move table menu to toolbar

erijo:
Move import/export/share to toolbar menu as discussed in #169?

How does this look @oliexdev?


oliexdev:
thx

PR #204, opened 2018-02-24
Strings

erijo:
Removed strings that aren't used and removed "Body" from many strings. Should be enough to just say fat, water, muscle and height.


oliexdev:
Why do you want to be less precise in the wording? We have enough space for the text.

erijo:
I did it for consistency. Height (without body) matches better with weight and is more common in daily use.

Water was body water in one place and just water in another. We could go with body fat and body water, but should it then be body muscle for consistency? That didn't sound all right so I went with the option without body. 

But I'm ok with keeping body in fat and water if that is what you think.

oliexdev:
Well I am not a English native speaker but lets keep the body in fat and water.

erijo:
Re-added body to fat and water.

oliexdev:
thx

PR #207, opened 2018-02-25
Fix memory leak

erijo:
I think this fixes the memory leak. At least there is only one instance of MainActivity and fragments after screen rotation (as seen in the profiler once the GC has run).


oliexdev:
great thanks for the memory leak fix :)

PR #208, opened 2018-02-27
Export

erijo:
Use ACTION_CREATE_DOCUMENT intent to select export path. Since an existing file can't be picked from this intent, remember the previous export path and ask the user if he/she wishes to overwrite that file instead of creating a new upon the next export.


oliexdev:
thx

PR #209, opened 2018-02-27
Misc

erijo:
A translation update and a change to center the measurement name when in add mode. I think it looks slightly better.


oliexdev:
thx

PR #211, opened 2018-02-28
Release old persistable permission when new export is done

erijo:
A minor cleanup. After this I'm hopefully done with the export thingy ;)


oliexdev:
then I can incremental release the next version of openScale.

PR #213, opened 2018-03-05
Improve measurement input dialog

erijo:
* Add inc and dec buttons to the dialog for quickly doing small
  adjustments in the same way as in activity.

* Improve "next" measurement behavior by not recreating the whole
  dialog and instead update the icon, title, view and buttons.


erijo:
Do you think we should remove the inc/dec buttons on the measurement view now @oliexdev, or do the still fill a purpose? It could be that you accidentally hit one of them when aiming at the pen.

Issue #216, opened 2018-03-07
Bodyfat / -water in kg uses the wrong scale


hirnsushi:
Not sure if this can be easily solved, the different values would need to be calculated for every new measurement.
![img_20180307_150530](https://user-images.githubusercontent.com/5706729/37096595-6609dea0-2219-11e8-952b-af0efea73be7.jpg)

erijo:
I don't think I understand the issue. Could you please elaborate?

hirnsushi:
If you look at body fat for example the arrow suggest I'm on the lower end of the green zone.
That would be correct if my weight was exactly 100kg, it isn't.
For comparison the same measurement in percent:
![img_20180307_162827](https://user-images.githubusercontent.com/5706729/37101005-c2be6e1c-2224-11e8-9fa4-6ff49986ec16.jpg)

erijo:
Ah, now I see.

I can look into this if you want @oliexdev.

oliexdev:
@erijo sure go ahead.

erijo:
@hirnsushi: please try the latest [dev build](https://github.com/oliexdev/openScale/releases/tag/travis-dev-build) and see if it looks better now.

hirnsushi:
Works great, thank you 👍

PR #217, opened 2018-03-07
Issue 216

erijo:
Fix for issue #216.


oliexdev:
thanks for fixing it @erijo

PR #221, opened 2018-03-08
iHealth HS3 scale

JohnLines:
Basic scale support - weight only ,which is all the scale supports, now working for me.

Attached it picture of the scale

dsc01135


oliexdev:
thanks for your effort to support the iHealth HS3 scale.

PR #222, opened 2018-03-08
Next v3

erijo:
Some more changes to the input dialog. 

I removed the inc/dec buttons from the measurement view as an experiment. Let me know what you think. It removed some clutter, but I'm not 100% sure it's a good idea.

About removing the open-by-default soft input: when the dialog was opened it was placed at one position, just to move up a short while later when the keyboard poped up. And as the common use case now should be to press the up or down buttons I don't think it needs to be shown by default any more.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/gui/views/CommentMeasurementView.java
+++ b/android_app/app/src/main/java/com/health/openscale/gui/views/CommentMeasurementView.java
@@ -76,11 +76,6 @@ public class CommentMeasurementView extends MeasurementView {
     }
 
     @Override
-    protected boolean showSoftInputForInputDialog() {
-        return true;
-    }
-
-    @Override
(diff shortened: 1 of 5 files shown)


oliexdev:
I think its good to have also an inc / dec button in edit mode. The icon idea to the next button is too much for me, I prefer it without the icon.

Yeah to show the soft input menu by default is not really necessary I am not sure if it an advantage or disadvantage to show it by default!? If you want to enter a complete different value then it is an advantage but if you make only small changes then you could also the inc/dec buttons...

erijo:
My thinking is that normally the measurements don't change that much between between each entry so that it's quicker to user the inc/dec buttons. And showing the input by default makes the input dialog jump which doesn't look good.

I will drop the first two commits and only push the last one (i.e. the don't show soft input).

PR #233, opened 2018-03-23
 Refactor preferences

### LINKED ISSUE #224: Feature Request: Option to disable comments
It would be great if there was an option to disable comments, like there is for the other measurements.

### LINKED ISSUE #225: Feature Request: Make the graph on the overview page adjustable
With muliple types of measurements enabled, the graph on the overview page basically defaults to a straight line.
Maybe it could get it's own settings for showing / hiding of measurements or inherit the options chosen in the diagram tab.

### LINKED ISSUE #226: BMI can't be disabled
As the title says - BMI calculation and display can't be disabled (and thus hidden in the overview graph; somewhat related to  #130).

erijo:
- Make it possible to disable all measurements (except weight). Fixes #226 and fixes #224.

- Add separate preference screens to all measurements to configure e.g. if the measurement should be displayed in the overview graph (this fixes #225 in a different way compared to previously), if relative or absolute weight should be used etc.

- Include reordering of measurements in the main measurement preference screen.

+ some other fixes.

@oliexdev: if you have time, please take a look and let me know what you think.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/OpenScale.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/OpenScale.java
@@ -259,21 +259,21 @@ public class OpenScale {
         }
 
         MeasurementViewSettings settings = new MeasurementViewSettings(prefs, WaterMeasurementView.KEY);
-        if (settings.isEstimationEnabled()) {
+        if (settings.isEnabled() && settings.isEstimationEnabled()) {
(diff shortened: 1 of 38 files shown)


oliexdev:
looks great @erijo I think now everything on the GUI is dynamically generated from the measurements. Huge step forward to support more measurements for issue #4. I have in my mind to have a separate tab page (in the data entry activity) for body measurements (e.g. thigh, waist, hip, chest, neck, biceps and so on) and scale measurements (e.g. weight, fat, muscle, water and so on). What do you think?

PR #243, opened 2018-04-13
Spelling: Language rework

comradekingu:
RMR as per https://en.wikipedia.org/wiki/Resting_metabolic_rate

### DIFF
--- a/android_app/app/src/main/res/values/strings.xml
+++ b/android_app/app/src/main/res/values/strings.xml
@@ -27,7 +27,7 @@
 
     <string name="label_weight">Weight</string>
     <string name="label_bmi">Body mass index (BMI)</string>
-    <string name="label_bmr">Basal metabolic rate (BMR)</string>
+    <string name="label_bmr">Resting metabolic rate (RMR)</string>
     <string name="label_fat">Body fat</string>
     <string name="label_water">Body water</string>
     <string name="label_muscle">Muscle</string>
@@ -47,7 +47,7 @@
     <string name="label_last_week">Last 7 days</string>
     <string name="label_last_month">Last 30 days</string>
     <string name="label_weight_difference">Weight difference</string>
-    <string name="label_goal_date_is">Goal date is</string>
+    <string name="label_goal_date_is">Goal date:</string>
     <string name="label_days_left">Days left</string>
 
     <string name="label_date">Date</string>
@@ -70,8 +70,8 @@
     <string name="label_export">Export</string>
     <string name="label_delete_all">Delete all</string>
 
-    <string name="error_value_required">Value is required</string>
-    <string name="error_value_range">The value is not in range</string>
+    <string name="error_value_required">Value required</string>
+    <string name="error_value_range">The value not in range</string>
     <string name="error_exporting">Error exporting</string>
     <string name="error_importing">Error importing</string>
     <string name="error_user_name_required">Error: Username required</string>
@@ -189,7 +189,7 @@
     <string name="label_weeks_view">Week view</string>
 
     <string name="permission_not_granted">Permission not granted</string>
-    <string name="permission_bluetooth_info">Coarse location permission is needed to search for Bluetooth devices. It can be revoked after the device is found.</string>
+    <string name="permission_bluetooth_info">Coarse location permission needed to search for Bluetooth devices. It can be revoked after the device is found.</string>
     <string name="permission_bluetooth_info_title">Info</string>
     <string name="label_next">Next</string>
     <string name="label_press_hold_reorder">Press and hold to reorder</string>
--- a/android_app/app/src/main/res/values/strings.xml
+++ b/android_app/app/src/main/res/values/strings.xml
@@ -23,7 +23,7 @@
     <string name="label_delete">Delete</string>
     <string name="label_add_user">Add user</string>
     <string name="label_share">Share</string>
-    <string name="label_share_subject">openScale CSV data export (%s)</string>
+    <string name="label_share_subject">CSV data export (%s)</string>
 
     <string name="label_weight">Weight</string>
     <string name="label_bmi">Body mass index (BMI)</string>
@@ -31,7 +31,7 @@
     <string name="label_fat">Body fat</string>
     <string name="label_water">Body water</string>
     <string name="label_muscle">Muscle</string>
-    <string name="label_lbw">Lean body weight</string>
+    <string name="label_lbw">Lean body mass</string>
     <string name="label_waist">Waist circumference</string>
     <string name="label_hip">Hip circumference</string>
     <string name="label_comment">Comment</string>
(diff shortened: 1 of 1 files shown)


android_app/app/src/main/res/values/strings.xml
```diff
@@ -74,51 +74,51 @@
...
     <string name="label_mergeWithLastMeasurement">Merge with last measurement</string>
     <string name="label_bluetooth_searching">Searching for Bluetooth scales</string>
-    <string name="label_bluetooth_searching_finished">Finished searching for Bluetooth scales</string>
+    <string name="label_bluetooth_searching_finished">Search finished</string>
     <string name="label_device_type">Device Type</string>
 
-    <string name="label_enable_labels">Label on data point</string>
-    <string name="label_enable_points">Point on data point</string>
+    <string name="label_enable_labels">Data point label</string>
+    <string name="label_enable_points">Data point</string>
```
comradekingu: I am not 100% sure about this

android_app/app/src/main/res/values/strings.xml
```diff
@@ -70,55 +70,55 @@
     <string name="label_export">Export</string>
     <string name="label_delete_all">Delete all</string>
 
-    <string name="error_value_required">Value is required</string>
-    <string name="error_value_range">The value is not in range</string>
+    <string name="error_value_required">Value required</string>
+    <string name="error_value_range">The value not in range</string>
```
erijo: Change "The value" -> "Value". Otherwise it sounds strange.

android_app/app/src/main/res/values/strings.xml
```diff
@@ -70,55 +70,55 @@
...
+    <string name="error_user_name_too_short">Error: Username must be 3 characters or more</string>
+    <string name="error_height_required">Error: Height required</string>
+    <string name="error_initial_weight_required">Error: Initial weight required</string>
+    <string name="error_goal_weight_required">Error: Goal weight required</string>
+
+    <string name="info_data_deleted">Entry deleted</string>
+    <string name="info_data_all_deleted">All entries deleted</string>
+    <string name="info_data_exported">Exported to</string>
+    <string name="info_data_imported">Imported from</string>
+    <string name="info_enter_value_cm">Value in cm:</string>
```
erijo: This and the two below should not end with colon. They are shown inside a text edit where I think it looks weird with colon.

android_app/app/src/main/res/values/strings.xml
```diff
@@ -70,55 +70,55 @@
...
     <string name="info_enter_user_name">Enter your name</string>
-    <string name="info_no_selected_user">No user exists. Please create a new user in the settings.</string>
-    <string name="info_no_evaluation_available">Can\'t evaluate the value</string>
+    <string name="info_no_selected_user">No user exists. Please create one in the settings.</string>
+    <string name="info_no_evaluation_available">Could not evaluate the value</string>
 
-    <string name="question_really_delete">Do you really want to delete database entry?</string>
-    <string name="question_really_delete_all">Do you really want to delete all database entries?</string>
-    <string name="question_really_delete_user">Do you really want to delete the user? </string>
+    <string name="question_really_delete">Confirm deletion of entry.</string>
```
erijo: The buttons in the dialog where this and the two below are shown are yes and no. Doesn't the previous text sound better?

android_app/app/src/main/res/values/strings.xml
```diff
@@ -153,34 +153,34 @@
...
-    <string name="label_feedback_message_negative">No, thanks</string>
+    <string name="label_feedback_message_positive">OK</string>
+    <string name="label_feedback_message_negative">No</string>
 
-    <string name="error_max_scale_users">Maximum number of concurrent scale users reached</string>
+    <string name="error_max_scale_users">Max. number of concurrent scale users reached</string>
     <string name="info_step_on_scale">Please step barefoot on the scale for reference measurements</string>
     <string name="info_measuring">Measuring weight: %.2f</string>
 
-    <string name="customactivityoncrash_error_activity_error_occurred_explanation">An unexpected error occurred.\n\nPlease create a new issue including the detailed error details on\nhttps://github.com/oliexdev/openScale/issues</string>
```
erijo: The "error details" part was a reference to the other string included in the crash screen to indicate that that information should be included.

erijo:
I've merged some of the changes in this pull request, but some remain. Could you please rebase your branch so we can see what's left?

android_app/app/src/main/res/values/strings.xml
```diff
@@ -27,7 +27,7 @@
 
     <string name="label_weight">Weight</string>
     <string name="label_bmi">Body mass index (BMI)</string>
-    <string name="label_bmr">Basal metabolic rate (BMR)</string>
+    <string name="label_bmr">Resting metabolic rate (RMR)</string>
```
erijo: @oliexdev: what's your view on changing BMR -> RMR?

oliexdev: I think most people don't know what RMR is (me included).

I think it is better to stay with BMR. I am not an expert in medicine but the [The Mifflin St Jeor Equation](https://en.wikipedia.org/wiki/Basal_metabolic_rate) which we use listed as BMR!?

erijo: That was my thinking as well. Even if RMR is more "correct" (not saying it is as I don't know that) there is a value in using a term that people recognize. BMR is also the term I've seen being used in scale documentation.

comradekingu: Most people don't know what BRM is either. In the field of non-medical people, that know either, they will often mistake one for the other.

Fortunately it states "metabolic rate", so even without any prior knowledge, it is descriptive. Should anyone know one and want to know how it differs, there is plenty of resources to consult.

My view is to use RMR for RMR and BRM for BRM.

oliexdev: RMR and BMR values differs only about 10% with *clinical* devices! 

Normally the BMR is calculated with the [Harris–Benedict_equation](https://en.wikipedia.org/wiki/Harris%E2%80%93Benedict_equation) but this formula was revised by Mifflin and St Jeor in 1990 which is the formula we use. openScale is fully transparent in the calculation we are using. 

In the end, all these formulas are estimates that are close to each other. 

Alternative we add a new RMR with the formula by Mifflin and St Jeor  and use BMR with the formula by Roza and Shizgal. But again both values are close to each other and most people don't know the different. I see here more an academic problem. 

What I don't want is to refactor all names in the source code, so I will close this PR.

comradekingu: I am not saying it doesn't. Nor do I care all that much, I can change it back if you open this again.
The 44 other fixes only exist in Norwegian Bokmål right now.

android_app/app/src/main/res/values/strings.xml
comradekingu: Fine by me

android_app/app/src/main/res/values/strings.xml
comradekingu: It is an unmarked noun error. Should be "the database entry". It is longer, in a more commonly used convention. I favor mine, but I can change it.

android_app/app/src/main/res/values/strings.xml
comradekingu: I would think that is implied, but i can change it to that effect.

erijo:
Reopen so we can get the non-RMR changes.

Issue #253, opened 2018-04-28
Improve first user experience


erijo:
It would be good ease the setup for new users. Here are some ideas I've been thinking on:
* Remove initial weight. It's confusing to have to enter it when you want to connect to a scale to get the weight.
* Ask the user if he/she wants to connect a Bluetooth scale after adding the user.
* Change "device type" in Bluetooth settings to something that indicates that it starts a device search.
* Make goal date/weight optional.

oliexdev:
> Remove initial weight. It's confusing to have to enter it when you want to connect to a scale to get the weight.

see issue #43 for the reason.

> Ask the user if he/she wants to connect a Bluetooth scale after adding the user.

Not all users have a Bluetooth scale. It is not hard to find the Bluetooth settings. Maybe if no device is selected and a user pressed the Bluetooth icon, we could show the Bluetooth settings menu...

> Change "device type" in Bluetooth settings to something that indicates that it starts a device search.

Why? It will indicate the searching process if you click on it.

> Make goal date/weight optional.

In my opinion the whole statistic page needs a rework. :/ Currently we don't really do some evaluation with the goal date and goal weight. But these two items needs to be set otherwise we can't draw the goal line or calculate the remaining days. If we reworked the statistic page, we could change it completely into something different!? You are welcome to post some ideas or sketch ups what we can do with the statistic page and how it should be looked like.

erijo:
I'll have to think some more about the initial weight problem. Thanks for the pointer to the reason it was added.

After adding the first user, we could perhaps show some help that indicates that the user should press + to add manual measurements or the Bluetooth button to connect a scale (assuming we do as you suggested which sounded like a good idea).

"Device type" says nothing about it starting a search for a scale (at least not to me). I changed it to "Your Bluetooth scale" which at least gives a hint about what it is for.

Agree that the statistics page needs some work. But that will have to wait for another rainy day :)

oliexdev:
> After adding the first user, we could perhaps show some help that indicates that the user should press + to add manual measurements or the Bluetooth button to connect a scale (assuming we do as you suggested which sounded like a good idea).

Only a few users will read that and it will only annoy the rest of the users. I don't think it is really complicated to understand the UI concept. Sure some exceptions will always exist that doesn't understand how it works but these users are not willing to read neither. 

[...]

PR #254, opened 2018-04-28
Remove user initial weight

erijo:
As stated in #253 it is confusing for a first time user to have to
enter an initial weight when adding an user.

The comment from a first time user was something like "Why do I need
to enter this? I don't know what my weight is. That's what the scale
is supposed to tell me!".


erijo:
I see. I'll have to think about this some more.

erijo:
Dropping this until I have some more time to look into it.

PR #256, opened 2018-05-02
Add italian language

gablau:


erijo:
Thanks for your contribution! Would you mind adding Italian to the top level README.md file as well?

erijo:
Never mind, I added it. Thanks again!

PR #257, opened 2018-05-02
Fix a translation

gablau:


erijo:
Thanks!

If you haven't seen it already, you can do the translation on https://hosted.weblate.org/engage/openscale/ and have it automatically committed. This way works, but weblate may be simpler to use.

PR #258, opened 2018-05-06
Add support for Easy Home (Aldi) 64050

o0-zzuf-0o:
Appears to be a clone of the MGB scale with a different device name and appearance

Image:
easyhome-64050


oliexdev:
thanks @o0-zzuf-0o for your contribution.

PR #259, opened 2018-05-06
Sets scale unit during initialisation

o0-zzuf-0o:
Use user preferences to set correct unit.
Lbs is currently hardcoded.

Tested with Easy Home 64050 (support added #258) which appears to be a clone of the MGB scale.

Probably worth testing with the original MGB scale, but works fine with the Easy Home unit

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothMGB.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothMGB.java
@@ -113,7 +113,7 @@ public class BluetoothMGB extends BluetoothCommunication {
                 break;
 
             case 6:
-                writeCfg(0xFE, 6, 1, 0);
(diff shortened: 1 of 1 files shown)


oliexdev:
could you check if you got always the right weight if you switch the unit in the app?
Otherwise we need to change the `setWeight(...) `to `setConvertedWeight(..)` in the `BluetoothMGB.java` file.
It will depends what the scale is always sending to the app the base kg weight or in the unit which were set.

o0-zzuf-0o:
Yes, the weight is always consistent and correct. 
The measurement data sent is always in kg.

I have verified this by:
- switching to a different unit
- disconnecting / reinitializing the scale
- checking the raw data in the logs
- verifying that the App does the correct conversion

I cycled through all the units and back to the original and it all checks out.

oliexdev:
thanks for verifying.

PR #260, opened 2018-05-08
Add measurement widget

erijo:
The widget can be added to the home screen and can be configured to
show any of the measurements that are enabled. It can also be resized
from 1x1 to <max width>x1, adapting to the size change and showing a
reasonable amount of information at each size. The default size is
4x1.


oliexdev:
awesome @erijo I just added an colour evaluation indicator and for one-cell-widget an icon on top.

erijo:
I like the evaluation indicator. Nice!

But the icon in 1x1 mode pushes the unit halfway outside the widget. Try e.g BMR or weight when it's 100+.

erijo:
The delta value is still not visible.

oliexdev:
not on my smartphone :/

erijo:
You may need to test with a less high end emulator then :)

oliexdev:
now it should work also with small density devices at least I have tested it with a simulated Nexus 5 device.

erijo:
I did a few more changes. Now I can see the unit as well on my device :)

oliexdev:
but now it looks a bit wired on devices with high resolutions. ;)
Anyway, we have something working.

erijo:
Which resolution/device does it look weird on? Also could you please elaborate on what you mean with weird? I would like to look into it a bit.

erijo:
I tested in an emulator using a Pixel 2 device and it seems like the widget is higher that it is wide. On a low res device width and height for a 1x1 widget seems more equal.

I did some small changes like changing text alignment to center instead of right and allowing value to be 2 lines (otherwise the unit isn't shown for 100+ weight). I think it looks acceptable on all devices I've tested on now.

What do you think?

PR #264, opened 2018-05-13
Body measurements

oliexdev:
I added some more body measurements, see issue #4. Unfortunately, the CSVHelperTest failed. Why are the old import functions doesn't work anymore if I add more measurements? I thought the library try out every combination!?

Do you have an idea @erijo ?


android_app/app/schemas/com.health.openscale.core.database.AppDatabase/2.json
```diff
@@ -80,6 +80,12 @@
             "affinity": "REAL",
             "notNull": true
           },
+          {
```
erijo: You need to revert all changes to version 2.

android_app/app/src/main/java/com/health/openscale/core/database/AppDatabase.java
```diff
@@ -75,5 +75,26 @@ public void migrate(SupportSQLiteDatabase database) {
             }
         }
     };
+
+    public static final Migration MIGRATION_2_3 = new Migration(2, 3) {
```
erijo: Perhaps add a test for this migration?

erijo:
I can look into the csv test tonight when I get to a computer.

erijo:
The CSV problem should be fixed now.

android_app/app/src/main/java/com/health/openscale/core/database/AppDatabase.java
```diff
@@ -75,5 +75,27 @@ public void migrate(SupportSQLiteDatabase database) {
...
+        @Override
+        public void migrate(SupportSQLiteDatabase database) {
+            database.beginTransaction();
+            try {
+                database.execSQL("ALTER TABLE scaleMeasurements ADD COLUMN visceralFat REAL NOT NULL DEFAULT 0");
```
erijo: It might be that you need to add the columns in the right order so that the final order is the same as when the table is created from scratch.

But a test will tell you if this is needed.

android_app/app/src/main/java/com/health/openscale/gui/views/Caliper1MeasurementView.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Override
+    protected void setMeasurementValue(float value, ScaleMeasurement measurement) {
+        measurement.setCaliper1(value);
+    }
+
+    @Override
+    public String getUnit() {
+        return "mm";
```
erijo: Shouldn't this be presented in cm? Feels a bit too exact to enter a mm value with two decimals.

PR #275, opened 2018-05-24
Spelling: Time to weigh → Time to weigh in

comradekingu:


erijo:
I'm not so sure about this. Could you please elaborate on why you think it should be like this?

@bontibon: as you just changed this string, do you have an opinion on this?

What about changing it to "Time to weigh yourself"?

bontibon:
"Time to weigh in" and "Time to weight yourself" are both an improvement, though I think the latter reads better.

comradekingu:
My sentiment is that "in" is shorter than "yourself", and it isn't always the case that the person reading it is the person weighed. I think the figure of speech attentive to the action, rather than the person, makes it seem more integrated with the language. Not in the least because English is very much focused around actions.

erijo:
Ok, you've convinced me. Thank you both!

oliexdev:
btw this text is editable by the user...

PR #293, opened 2018-06-23
Danish translation

rolfmadsen:
Thank you for a great app, which is leaps and bounds over the official Beurer Healthmanager app supplied with my Beurer BF700 scale.

My humble contribution is a danish translation, but my technical skill-set is not a level where I can test the translation in the app, and so some translations may be out of context.


erijo:
Thanks for your contribution! You can test it in the [dev version](https://github.com/oliexdev/openScale/releases/tag/travis-dev-build) and then make changes using [weblate](https://hosted.weblate.org/engage/openscale/). The changes will then automatically become available in the dev version (potentially after some delay).

PR #297, opened 2018-07-01
Ignore 17th byte of the message in the Excelvan scales; some variants…

funkwit:
… include an optional 'physiological age'.

This allows supporting (at least) the Excelvan CF366BLE using the existing driver with no other changes needed.


funkwit:
Argh, had a typo. Closing.

PR #298, opened 2018-07-01
Ignore 17th byte of the message in the Excelvan scales; some variants…

funkwit:
… include an optional 'physiological age'.

This allows supporting (at least) the Excelvan CF366BLE using the existing driver with no other changes needed.


erijo:
Thanks for the PR!

Could you please take a photo of the scale as well and attach it here? For the [gallery](https://github.com/oliexdev/openScale/wiki/Supported-scales-in-openScale)...

erijo:
Interesting! What does a search for your scales MAC address return?

Issue #82, opened 2017-10-16
Feature request - MQTT support


KavajNaruj:
Hi, first i want to thank you for this great app, definitely better than original [Health Coach](https://play.google.com/store/apps/details?id=de.sanitas_online.healthcoach&hl=en)

I'd like to ask if is possible to add support to publish data after measurement to [MQTT broker](https://github.com/mqtt/mqtt.github.io)? I am not fan of regularly exporting. In addition to backups it can add other uses such as TTS to the BT speaker, or send data to Influx and visualize them in Grafana

oliexdev:
I am not a big fan to send health data to any public clouds for privacy reasons (one reason why I started openScale, see [here](https://github.com/oliexdev/openScale/wiki/Custom-Bluetooth-Scale#how-its-started-)) but of course you are free to do it. 
If you want you can export the data and write a script to send it via MQTT protocol. Another solution  is to write for e.g. lets call it "openScale Connector" app which listen and access the internal sql database of openScale and send any new data via MQTT or to any company clouds (GARMIN,Google Fit and so on). I am willing to maintain and help this new "connector app" but the cloud implementations must come from the open source community.

KavajNaruj:
No clouds! I am running MQTT broker on LAN and I'd like to have a backup on NAS locally .

The MQTT is much better than SQL. If you subscribe for "topic" like lets say "openScale/user/weigth" you get values in real time, instead of repetable asking SQL database for changes.

So that kind of application that you mention (GARMIN,Google Fit and so on) will be esier to write, but that is not my consern.

Acually my plan is to connect openScale to my [HomeAssistant](https://home-assistant.io) instance and use it for some cool automation, like get praise when I lose some weigth an so on.

I think the best solution for this is to put some MQTT config ti openScale settings like:
- Host
- Port
- User
- Password

And after measurement send data to MQTT topics trougth [MQTT client](https://github.com/eclipse/paho.mqtt.android)

oliexdev:
ok send a PR with your implementation. Either I will create a connector app as describe above or I will integrate it directly into openScale.

PR #304, opened 2018-07-04
Add a ContentProvider so other apps can access (with permission) the …

funkwit:
…openScale data.

This could be used for e.g. syncing weight logs to a third-party service such as Google Fit,
without the openScale app having to do so; with this, a 'connector' app could be written that
reads this data and does the push to a third-party service (or, indeed, an HTTP RPC connector,
an MQTT connector, SMS notifications, or anything else) without openScale needing this
functionality (or additional permissions).

This is a straw-man implementation; happy to discuss how you think this should work, if this is a dumb idea altogether and I should abandon it, etc. Really a proof-of-concept as much as anything else. This is, though, basically the approach suggested by oliexdev at https://github.com/oliexdev/openScale/issues/82 - "Another solution is to write for e.g. lets call it "openScale Connector" app which listen and access the internal sql database of openScale and send any new data via MQTT or to any company clouds (GARMIN,Google Fit and so on)", and I think a ContentProvider is the most "Androidy" way to do this (and allows locking down this sensitive data behind permissions to prevent malicious apps doing so without user knowledge).

Here's how the flow looks from a dummy third-party app I hacked together which just dumps the data as text:
screenshot_1530710525
screenshot_1530710530
screenshot_1530710576

### DIFF
--- a/android_app/app/src/main/AndroidManifest.xml
+++ b/android_app/app/src/main/AndroidManifest.xml
@@ -9,6 +9,15 @@
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED"/>
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
 
+    <!-- Permission to allow read of data from the database through a ContentProvider.
+         Marked "dangerous" so that explicit user approval is required to read this data, not
+         just the permission implied from installing the app from the Play Store. -->
+    <permission
+        android:name="com.health.openscale.READ_DATA"
+        android:description="@string/permission_read_data_description"
+        android:label="@string/permission_read_data_label"
+        android:protectionLevel="dangerous" />
+
     <uses-feature android:name="android.hardware.bluetooth_le" android:required="false"/>
 
     <application
@@ -58,6 +67,13 @@
             <meta-data
                 android:name="android.support.FILE_PROVIDER_PATHS"
                 android:resource="@xml/file_provider_paths" />
+        </provider>
+        <provider
+            android:name=".core.export.OpenScaleContentProvider"
+            android:authorities="com.health.openscale.provider"
+            android:enabled="true"
+            android:exported="true"
+            android:readPermission="com.health.openscale.READ_DATA">
         </provider>
     </application>
 
--- a/android_app/app/src/main/java/com/health/openscale/core/database/ScaleUserDAO.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/database/ScaleUserDAO.java
@@ -21,6 +21,7 @@ import android.arch.persistence.room.Delete;
 import android.arch.persistence.room.Insert;
 import android.arch.persistence.room.Query;
 import android.arch.persistence.room.Update;
+import android.database.Cursor;
 
 import com.health.openscale.core.datatypes.ScaleUser;
 
@@ -45,4 +46,13 @@ public interface ScaleUserDAO {
 
     @Delete
     void delete(ScaleUser user);
+
+    // selectAll() and select() are equivalent to getall() and get(), but return a Cursor,
+    // for exposing via a ContentProvider.
+    @Query("SELECT * FROM scaleUsers")
+    Cursor selectAll();
+
+    @Query("SELECT * FROM scaleUsers WHERE id = :id")
+    Cursor select(int id);
+
 }
--- a/android_app/app/src/main/java/com/health/openscale/core/export/OpenScaleContentProvider.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/export/OpenScaleContentProvider.java
@@ -1,0 +1,129 @@
+/* Copyright (C) 2018 Paul Cowan <paul@custardsource.com>
+ *
+ *    This program is free software: you can redistribute it and/or modify
+ *    it under the terms of the GNU General Public License as published by
+ *    the Free Software Foundation, either version 3 of the License, or
+ *    (at your option) any later version.
+ *
+ *    This program is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *    GNU General Public License for more details.
+ *
+ *    You should have received a copy of the GNU General Public License
+ *    along with this program.  If not, see <http://www.gnu.org/licenses/>
+ */package com.health.openscale.core.export;
+
+import android.content.ContentProvider;
+import android.content.ContentUris;
+import android.content.ContentValues;
+import android.content.Context;
+import android.content.UriMatcher;
+import android.database.Cursor;
+import android.net.Uri;
+import android.text.TextUtils;
+import android.util.Log;
+
+import com.health.openscale.core.OpenScale;
+
+/**
+ * Exposes the user and measurement data from openScale via
+ * <a href="https://developer.android.com/guide/topics/providers/content-providers">Android
+ * Content Providers</a>. This allows other apps to access the openScale data for their own purposes
+ * (e.g. syncing to third-party services like Google Fit, Fitbit API, etc) without openScale itself
+ * needing to do so or request additional permissions. <br />
+ *
+ * This access is gated by the com.health.openscale.READ_DATA permission, which is defined in the
+ * manifest; it is not accessible to any other app without user confirmation.<br />
+ *
+ * The following URIs are supported:
+ * <ul>
+ *     <li><code>content://com.health.openscale.provider/user</code>: list all users.</li>
+ *     <li><code>content://com.health.openscale.provider/user/$ID</code>: retrieve single user
+ *         by ID.</li>
+ *     <li><code>content://com.health.openscale.provider/user/$ID/measurements</code>:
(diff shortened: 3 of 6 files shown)


oliexdev:
@funkwit thanks for your PR. Can you provide us your dummy third party app, so I can look deeper how it works. Thanks in advance.

funkwit:
Hey @oliexdev - unfortunately just for the sake of quickly throwing it together, I've done the demo app on top of an existing app framework that isn't open-source, so giving you the app itself will take some work. But that said, there's only about 20 lines involved so I'll pop them here:

This goes in the manifest of the client app:
` <uses-permission android:name="com.health.openscale.READ_DATA" />`

The permissions check is done like this, in onCreate():
```java
    private static final String REQUIRED_PERMISSION = "com.health.openscale.READ_DATA";

    public static final int REQUEST_CODE = 1;

        if (ActivityCompat.checkSelfPermission(getApplicationContext(),
                REQUIRED_PERMISSION) != PackageManager.PERMISSION_GRANTED) {
            // We don't have permission, give the user a permission button:
            findViewById(R.id.permissionsButton).setOnClickListener(new View.OnClickListener() {
                public void onClick(View v) {
                    ActivityCompat.requestPermissions(PermissionsActivity.this,
                            new String[]{REQUIRED_PERMISSION},
                            REQUEST_CODE);
                }
            });
        }
```

Then retrieving the data and building that debug output is literally just this:

```java
        Cursor c = getContentResolver().query(Uri.parse("content://com.health.openscale.provider/user"), null, null, null, null);
        StringBuilder s = new StringBuilder();
        s.append("=========== iterate all users ============" + System.lineSeparator());
        try {
            while (c.moveToNext()) {
                s.append("------ new record ------" + System.lineSeparator());
                for (int i = 0; i < c.getColumnCount(); i++) {
                    s.append(c.getColumnName(i) + ": " + c.getString(i) + System.lineSeparator());
                }
            }
        } finally {
            c.close();
        }
```

That's basically the entire relevant code of that app :)

erijo:
I like this idea and it's how I was planning to add some kind of Google fit integration. Thanks for taking the time doing it!

My main concern has to do with exposing the internal format of the database. It seems like that could tie our hands a bit if we wanted to change it (and don't break other apps using our data). Do you have any input on that? 

Do you know how others do? Convert the format to an external stable one? Include a version in the export?

android_app/app/src/main/java/com/health/openscale/core/database/ScaleUserDAO.java
```diff
@@ -45,4 +46,13 @@
 
     @Delete
     void delete(ScaleUser user);
+
+    // selectAll() and select() are equivalent to getall() and get(), but return a Cursor,
+    // for exposing via a ContentProvider.
+    @Query("SELECT * FROM scaleUsers")
```
erijo: SELECT id FROM scaleUsers? Perhaps also rename to selectAllIds.

android_app/app/src/main/AndroidManifest.xml
```diff
@@ -9,6 +9,15 @@
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED"/>
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
 
+    <!-- Permission to allow read of data from the database through a ContentProvider.
+         Marked "dangerous" so that explicit user approval is required to read this data, not
+         just the permission implied from installing the app from the Play Store. -->
+    <permission
+        android:name="com.health.openscale.READ_DATA"
```
erijo: Can this be just ".READ_DATA"? The dev version built by Travis adds ".dev" to the application id.

android_app/app/src/main/java/com/health/openscale/core/export/OpenScaleContentProvider.java
```diff
@@ -0,0 +1,129 @@
...
+
+    private static final String AUTHORITY = "com.health.openscale.provider";
+
+    private static final int MATCH_TYPE_USER_LIST = 1;
+    private static final int MATCH_TYPE_USER_ENTRY = 2;
+    private static final int MATCH_TYPE_USER_MEASUREMENTS = 3;
+
+
+    static {
+        uriMatcher.addURI(AUTHORITY, "user", MATCH_TYPE_USER_LIST);
```
erijo: I think this should be "users" instead of "user".

android_app/app/src/main/java/com/health/openscale/core/export/OpenScaleContentProvider.java
```diff
@@ -0,0 +1,129 @@
...
+ *     <li><code>content://com.health.openscale.provider/user/$ID</code>: retrieve single user
+ *         by ID.</li>
+ *     <li><code>content://com.health.openscale.provider/user/$ID/measurements</code>:
+ *         retrieve all measurements for the supplied user ID.</li>
+ * </ul>
+ */
+public class OpenScaleContentProvider extends ContentProvider {
+    private static final UriMatcher uriMatcher = new UriMatcher(UriMatcher.NO_MATCH);
+
+    private static final String AUTHORITY = "com.health.openscale.provider";
```
erijo: Use BuildConfig.APPLICATION_ID instead of hard coding "com.health.openscale". Here and in all other places.

android_app/app/src/main/java/com/health/openscale/core/export/OpenScaleContentProvider.java
```diff
@@ -0,0 +1,129 @@
...
+
+import android.content.ContentProvider;
+import android.content.ContentUris;
+import android.content.ContentValues;
+import android.content.Context;
+import android.content.UriMatcher;
+import android.database.Cursor;
+import android.net.Uri;
+import android.text.TextUtils;
+import android.util.Log;
```
erijo: Seems unused, please remove. If logging is needed, please use Timber.

android_app/app/src/main/java/com/health/openscale/core/export/OpenScaleContentProvider.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        cursor.setNotificationUri(context.getContentResolver(), uri);
+        return cursor;
+    }
+
+
+    @Override
+    public int delete(Uri uri, String selection, String[] selectionArgs) {
+        throw new UnsupportedOperationException("Not yet implemented");
```
erijo: Perhaps "Not supported" would be better (here and below).

erijo:
Regarding the exposure of the database, perhaps the easiest and best solution would be to add some kind of meta path that returns the db_version (and perhaps the openScale version as well) and leave it to the external application to keep up to date with any incompatible changes in the database format?

funkwit:
Yeah, I have been thinking about the stable data format issue, and agree this is tricky. Exposing a version is one option, but it's messy (do you support previous versions always? URLs something like /v1/user/100? You can do that, but then you end up with a spaghetti of input/output formats and conversions and awfulness. And just replacing doesn't really help; if a client app does a fetch of /db/version and gets "6", and they only know how to deal with 5... they can't do anything, anyway, apart from 'throw an error' or 'try it anyway and hope the changes are backwards-compatible (adding columns, removing ones they don't use anyway?)', which isn't much different to them not checking the version in the first place.

More realistically, I think you probably just commit to an external API, and then ONLY make backwards-compatible changes. This is a really hard problem and, hey, books have been written on it…

For this *specific* case, maybe it's sufficient to declare the absolute minimum API that it seems unlikely will ever NOT be supported?

e.g. at the moment I have

```
@Query("SELECT * FROM scaleUsers WHERE id = :id")
Cursor select(int id);
```

instead, maybe it makes more sense to trim this down to a 'supportable for the forseeable future' set of data, which is enough for e.g. the Fit syncing:

```
@Query("SELECT id, username FROM scaleUsers WHERE id = :id")
```

Because it seems extremely unlikely that id and username will go anywhere in future versions (goalWeight and goalDate might go into different tables later, or be renamed, but id and username seem safe; worst case you might later have to do ```SELECT userID AS id, some_new_column_name AS username ...``` (or whatever) to deal with renamed columns? So expose the bare minimum data which you're happy to commit to, and then other columns later become a choice of 'how much do we need this for the new use case, and how willing are we to commit to this being part of the API forever?'

For the Fit (etc) sync case, I think:
```
scaleUsers: id, username
scaleMeasurements: userid, datetime, weight, fat
```
is probably the minimal set, and is (probably) something that's easy to commit to always supporting in some way or another?

oliexdev:
Thanks @funkwit I merged your PR into the master repository (just renamed and moved the ContentProvider class file)

I follow with @erijo comments on your code. We need to have flexibility between the dev and release version of openScale.

Apply our changes in database formats should be in the responsible of the third party Apps. But to handle backward compatible I would also say to publish the database version via the content provider.

Your idea to expose only a minimum set sounds interesting but I am sure we have then many request to grant access to the full database.

Now, should we launch a second project "openScale Connector" App or do we let the community write some third party apps? Also I am not sure which license is the best because some health libraries doesn't allow open source Apps or we only support free libraries like MQTT and so on.

erijo:
I did some changes to how and what data we expose on https://github.com/erijo/openScale/commits/connector

Also added a simple connector app to be able to test the feature. Dev build here: https://github.com/erijo/openScale/releases/tag/travis-erijoconnector-build

PR #328, opened 2018-09-24
Refactor content provider support and add simple connector app

erijo:
Rework the content provider to only expose a small subset of data. A subset that hopefully never changes. Also change the way the data is provided (URI format) so that some help classes can be used.

A small and very basic "connector" app is also created to be able to test the provider. It is auto built by travis (see https://github.com/erijo/openScale/releases/tag/travis-erijoconnector-build).


oliexdev:
Thanks @erijo looks good but could you create a PR request without the connector app. I don't want to mix the connector with the openScale repository. If it for an example app then we could add it in the wiki or create another example project.

erijo:
I cherry-picked the parts that didn't involve the connector app and pushed them (0cc9d0fd and cb348f70). Will look into adding the api version we discussed over mail as well.

erijo:
Api version added in 12cb1d0fafeda61031ff250a3180819de5ff6836.

PR #333, opened 2018-10-07
Add support for Trisa Body Analyze 4.0 smart scale.

maksverver:


oliexdev:
thanks @maksverver looks good. I send you an e-mail with some information.

oliexdev:
@maksverver could you provide an image of your scale for our wiki gallery? (in the same angle)
I can't use images from the internet due copyright reasons.

maksverver:
I've added a photo and sent another pull request.

By the way, what is meant by "history data" as a scale feature on the wiki? If it means you can extract past measurements, then the Trisa scale supports this. However, once the measurements are downloaded, they are gone from the device (i.e., you cannot download them again with a different app). For example, if you weigh yourself every day of the week and only connect the app on Sunday, it will retrieve measurements from the past seven days (with correct timestamps).

android_app/app/src/main/java/com/health/openscale/core/bluetooth/lib/TrisaBodyAnalyzeLib.java
```diff
@@ -0,0 +1,92 @@
...
+
+    // Timestamp of 2010-01-01 00:00:00 UTC (or local time?)
+    private static final long TIMESTAMP_OFFSET_SECONDS = 1262304000L;
+
+    /**
+     * Converts 4 little-endian bytes to a 32-bit integer, starting from {@code offset}.
+     *
+     * @throws IndexOutOfBoundsException if {@code offset < 0} or {@code offset + 4> data.length}
+     */
+    public static int getInt32(byte[] data, int offset) {
```
erijo: Couldn't Converters.fromUnsignedInt32Le be used instead?

android_app/app/src/main/java/com/health/openscale/core/bluetooth/lib/TrisaBodyAnalyzeLib.java
```diff
@@ -0,0 +1,92 @@
...
+
+    /** Converts 4 bytes to a floating point number, starting from  {@code offset}.
+     *
+     * <p>The first three little-endian bytes form the 24-bit mantissa. The last byte contains the
+     * signed exponent, applied in base 10.
+     *
+     * @throws IndexOutOfBoundsException if {@code offset < 0} or {@code offset + 4> data.length}
+     */
+    public static double getBase10Float(byte[] data, int offset) {
+        int mantissa = (data[offset] & 0xff) | ((data[offset + 1] & 0xff) << 8) |
```
erijo: And this could use fromUnsignedInt24Le.

PR #342, opened 2018-10-13
Improve BluetoothTrisaBodyAnalyze

maksverver:
This pull request covers two changes:

  1. I added calculation of body composition data for the Trisa Body Analyzer scale, based on logic from the reverse-engineered Trisa Vital Bluetooth app. The reliability of these figures is questionable, but it's better than nothing.

  2. I expanded Converters.java, adding support for decoding signed integers, and encoding to an existing byte array. Using these functions in the Trisa Body Analyze driver allowed me to remove some of the custom encoding/decoding logic there.


erijo:
Thanks for your contribution! I removed some unused imports (from here and earlier) and merged your changes.

PR #363, opened 2018-12-14
Add Medisana BS430 to list of supported scales

vershinin:
Hi! I had reversed BS430 protocol in order to check if it works the same way as BS440. My investigations show that it's definitely true, however time offset is not required. I'd like to contribute this PR .


oliexdev:
thanks for your contribution :+1:

PR #367, opened 2018-12-20
FitIndex -  QNScale

dockerss:
Support for mazon.com/Bluetooth-FITINDEX-Wireless-Bathroom-Composition/dp/B078HW8BPY FitIndex -
 QNScale as provided by https://github.com/srwareham/openScale/blob/feature_qn_scale/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothQNScale.java

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothFactory.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothFactory.java
@@ -99,6 +99,9 @@ public class BluetoothFactory {
                 || deviceName.equals("042FatScale01")) {
             return new BluetoothInlife(context);
         }
+        if (deviceName.startsWith("QN-Scale")) {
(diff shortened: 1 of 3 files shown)


srwareham:
@dockerss I'm quite glad a PR was submitted for this as I haven't been able to get back to this for a good while.

@oliexdev Not to be a nuisance, but is there a way I can be properly attributed as a contributer for writing this? Maybe some token PR re-adding the file?

Thanks

srwareham:
Also, as a heads up, my code definitely has room for improvement. Most notably, I never finished reverse engineering some of the magic numbers--I'm pretty sure some of what I have forces metric and there may be some userID specific stuff as well. I was also in the process of implementing a method to obtain a steady-state weight rather than just the chronologically first (I haven't looked at the cleanups by @oliexdev, so it's possible some of this has already been resolved).

Issue #313, opened 2018-08-06
Add overall overview


natrius:
Add an overview over all input data. It would be nice to see the whole picture over a longer period of time. On the top right is the option for weekly and monthly, add new one for "overall" for example.

And a good one would be "defined range" so if i want to show someone something special that would be nice too.

sik0vny:
I agree on this one.

oliexdev:
please test the latest openScale branch build of [MPAndroidChart](https://github.com/oliexdev/openScale/releases/tag/travis-dev_MPAndroidChart-build)

PR #371, opened 2018-12-27
Replaced HelloCharts with MPAndroidChart library

oliexdev:
I replaced the [HelloCharts](https://github.com/lecho/hellocharts-android) library with the [MPAndroidChart](https://github.com/PhilJay/MPAndroidChart) library. The main reason was that the HelloChart library are not maintained anymore.

With the new library I implemented also some nice features:
- Rolling chart on the overview, see issue #313 (the last 14 days are shown but you can scroll all way back to the first measurement)
- replaced statistic tables for average week / month with a radar graph
- measurements can now be set on the left or right Y-axis
- now a marker display a value on a value click
- a "value menu" is displayed in graph page for editing, show and delete a measurement 
- display a graph legend (with an option in the settings to disable it)
- on the overview page the icons are coloured
- a marked value is now highlighted
- animation for the graph (graphs are rolled in from below)

Some improvements ideas or questions are:
- show the average symbol also in the marker if the value is an average value
- should we show the Y-axis scale on both sides?
- do we need the "value menu" on the overview page too?
- another question which is related with the previous question, do we need an edit button on the overview page?
- is there a better place for the "value menu"?
- should we add more radar graphs on the statistic page for the other measurements?
- show more information on the marker view (e.g. date, difference since last measurement)

All in all, while I was refactoring I was really happy that we have a good source code quality and software design approach, which makes the refactoring quite easy. 

@erijo can you please review the PR when you have time? I hope I have catched all bugs :wink:


erijo:
I did a quick test of the branch and it's definitely a big step in the right direction. Some initial comments:
- without both axis and data label it's hard to make sense of the graph without clicking on a measurement
- I don't understand the radar graph

oliexdev:
> without both axis and data label it's hard to make sense of the graph without clicking on a measurement

I added an option to enable/disable the y-axis

> I don't understand the radar graph

It's overlap the last measurement with the average of the last weekend. If they are the same you don't see anything. If they are different you should see a difference.

> it would be nice if the regular graph was scrollable like the overview

No, I don't want to change the graph to a scrollable graph. You can better select the month / year / days in this way and you don't have to scroll it.

Cathryne:
Thank you very much for this :-) The rolling chart is sooo useful!

PR #375, opened 2019-01-11
QNScale additional attributes support

dockerss:
QNScale additional attributes support


oliexdev:
You use the same formulas like the Trisa body analyze library. 
Did you test the measurements with the original app? Do you got the same results? 
I refactored with commit https://github.com/oliexdev/openScale/commit/4dfdb233f4c88ddbdd142a81cbae6ff0ca454b9c the `TrisaBodyAnalyzeLib.class` so you can use it easily instead of duplicating the formulas. Can you refactor your code accordingly?

dockerss:
I am getting near similar values. Will close this PR and open a new PR with the changes to use TrisaBodyAnalyzeLib

PR #385, opened 2019-01-19
Added support for Libra-W

LaurentMox:
My Runtastic/Adidas Libra shows as "Libra-W", and was marked as non-supported in openScale. I tried using this tiny patch and everything is working great now!


oliexdev:
thanks for your PR @LaurentMox

PR #396, opened 2019-02-05
Change color depending on if you are going towards or away from your goal

### LINKED ISSUE #395: Widget arrow colours should match progression towards goal
Currently it seems when you gain weight the little arrow indicating weight difference is always green, and when losing weight is red.

This should instead be green if you've progressed towards your goal and red if you've gone away from your goal

RichyHBM:
Fixes #395
This sets the arrow color on the widget to be green for positive change towards your goal and red for negative


oliexdev:
@RichyHBM thanks for your PR. I implemented with https://github.com/oliexdev/openScale/commit/56cd6004ec55ae166d2ccd245e3554c256667190 a more simple version of your implementation.

PR #421, opened 2019-02-27
fix average array size and index

unimiso:
fix average array size in order to allocate 0...(maxXValue-minXValue)
and convert to 0-origin index from binNR(is a value of range minXValue <=> maxXValue).


oliexdev:
thanks for your PR :1st_place_medal:

oliexdev:
had to revert your PR on the week view and rolling chart on I had some index out of bounce exceptions

PR #429, opened 2019-03-06
Fix force QN-Scale to KG

### LINKED ISSUE #428: QN-Scale Forces KG on Scale Display
**Describe the bug**
Connecting to a QN-Scale scale forces the scale to display on-scale measurements KG. This will overwrite any settings for previously setting the units to LBs through the vendor app.

**To Reproduce**
1. Step on scale without pairing (with units set to LB from vendor app)
2. Connect to scale using openScale and weigh self
3. Observe scale units now set to KG

Reproduced with v2.0.1 (42)

**Expected behavior**
The scale should use whichever units are set in the ScaleUser's preferences.

**Additional context**
N/A

**Debug log**
Not provided

srwareham:
Fixes #428. The magic number sent to initiate weight readings contains a byte describing which units the scale should be set to. The previous implementation hard coded this as 0x01 for KG. This PR switches the unit (either LB or KG) to be based on the value retrieved by `OpenScale.getInstance().getSelectedScaleUser().getScaleUnit()`

Please note that my ability to build the apk is currently down and I am unable to test this (although the code is extremely simple, I don't think it should be a problem). Hoping Travis will accomplish this.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothQNScale.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothQNScale.java
@@ -85,6 +85,13 @@ public class BluetoothQNScale extends BluetoothCommunication {
 
     @Override
     protected boolean onNextStep(int stepNr) {
+    final ScaleUser scaleUser = OpenScale.getInstance().getSelectedScaleUser();
+    final Converters.WeightUnit scaleUserWeightUnit = scaleUser.getScaleUnit();
+    // Default weight unit KG. 0x01 weight byte = KG. 0x02 weight byte = LB.
+    byte weightUnitByte = (byte) 0x01;
+    if (scaleUserWeightUnit == Converters.WeightUnit.LB){
+        weightUnitByte = (byte) 0x02;
+    }
         switch (stepNr) {
             case 0:
                 // set notification on for custom characteristic 1 (weight, time, and others)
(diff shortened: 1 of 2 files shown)


srwareham:
Finally sorted out my build issues and unfortunately the current implementation doesn't work when the user is set to pounds. Will hopefully have time soon to fix this.

srwareham:
The code is now working, the only thing missing was updating the checksum byte at the end. I have tested the apk, which now has QN-Scale respect the weight unit choice in the openScale user's preferences (KG is used as a default).

Should be good to pull! Thanks!

android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothQNScale.java
```diff
@@ -85,6 +85,16 @@ public String driverName() {
...
+    final ScaleUser scaleUser = OpenScale.getInstance().getSelectedScaleUser();
+    final Converters.WeightUnit scaleUserWeightUnit = scaleUser.getScaleUnit();
+    // Value of 0x01 = KG. 0x02 = LB. Requests with stones unit are sent as LB, with post-processing in vendor app.
+    byte weightUnitByte = (byte) 0x01;
+    // Checksum at end of message. For now, simply switching based on weight unit only
+    byte weightRequestChecksum = (byte) 0x42;
+    if (scaleUserWeightUnit == Converters.WeightUnit.LB){
+        weightUnitByte = (byte) 0x02;
+        weightRequestChecksum = (byte) 0x43;
+    }
```
oliexdev: move this whole part into the step nr 2

android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothQNScale.java
```diff
@@ -95,8 +105,9 @@ protected boolean onNextStep(int stepNr) {
                 setIndicationOn(CUSTOM2_MEASUREMENT_CHARACTERISTIC);
                 break;
             case 2:
-                // write magicnumber 0x130915011000000042 to 0xffe3
-                byte[] ffe3magicBytes = new byte[]{(byte) 0x13, (byte) 0x09, (byte) 0x15, (byte) 0x01, (byte) 0x10, (byte) 0x00, (byte) 0x00, (byte) 0x00, (byte) 0x42};
+                // write magicnumber 0x130915[WEIGHT_BYTE]1000000042 to 0xffe3
+                // 0x01 weight byte = KG. 0x02 weight byte = LB.
+                byte[] ffe3magicBytes = new byte[]{(byte) 0x13, (byte) 0x09, (byte) 0x15, weightUnitByte, (byte) 0x10, (byte) 0x00, (byte) 0x00, (byte) 0x00, weightRequestChecksum};
```
oliexdev: please use the `xorChecksum(...)` method for calculating the checksum

android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothQNScale.java
srwareham: Code moved as requested.

android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothQNScale.java
srwareham: I added a new method in `BluetoothCommunication.java` named `sumChecksum`, as this scale uses a normal sum rather than xorChecksum.

srwareham:
Changes implemented and testing confirms the scale performs as expected

oliexdev:
thanks but strange that they use a sum for the checksum!?

PR #439, opened 2019-03-17
1byone ignore duplicate data

oliexdev:


oliexdev:
will ignore this PR since there are no problems reported anymore with duplicate entries

PR #453, opened 2019-03-31
Support Yunmai Mini 2 China version

xcffl:
Since Yunmai Mini 2 China version uses bluetooth name "YUNMAI-ISM2T", the current matching rules does not work. This patch makes it work.


oliexdev:
Thanks for your PR

Issue #477, opened 2019-07-03
Xiaomi Body Fat Scale v2


lesavonfou:
Hello everyone,
I have bought a Xiaomi Body Fat Scale 2 recently and it is not recognized by openScale.
aliexpress link: www.aliexpress.com/store/product/-/1872811_33008790826.html

The scale shows up as MIBFS, here is the log:
```
2019-04-30 11:50:24.552 Debug [2] AboutPreferences: Debug log enabled, openScale v2.0.3 (44), SDK 28
2019-04-30 11:50:29.446 Debug [2] OpenScale: Disconnecting from bluetooth device
2019-04-30 11:50:29.447 Debug [2] BluetoothCommunication: Bluetooth disconnect
2019-04-30 11:50:44.108 Debug [2] OpenScale: Trying to connect to bluetooth device [CE:8E:E5:5B:58:B6] (MIBCS)
2019-04-30 11:50:44.112 Debug [2] BluetoothCommunication: Do LE scan before connecting to device
2019-04-30 11:50:44.115 Debug [2] RxBle#ClientOperationQueue: QUEUED   ScanOperationApi21(54351201)
2019-04-30 11:50:44.116 Debug [2401] RxBle#ClientOperationQueue: STARTED  ScanOperationApi21(54351201)
2019-04-30 11:50:44.119 Info [2407] RxBle#QueueOperation: Scan operation is requested to start.
2019-04-30 11:50:44.139 Debug [2401] RxBle#ClientOperationQueue: FINISHED ScanOperationApi21(54351201) in 22 ms
2019-04-30 11:50:44.298 Debug [2] BluetoothCommunication: Stop Le san
2019-04-30 11:50:44.300 Info [2407] RxBle#CancellableDisposable: Scan operation is requested to stop.
2019-04-30 11:50:45.305 Debug [2] BluetoothCommunication: Try to connect to BLE device CE:8E:E5:5B:58:B6
2019-04-30 11:50:45.310 Debug [2] BluetoothCommunication: Step Nr 0
2019-04-30 11:50:45.315 Debug [2] BluetoothCommunication: Invoke write bytes [06 04 00 01] on 00001542-0000-3512-2118-0009af100700
2019-04-30 11:50:45.332 Debug [2] BluetoothCommunication: Step Nr 1
2019-04-30 11:50:45.334 Debug [2411] RxBle#ClientOperationQueue: QUEUED   ConnectOperation(46091349)
[...]

PR #483, opened 2019-07-14
Add MIBFS identifier to bluetooth factory

lesavonfou:
After some investigations this weekend, I managed to make my scale recognizable.
The app seems to work normally, but I would like to take some time to validate that everything is alright.

Thus, if you could let that merge request open for a week or two, it would be great ! :D


lesavonfou:
#477

oliexdev:
why you do a PR then? And please provide the openScale log file as I wrote in issue #477 it seems your scale name is still MIBCS

PR #505, opened 2019-10-07
Additional Medisana BS444 BT Identifier

dhuizinga:
Added additional BT identifier/name for Medisana BS444 model (currently unrecognized in openScale Pro app), not entirely sure whether the current code works, please check.


dhuizinga:
image

oliexdev:
I don't own this scale? Does this scale works with the Medisana BS444 code?

dhuizinga:
it is the same scale, just an additional range of BT addresses :)

oliexdev:
thx

dhuizinga:
Just for documenting's sake: It now works like a charm

PR #517, opened 2019-11-23
Add Support for Vigorun scale

wiomoc:
https://www.amazon.co.uk/Vigorun-Bluetooth-Electronic-Measurement-Indicators/dp/B07F112PXG


oliexdev:
thanks does your Vigorun scale also works with WRITE_TYPE_DEFAULT? Because it might be that other scales will not work with WRITE_TYPE_NO_RESPONSE?
Also it seems that your scale is a sub-brand of MGB scale, we don#t need to highlight it in the class name.

wiomoc:
The Vigorun scale requires `WRITE_TYPE_NO_RESPONSE` this is the only difference from the MGB scale

oliexdev:
I hope that the MGB scale still work with WRITE_TYPE_NO_RESPONSE but we will see.

PR #522, opened 2019-12-02
Add YUNMAI LBM & visceral fat value.

d3m3vilurr:
- LBM: scale machine just use `weight - fat` formula
- vis fat: function value range is 1 to 30, if openScale want to diff range value, we should rearrange it. also original app applied `floor` calculation to return value. but I didn't apply it.


oliexdev:
awesome thanks, good reverse engineering :1st_place_medal:

PR #534, opened 2020-01-01
Update to SENSSUN scale Bluetooth code

uyllii:
I was having an issue with a SENSSUN scale and the openScale app where the app would only write a new record about 20% of the time. 

The scale seemed to ignore the user synchronization message and the app would sit waiting for the measurement data (fat/water...) which never arrived. 

This change resends the userSynchronisation message to the scale (after stabilization) until it responds with the other metrics. 

This change has my scale responding correctly without fail since I changed the code.


oliexdev:
As I don't own this scale, what happened if you weight yourself for example with socks on? (So you don't get the body measurements). Do you get multiply measurements or no measurements?

uyllii:
Only one line per reading. All the measurements are zero (except weight). The play store version works still only works about 1 in five times socks or bare feet.

PR #539, opened 2020-01-16
Spelling: Clarifications and shortened sentences

comradekingu:
I didn't change the strings for this yet, as I wanted to not make it seem like there was no consideration behind it:

"Open-source" is not the OSI-trademarked term, so it effectively only means freedom 1.
In practice, the term does not translate one way or the other, so it will still only communicate only freedom 1. (That is selling it short by a lot).
What's worse, translators get it wrong, in much the exact way it is wrong already, one way or the other, so now we have "Open-source" verbatim in German and Dutch, where it does not belong. In Spanish it is translated as "source that is open". That means there are at-least 3 different variants communicated.
https://hosted.weblate.org/translate/openscale/strings/en/?checksum=eac7f35c57cf4597

"Free software" is possibly meaningless in English now that people don't distinguish freeware from it. It seems more like it is "gratis", because of the duality. (This sells it short even more). It does however work in other languages. (Frei in German etc.)

"Libre" works in both English and does not get translated erroneously into other languages.

Moreover, openScale actually also has freedom 4, as it is copyleft. "Copylefted libre software" communicates something (4) beyond libre software (0-3), the same way libre (still 0-3) is better than opened (1).

0 the freedom to run the program as you wish
1 the freedom to study the program
2 the freedom to change it
3 the freedom to share it
4 the reciprocal requirement that necessitates sharing

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ https://travis-ci.org/oliexdev/openScale)
 [![Translation status](https://hosted.weblate.org/widgets/openscale/-/strings/svg-badge.svg)](https://hosted.weblate.org/engage/openscale/?utm_source=widget)
 =========
 
-Open-source weight and body metrics tracker, with support for Bluetooth scales
+Copylefted libre software weight and body metrics tracker, with support for Bluetooth scales
 
 <a href="https://f-droid.org/repository/browse/?fdid=com.health.openscale" target="_blank">
 <img src="https://f-droid.org/badge/get-it-on.png" alt="Get it on F-Droid" height="80"/></a>
@@ -14,15 +14,15 @@ Install [openScale-dev-build.apk](https://github.com/oliexdev/openScale/releases
 
 # Summary :clipboard:
 
-Monitor and track your weight, BMI, body fat, body water, muscle and other body metrics in an open source app that:
-* has an easy to use user interface with graphs,
-* supports various Bluetooth scales,
-* doesn't require you to create an account,
-* can be configured to only show the metrics you care about, and
-* respects your privacy and lets you decide what to do with your data.
+Use the app to monitor and track your weight, BMI, body fat, body water, muscle and other body metrics:
+* Easy to use user interface with graphs.
+* Support for various Bluetooth scales.
+* No account required.
+* Show only the metrics you care about.
+* Privacy and the ability to do what you want with _your_ data.
 
 # Supported Bluetooth scales :rocket:
-openScale has built-in support for a number of Bluetooth (BLE or "smart") scales from  many manufacturers, e.g. Beurer, Sanitas, Yunmai, Xiaomi, etc. (see model list below). Together with our users we constantly improve and extend the set of supported scales and in many cases pick up where the original app falls short.
+Built-in support for a number of Bluetooth (BLE or "smart") scales from  many manufacturers. Beurer, Sanitas, Yunmai, Xiaomi, etc. (listed below). Together with our users we constantly improve and extend the set of supported scales and in many cases pick up where the original app falls short.
 
 - Custom made Bluetooth scale
 - Beurer BF700, BF710 and BF800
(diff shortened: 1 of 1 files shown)


oliexdev:
I know there is a terminology discussion between open-source and libre software since the beginning of libre/open-source software but in my opinion open-source is widely more used in press and understandable by normal users. I don't know what you mean that open-source is not a trademark. You can read the OSI definition [here](https://opensource.org/osd-annotated).
I think your suggestion "Copylefted libre software" is more irritating than helping. They even introduced [FLOSS/FOSS ](https://en.wikipedia.org/wiki/Free_and_open-source_software) as a new word to combine these two words but again for the normal users open-source is more understandable, so for me I want to stick to it.

Your shortenings in the sentences sounds for me, how can I say that, not natural anymore!? But if you found grammar mistakes I am happy to correct them.

comradekingu:
@oliexdev The terminology problem arose in 1998, as "Open Source" entered the fray, though the problem of "Free/gratis" was always there with "Free software" from the early 80s onwards. https://en.wikipedia.org/wiki/Gratis_versus_libre We still have the problem, "copyleft" and "libre" being distinct and meaningful concepts to that effect. You will find https://weblate.org itself says so up front and center.

Going the other way, because "Open Source" alone is problematic, "FLOSS"/"FOSS" OTOH is an amalgam of compromise arising in writing about libre software in academia.
For the same reason the economy of words caters to shorter sentences, less text doing the same is important, _especially_ in software.
The trademarked business-word "Open Source" was specifically adopted to _not be understandable_, as it focuses on only one freedom, importantly hiding the others. That is the whole point of the OSI. The OSI definition is just the DFSG definition rewritten to the effect of changing the word "free software" for "Open Source".

The fairly new usage of "Libre" (to mean freedom 0-3) is fortunately unencumbered with any organisation, such as "OSI - Open Source" and "FSF - Free software"

The resulting effect of redundant text is https://en.wikipedia.org/wiki/Wikipedia:Wall_of_text
This induces stress for the reader, and the burden of reading it all is directly related to less adoption.

Further, it is a pain to translate (where I come in), and a lot harder to understand when arbitrary components of a sentence are added, especially in translation. I already showed how "open-source" fails being translated in openScale. Having now detailed the etymology and history, I don't know what more is needed. For normal users "Open Source" or "open-source", _at best_ isn't understood, with no promise of change anytime soon.
For structure, I could annotate all the changes if you like. A lot of thought went into this, it isn't at all random.

oliexdev:
I understand your point in an academic view but I have a little bit other view of it. If you do a simple test googling "open-source" vs "libre software" you get a high much more results. Even at least in Germany the media use open-source more widely. Additionally the project name **open**Scale is used to give a hint that this app is **open**-source. Currently I also see no translation problem with open-source if I see the weblate status of the project.

comradekingu:
@oliexdev It was never "open-source", which in neither variant has anything to do with German.

If you ask what "Open Source" means, you will get answers that don't have to do with other freedoms. Search for "Copyright", you will get more results than for "Copyleft", which is the argument of use turned in the favour of using the latter, but I take it we just disagree there too.

The popularity enjoyed by "Open Source" does not seem to have rubbed off on the once proud OpenOffice vis-a-vis LibreOffice, but if people don't understand what "Open Source" means, or why that isn't the part that matters, surely an allusion to the fact is what it will take.

What the media does is arguably a low bar to clear in 2020, but in the very least it doesn't spell things the wrong way…
https://www.heise.de/newsticker/meldung/Umfrage-Open-Source-wird-proprietaere-Software-in-Unternehmen-bald-ueberfluegeln-4667888.html
https://www.heise.de/newsticker/meldung/Solus-Version-4-1-aktualisiert-Software-und-unterstuetzt-zusaetzliche-Hardware-4649488.html

That you don't see it doesn't mean it doesn't exist. The idea that I would present you with all sorts of evidence, at every possible level, amounts to what?

The issue is present in German, Catalan and Spanish:
https://hosted.weblate.org/translate/openscale/strings/en/?checksum=eac7f35c57cf4597#translations https://hosted.weblate.org/translate/openscale/strings/en/?checksum=c5b184b614c1a527#translations https://hosted.weblate.org/translate/openscale/strings/en/?checksum=eac7f35c57cf4597

It is quite puzzling not accepting help rectifying objectively poor language, and arguing semantics over a smaller issue, from a position of being inconsistent in the matter.

PR #631, opened 2020-09-19
Remove duplicate line

naoliv:


oliexdev:
thanks its already removed in my local source-code

Issue #593, opened 2020-07-07
Insmart scale support


ChristianB86:
Insmart
[Android App](https://play.google.com/store/apps/details?id=cn.fitdays.fitdays&hl=en_US)
[Product Page](https://www.amazon.co.uk/Bluetooth-INSMART-Bathroom-Weighing-Composition/dp/B07Q9PS28M)
I can not found a webpage from the vendor

_To support a new scale it is necessary to gather some information._

**Step 1: Read the general reverse engineer process**

- [X] I have read the [how to reverse engineer a Bluetooth 4.x wiki](https://github.com/oliexdev/openScale/wiki/How-to-reverse-engineer-a-Bluetooth-4.x-scale)  entry

**Step 2: Acquiring some Bluetooth traffic**
_Attach 3 log files with the corresponding true values, read [here](https://github.com/oliexdev/openScale/wiki/How-to-reverse-engineer-a-Bluetooth-4.x-scale#1-acquiring-some-bluetooth-traffic) for further information._

1. Bluetooth HCI Snoop log file
user settings in the vendors app:
```
sex male, body height 193cm, age 35, activity level: not a setable value
```
measured true values in the vendors app for the 1. HCI Snoop log file:
```
89.60kg, 20:49 06/07/2020, water 59,4%, muscle 53,2% , fat 17,7% 
```
_--> Attach the 1. HCI Snoop log file here <--_
[06_07_2020__20_50_02-btsnoop_hci.log](https://github.com/oliexdev/openScale/files/4886763/06_07_2020__20_50_02-btsnoop_hci.log)

2. Bluetooth HCI Snoop log file
user settings in the vendors app:
```
same as first one
```
measured true values in the vendors app for the 2. HCI Snoop log file:
```
93.35kg, 20:50 06/07/2020, 58.3%, 52.2% , 19.2%
```

_--> Attach the 2. HCI Snoop log file here <--_
[06_07_2020__20_51_36-btsnoop_hci.log](https://github.com/oliexdev/openScale/files/4886775/06_07_2020__20_51_36-btsnoop_hci.log)

3. Bluetooth HCI Snoop log file
user settings in the vendors app:
```
same as first one
```
measured true values in the vendors app for the 3. HCI Snoop log file:
```
110.75kg, 20:52 06

110.75kg, 20:52 06/07/2020, 53.2%, 47.6% , 26.4%
```
_--> Attach the 3. HCI Snoop log file here <--_
[...]

PR #632, opened 2020-09-22
add Icomon.java

ChristianB86:
Deocde weight data for Insmart scale (and probably other icomon scales)


oliexdev:
Thanks I replied it in the issue #593

PR #641, opened 2020-10-30
bugfix #603: make app resizable on android desktops

### LINKED ISSUE #603: Make openScale resizable

I'm using [Samsung DeX](https://www.samsung.com/us/explore/dex/), which gives a desktop like experience and makes use of Android windowing support.
openScale works fine on that except that it isn't resizable.

Regarding [Samsung documentation this on how to enable window resizing for Android applications](https://developer.samsung.com/samsung-dex/modify-optimizing.html):

```xml
<application
    android:resizeableActivity="true">
</application>
```
This documentation also has more tweaks and information on how to improve the desktop experience in case you're interested.
They also have [an app testing guide](https://developer.samsung.com/samsung-dex/testing.html) (although some apps are fine with this guide but won't work on "real" DeX sometimes).

**Describe alternatives you've considered**

Not resizing openScale

**Additional context**

At the moment the resizing button is greyed out and the application is always nearly squarish:

Screenshot of openScale on Samsung DeX

**PS:** I know that it is possible to _force_ all apps to be resizable via `Samsung Labs` > `Force Resize` but it would obviously better to make the application natively resizable to create less friction.

alexanderadam:
Hi @oliexdev,

This PR fixes #603.

I admit that setting up the Android development environmet was pretty terrible.
Nothing worked out of the box, Android Studio got stuck twice on the openScale project and all that just for adding three lines and testing whether it worked.

Here's the bugfix for making openScale working properly on Android Desktops like Samsung DeX.
I was able to resize everything.

**PS:** If you're going to merge this: would you be so kind and add the [`hacktoberfest-accepted`](https://hacktoberfest.digitalocean.com/hacktoberfest-update) label to this PR? Thank you in advance!

### DIFF
--- a/android_app/app/src/main/AndroidManifest.xml
+++ b/android_app/app/src/main/AndroidManifest.xml
@@ -8,6 +8,8 @@
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED"/>
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
+    <uses-configuration  android:reqTouchScreen="finger" />
+    <uses-feature  android:name="android.hardware.touchscreen"  android:required="true" />
 
     <!-- Permission to allow read of data from the database through a ContentProvider.
          Marked "dangerous" so that explicit user approval is required to read this data, not
@@ -27,6 +29,7 @@
         android:roundIcon="${appIconRound}"
         android:label="@string/app_name"
         android:name=".core.Application"
+        android:resizeableActivity="true"
         android:theme="@style/AppTheme" >
         <activity
             android:name=".gui.MainActivity"


oliexdev:
Thanks for your PR. Unfortunately, I see a some layout error if I resize openScale on icons and graphs.

Since you can force any app to be resizeable it is not mandatory to enable it in the manifest. I will add this function if it works properly.

Why you enable to require a touch screen? On a non touchable monitor it will not work.

Thanks for your work.

alexanderadam:
> Since you can force any app to be resizeable it is not mandatory to enable it in the manifest. I will add this function if it works properly.

It doesn't work the same if you force it. Apps without the resizeable attribute can only be toggled between two states: a fixed small view and maximised. So no dragging in the corners, no window tiling and no free resizing.
Furthermore the app will be completely restarted and initialized when you toggle the size (a dialogue will be opened that informs you about this every time).

So the only way to have a regular resizing like people are used to, is by enabling this in the app itself.

> Why you enable to require a touch screen? On a non touchable monitor it will not work.

Only by accident. I can remove that.

PR #672, opened 2020-12-28
Reminder day preference fix

dankeast:
The reminder day preference summary was listing the days out of order, and also always using the english text of the values. I've also removed the square brackets around the list, I thought it made it look a bit like debug output.

Before:

Screenshot_1609194427

After:
Screenshot_1609194522
Screenshot_1609194539


oliexdev:
thx

PR #682, opened 2021-01-08
Replace %i format specifier with %d

PeterHuewe:
`%i` is not a valid format specifier in Java and if this code is executed creates an exception. Use `%d` instead.

Signed-off-by: Peter Huewe <peterhuewe@gmx.de>


oliexdev:
thx

PR #704, opened 2021-02-26
Add support for Hoffen BBS-8107 bathroom scale

ppkt:
In this PR I added a support for Hoffen smart smart scale available in Poland: [User Guide PL/EN](https://hoffen.com.pl/wp-content/uploads/2018/02/WAGA-%C5%81AZIENKOWA-9-W-1-Z-FUNKCJ%C4%84-ANALIZY_BBS-8545.pdf) (it seems currently in sale is newer model BBS-8545, but both looks the same). Fortunately, the protocol was quite simple because scale itself doesn't seem to store any data.


oliexdev:
awesome thanks for your PR :+1: 

Could you take a photo of your scale, like the same angle for our [gallery](https://github.com/oliexdev/openScale/wiki/Supported-scales-in-openScale)?
Unfortunately I can't use the official photos due copyright issues.

Issue #496, opened 2019-09-15
Support request for Chipsea based Body Fat analyzers


MetalGearSolidNerd:
Hi, It would be great if support for Chipsea based weight scales and body fat analyzers can be added to OpenScale. These chipsets are used in multiple bluetooth enabled weight scales and there is a common app to get the data from sensors - [OKOK International](https://play.google.com/store/apps/details?id=com.chipsea.btcontrol.en)

I was able to find a reference ble-lib implementation [here](https://github.com/mmengchen/chipsea-ble-lib)

I also converted the document (bluetooth protocol) on the link to english which I am attaching here
 for your reference

[Core Sea Bluetooth Scale Cloud Edition Communication Protocol v0.3.pdf](https://github.com/oliexdev/openScale/files/3614062/Core.Sea.Bluetooth.Scale.Cloud.Edition.Communication.Protocol.v0.3.pdf)

Let me know if the above is sufficient or should I also provide all the logs and data as per the reverse engineering instructions on the wiki.

Thank you.

oliexdev:
Yes please provide all logs and data according to the [openScale wiki](https://github.com/oliexdev/openScale/wiki/How-to-reverse-engineer-a-Bluetooth-4.x-scale).
The pdf is only a small part of the protocol. The calculation is done in the [.so library](https://github.com/mmengchen/chipsea-ble-lib/blob/master/blelib/libs/armeabi/libcsalgorithmex.so) which needs to be reverse engineered.

MetalGearSolidNerd:
I returned the body fat analyzer due to some issues and will not be able to provide requested data. Sorry about that, please close this issue if required.

VarshaVijayvargiya:
Please give me Chipsea based Body Fat analyzers in ios. If any sdk available for ios

VarshaVijayvargiya:
@MetalGearSolidNerd : Let me know what issue you found in android. I don't found any issue , let me know there is any problem with your reference sdk which you found: [here](https://github.com/mmengchen/chipsea-ble-lib)

Your reference help me alot to get weight data.

inzanity:
There's relevant log and data in #410, it seems to be a chipsea based scale using the "OKOK protocol". As suggested by the doc, the scale does not actually do any calculation, it only provides weight and 1 or 2 resistance values (depending on model).

From logcat with a "EB8217":
[...]

PR #706, opened 2021-03-03
Add support for chipsea / OKOK 2.0 scales

inzanity:
For now only handles weight, impedance is read, but not used for
anything, algorithms not known.

The scale does not indicate whether it is sending the weight in kg or in
lbs, always assuming kg.

Based on findings from #496

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothFactory.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothFactory.java
@@ -118,6 +118,9 @@ public class BluetoothFactory {
         }
         if (deviceName.equals("Hoffen BS-8107")) {
             return new BluetoothHoffenBBS8107(context);
+	}
+        if (deviceName.equals("ADV")) {
+            return new BluetoothOKOK(context);
         }
         return null;
     }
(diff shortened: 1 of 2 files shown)


oliexdev:
thanks for your work but could rewrite your code that you don't overwrite connect/disconnect and onDiscoveredPeripheral. You can put the logic in onNextStep and onBluetoothNotify

inzanity:
I didn't figure out a way to get the advertisement data any other way, am I missing something?

oliexdev:
You need to turn on the notification in onNextStep with:

```
        switch (stepNr) {
            case 0:
                //Tell device to send us weight measurements
                setNotificationOn(WEIGHT_MEASUREMENT_SERVICE, WEIGHT_MEASUREMENT_SERVICE);
                break;
         }
```

Then you will receive any weight advertisement in onBluetoothNotify.

The WEIGHT_MEASUREMENT_SERVICE and WEIGHT_MEASUREMENT_SERVICE could be different but the default values are:

```
    private final UUID WEIGHT_MEASUREMENT_SERVICE = BluetoothGattUuid.fromShortCode(0xfff0);
    private final UUID WEIGHT_MEASUREMENT_CHARACTERISTIC = BluetoothGattUuid.fromShortCode(0xfff1);
```

inzanity:
The device is a pure broadcaster and does not advertise any GATT services (nor characteristics). All the data is simply in a manufacturer specific data field in the advertisement.

The fields present in the advertisement are:
 * 02 01 04: flags = BLE_GAP_ADV_FLAG_BR_EDR_NOT_SUPPORTED
 * 04 09 04 09 41 44 56: name = ADV
 * 16 ff  ... = 23 bytes of manufacturer specific data

As far as I understand, this data would only be received in the `onDiscoveredPeripheral()` callback in `BluetoothCommunication.java` which throws the `ScanResult` object away and stops receiving updates after the first received package (which contains only the very first measurement, likely quite close to zero).

Still, this is the only way I've perceived this kind of broadcaster-listener can be implemented without altering `BluetoothCommunication`.

oliexdev:
that would be really strange, could you do step 2 from https://github.com/oliexdev/openScale/wiki/How-to-reverse-engineer-a-Bluetooth-4.x-scale and post the results here?

inzanity:
[openScale_2021-03-04_09-57.txt](https://github.com/oliexdev/openScale/files/6081791/openScale_2021-03-04_09-57.txt)

When I click on the scale, it pops up a never ending spinner.

oliexdev:
you are using an old version of openScale?

But if its not work please use the BLE scanner app.

inzanity:
blescanner

This is all that I got with BLE Scanner, which says the same, not connectable and 3 fields in the advertisement.

PR #718, opened 2021-03-29
SBF70: Make BeurerSanitas state machine more robust

### LINKED ISSUE #717: SBF70 connection issues
**Describe the bug**
I'm aware that very similar bugs have been submitted before and that the underlying problem is a timing problem in the scale itself. But I'm willing to dig into that and try to come up with a solution that makes things more stable, e.g. by retrying when timing failed. But I still like to have this bug open here as a place to discuss ideas and possible solutions.

Most important thing I noticed: After the first unhandled exception the app itself seems to end up in a undefined state where it will not be able to connect until it is hard-killed and restarted. So I guess that's the problem I need to solve first.

I will need to learn how to build an android app first, so it may take a couple of days before I have new ideas.

If anyone has good ideas about what to try first, please bring them up.

**To Reproduce**
1. Open app
2. Click Bluetooth button
3. Step on scale
4. Wait for scale to finish measuring
5. No new measurement visible in app
6. Repeat as often as you want
7. Kill app
8. Open app
9. Touch scale to activate, but do not step on it
10. Click Bluetooth button
11. Measurement is transmitted

Reproduced with 2.3.5-pro (54) on two devices.

**Expected behavior**
Measurement is transmitted.

**Debug log**
Here is an exemplary log file. Unfortunately the logging disabled itself between step 7 and 8, so I do not have a log of the successful connection. I'll add that later.
[openScale_2021-03-27_09-04.txt](https://github.com/oliexdev/openScale/files/6215548/openScale_2021-03-27_09-04.txt)

holgerpieta:
The SBF70 has three bad habits:
- It sends the scale status reply multiple times (three, as it seems).
- When the user is standing on the scale, measurement are sent out unrequested.
- In both situations the scale ignores other commands.

Previously, the BeurerSanitas state machine was not robust against those effects:
- When it received messages that didn't not match the current state, it sometimes still advanced to the next state. This caused missing information and undefined behavior in later steps.
- When it sent out a request while the scale was busy with something other, the request got lost, the state machine never received the required answer and it got stuck in a state.
- When a measurement message was received before the user list was compiled, a null dereference cause an unhandled exception and the state machine stopped working.

With the attached patches the state machine will:
- Ignore / discard measurement messages before the user list is compiled.
- Retries at least some states when unexpected messages are received.

I think there will still be a lot of corner cases and timing problems that may cause unexpected behavior, crashes or stuck state machines, but it should happen a lot less frequently now.

I do not think that it will break other BeurerSanitas scales, but I cannot test this. So if there are reports, notify me and I'll try to take care about it.

This should fix #717.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothBeurerSanitas.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothBeurerSanitas.java
@@ -369,9 +369,10 @@ public class BluetoothBeurerSanitas extends BluetoothCommunication {
 
             if (currentRemoteUser.remoteUserId != remoteUsers.get(remoteUsers.size() - 1).remoteUserId) {
                 // Only jump back to state 5 if we are in 5
-                jumpNextToStepNr( 5, 5 );
-                // Only resume state machine when we are in state 5
-                resumeMachineState( 5 );
+                if( jumpNextToStepNr( 5, 5 ) ) {
+                    // Now resume
+                    resumeMachineState();
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


oliexdev:
Thanks for your work :1st_place_medal:

PR #738, opened 2021-06-05
WIP: Add support for OKOK 1.1 scales (weight only)

inzanity:


ag88:
hi @inzanity, 
Thanks, nice update. But some thoughts:
there seem to be some initialization in com/health/openscale/core/bluetooth/BluetoothFactory.java
https://github.com/oliexdev/openScale/pull/706/commits/98aa80915b4a229c64f8d74ba56aa019b2e2296e#diff-276c3a918a43e28517b26673e1068d9f96f2ef9e6de848b60d9c28376f31c64a

the selection of the 'driver' apparently depends on device name.
as discussed in the issue log
https://github.com/oliexdev/openScale/issues/496#issuecomment-855212614
I'm not too sure if the device would be detected as 'Chipsea-BLE' instead for a 1.1 device

inzanity:
Yes, very true. Added the (only known) name in two places.

ag88:
my goof, when i presented my data, i missed one byte, this affected index of the checksum byte
https://github.com/inzanity/openScale/blob/okok11/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothOKOK.java#L35
i check that once again the checksum should be the 19th byte
checked that in the specs as well
https://github.com/oliexdev/openScale/files/5224454/OKOK.Protocol.pdf
hence it should be
```
 private static final int IDX_V11_CHECKSUM = 18;
```

inzanity:
Actually all the indices were wrong, that one just a bit less than the others. The data array doesn't include the first two bytes.

ag88:
i see, i tried to verify the logic by 'testing' it against my data posted prior ;)
maybe it might read as follows:
```
    private static final int IDX_V11_WEIGHT_MSB = 3;
    private static final int IDX_V11_WEIGHT_LSB = 4;           
    private static final int IDX_V11_BODY_PROPERTIES = 9;
    private static final int IDX_V11_CHECKSUM = 16;
```
it seemed in your code  ``private static final int IDX_V11_WEIGHT_LSB = 5;`` is offset one more byte away?
To get the index I subtract 3 from the byte count position.
weight MSB is 6th byte so IDX_V11_WEIGHT_MSB = 3;
weight LSB is 7th byte IDX_V11_WEIGHT_LSB = 4; 
'body properties' is 12th byte so IDX_V11_BODY_PROPERTIES = 9;
and checksum is 19th byte so IDX_V11_CHECKSUM = 16;

i've kind of verified that the checksum for that first 18 bytes using a straight xor as in the codes matches that in the 19th byte.
this is done 'manually'

inzanity:
Oops, the LSB was a typo.

Indeed you are correct, the 0xca is part of the checksummed data (unlike with v2.0).

The last 6 bytes of the data are (almost) the MAC of the device, don't know why, but every reported device sends it.

ag88:
hi inzanity, could you mark this pr as ready for review? this is so that it can be included say in a next (beta) release.
i'd download the released binaries to try that out so that we can iterate it further if any issues is found.
Thanks !

oliexdev:
As I don't own this scale. Does this PR works for your scale, so I can merge it into the master branch?

ag88:
hi oliexdev,
as we've kind of done a 'code review' checking against actual ble data frames, verifying them byte for byte. my guess is there is a good chance this would work. i'd install the next binary that's released and test it out. should there be any issues i'd feedback again about it.

PR #741, opened 2021-06-13
Fix bodyfat calculation for 1byone

lbdroid:
Adjust the value of bodyVar instead of replacing bodyVar with the adjustment factor.


oliexdev:
thanks I guess you tested the changes with the 1byone scale for various users, right?

Issue #159, opened 2018-01-22
Add support for 1byone scale


dinki:
I'm writing in hopes that you can add 1byone BT scale (https://www.amazon.com/1byone-Bluetooth-Wireless-Bathroom-Visceral/dp/B01FHELB56/)  to your list of supported devices.  I received this as a gift from my wife but it is tied to a proprietary app that leaves much to be desired.

While I am much more of a hacker than a programmer, I am pretty sure that this is possible as someone has written some software that pulls the information from the scale:

https://github.com/ec1oud/qt-ble-weightscale

Further, I was able to use BT Snoop to pull the log where the scale connects and passes the information on to the phone.  You can see the same type of information listed at the above repo here:

https://github.com/ec1oud/qt-ble-weightscale/blob/master/doc/notes.txt

Here's what it looks like:

```
Frame 13693: 32 bytes on wire (256 bits), 32 bytes captured (256 bits)
Bluetooth
Bluetooth HCI H4
Bluetooth HCI Event - LE Meta
    Event Code: LE Meta (0x3e)
    Parameter Total Length: 29
    Sub Event: LE Advertising Report (0x02)
    Num Reports: 1
    Event Type: Connectable Undirected Advertising (0x00)
    Peer Address Type: Public Device Address (0x00)
    BD_ADDR: TexasIns_2a:8d:a6 (f0:c7:7f:2a:8d:a6)
    Data Length: 17
    Advertising Data
        Flags
            Length: 2
            Type: Flags (0x01)
            000. .... = Reserved: 0x0
            ...0 .... = Simultaneous LE and BR/EDR to Same Device Capable (Host): false (0x0)
            .... 0... = Simultaneous LE and BR/EDR to Same Device Capable (Controller): false (0x0)
            .... .1.. = BR/EDR Not Supported: true (0x1)
            .... ..0. = LE General Discoverable Mode: false (0x0)
            .... ...1 = LE Limited Discoverable Mode: true (0x1)
        16-bit Service Class UUIDs (incomplete)
            Length: 3
            Type: 16-bit Service Class UUIDs (incomplete) (0x02)
            UUID 16: Unknown (0xfff0)
        Manufacturer Specific
            Length: 9
            Type: Manufacturer Specific (0xff)
            Company ID: Ericsson Technology Licensing (0x0000)
            Data: f0c77f2a8da6
    RSSI (dB): -60

Frame 13694: 33 bytes on wire (264 bits), 33 bytes captured (264 bits)
[...]

PR #747, opened 2021-06-16
Update calculations for OneByone scales

ASerbinski:
Previously, the value reported for muscle mass was lean body mass
less skeletal mass. This was not a correct report for muscle mass
since it included additional values, such as organs and skin mass.

In addition, the impedance value previously used was incorrect.
There were, in fact, two different calculations being performed
for impedance based on data fields that were not relevant to it,
therefore the value for impedance was entirely incorrect, jumpy,
and unpredictable. In fact, the correct bytes for impedance are
the second and third bytes in the data packet received from the
scale, which produces correct and reasonable impedance values.

This commit corrects the prior errors, provides a calculation for
lean body mass, and calculates muscle mass based on the Janssen
formula;

https://journals.physiology.org/doi/pdf/10.1152/jappl.2000.89.2.465

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothOneByone.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothOneByone.java
@@ -145,9 +145,8 @@ public class BluetoothOneByone extends BluetoothCommunication {
 
     private void parseBytes(byte[] weightBytes) {
         float weight = Converters.fromUnsignedInt16Le(weightBytes, 3) / 100.0f;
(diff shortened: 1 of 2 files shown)


oliexdev:
Thanks for your PR. did you compare the results with the vendors app for the muscle mass? Do they have the same results?

ASerbinski:
They are absolutely NOT the same results. The vendor software is suggesting muscle mass of around 75-80%, which is simply wrong.

oliexdev:
I don't own this scale but regarding to #159 they provide the muscle mass in kg, maybe the 75-80 are kg?

ASerbinski:
That would bring the muscle mass percentage to 100%.

PR #748, opened 2021-06-16
Bring muscle mass eval sheet to proper ranges

ASerbinski:
The original values were obtained from a commercial scale user
manual, which was not consistent with ranges found from reputable
academic or health sources. This updates the range limits using a
health source that is consistent with academic sources and MRI
based calculations.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/evaluation/EvaluationSheet.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/evaluation/EvaluationSheet.java
@@ -119,25 +119,13 @@ public class EvaluationSheet {
 
         waterEvaluateSheet_Woman.add(new sheetEntry(10, 1000, 45, 60));
 
-        muscleEvaluateSheet_Man.add(new sheetEntry(10, 14, 44, 57));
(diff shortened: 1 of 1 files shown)


oliexdev:
Thanks, do you have a reference of an academic paper for these muscle values?

ASerbinski:
> Thanks, do you have a reference of an academic paper for these muscle values?

Table 1 on the top of the 3rd page;
https://journals.physiology.org/doi/pdf/10.1152/jappl.2000.89.1.81#

oliexdev:
Thanks but your PR doesn't match with the values from the paper does it?

ASerbinski:
No matter how you look at it, these are going to be ballpark figures not exact. So yes it is somewhat simplified, but reasonably consistent with proper sources. I can make it dead on with that academic paper if you prefer.

https://www.fitnesss.net/everything-about-muscle-mass-percentage-with-charts/

oliexdev:
yes that would be nice, so we can comment in the source code from which paper we get these values.

ASerbinski:
OK, I'll update it this evening. I don't have a computer with me where I am right now.

PR #767, opened 2021-08-18
Initial driver for Salter MiBody.

chrismonks:
Hi,
Did you ever get this working with the Salter MiBody?


oliexdev:
I don't have this scale but if you own one, you can try out the @agatti implementation if it works.

PR #778, opened 2021-10-17
Save body water as percentage instead of kg on BF105

Gerii:
The BF105 reports the body water in kg while the app (and the scale) displays body water as percentage.
This pull request converts the received value so that it is displayed correctly.
As I was unsure if this issue also applies to the other Beurer scales that inherit from the BluetoothStandardWeightProfile I put the fix into the BF105 specific class.


oliexdev:
thanks I cannot confirm that because I don't own any Beurer scale but I will merge it. We will see if some other Beurer scales reports that problem.

PR #806, opened 2021-12-09
fixed new user registration for BF105 (and may be other scales) and…

jensMF:
…fixed crashes because of `Fragment BluetoothSettingsFragment{...} (...) not attached to a context.` error(s)

Might have side effects for other scales, so it would be good, if some people would test it.


oliexdev:
Thanks for your fix :+1:

PR #811, opened 2021-12-28
Added support for Sanitas SBF 72

Xalag:
The Sanitas SBF 72 is compliant to the BLE Weight Scale Profile and has an additonal custom service which is very similar to the one of Beurer 105.
It does support:
- 8 users
- activity level
- gender
- height
- date of birth
- displaying a users pin on the scale's display (for assigning an existing scale user to an app user)

It does not support:
- user initials
- target weight

Supported measurments:
- weight
- BMI
- fat percentage
- water percentage
- muscle percentage
- bone weight
- base energy turnover
- activity energy turnover


oliexdev:

PR #820, opened 2022-01-16
Support for 1byone scale (new version)

mirko-laruina:
Support for the new 1byone scale.

It includes multi-user handling, live and historic measurements and different units (kg, lb and st).  
Live measurements are assigned to the selected user, historic ones can use the intelligent user recognition if the option is active.

The scale is able to measure:
- Weight
- BMI
- Body Fat
- Water
- Muscle
- Lean Body Mass
- Bone Mass
- Visceral Fat

Note to the maintainers:
- I've modified some project dependecies to make it compile on my machine (see the modified gradle files).
- I've added a method to the OpenScale class to obtain the "assignable" user (didn't know how else to call it): it returns the closest user to the given weight if smart user assignment is active, otherwise it returns the selected user.


oliexdev:
Thanks for your great work and sending the PR :1st_place_medal: 

I needed to revert the gradle updates because Travis doesn't work with the new version.

PR #832, opened 2022-02-23
CubicSpline interpolation

BhasherBEL:
Use cubic spline interpolation instead of horizontal bezier interpolation for cleaner and smoother interpolation.

**Before:**

**After:**


oliexdev:
What the difference in your cubic spline calculation and the one if you can set with CUBIC_BEZIER?

I actually did set the CUBIC mode but as you can see values goes more outside of your real data points which impress a wrong measurement in my opinion.

PR #850, opened 2022-04-23
Fix Senssun excessive toasts

quthla:
This PR aims to fix the excessive toast messages when taking measurements with a Senssun scale.

With these changes, the app will show only one toast when the weight has stabilized instead of 20 or more toast messages like right now.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothSenssun.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothSenssun.java
@@ -115,7 +115,7 @@ public class BluetoothSenssun extends BluetoothCommunication {
         switch(data[5]) {
             case (byte)0xAA:
             case (byte)0xA0:
-                if (values > 1) {
+                if (weightStabilized) {
                     return;
                 }
                 if (!stepMessageDisplayed) {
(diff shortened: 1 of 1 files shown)


android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothSenssun.java
```diff
@@ -115,7 +115,7 @@ private void parseMeasurement(byte[] data) {
         switch(data[5]) {
             case (byte)0xAA:
             case (byte)0xA0:
-                if (values > 1) {
+                if (weightStabilized) {
```
oliexdev, line 118: You checked the weightStabilzed value before it will bet set?

quthla: Not sure what you mean? The value is false until set to true [here](https://github.com/oliexdev/openScale/blob/7e644087266c866ee08d13f60f7c8f860e4a79d1/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothSenssun.java#L126)

oliexdev: What happened if the first parse measurement the weight stabilized is true? You have to call again parse measurement...

quthla: It is set to true [here](https://github.com/oliexdev/openScale/blob/7e644087266c866ee08d13f60f7c8f860e4a79d1/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothSenssun.java#L126) so if that happens, it will just go past the if condition, save the weight and weightStabilized and go on with the next step.

oliexdev: Ok thanks I can't try this fix out but I guess you tested it.

PR #861, opened 2022-05-24
Support Sinocare CW286 bluetooth bathroom scales

MoralCode:
This PR adds support for the Sinocare [CW266 Digital bluetooth bathroom scale](https://en.sinocare.com/products/sinocare-digital-bathroom-scales-weighing-scale-high-precision-sensors-bluetooth-health-analyzer-with-smart-app-body-weight-scale).

This scale appears to be similar in operation to the scale added in #706 with regard to the contents of [this](https://github.com/oliexdev/openScale/pull/706#issuecomment-790393142) comment which basically says the scale only uses the manufacturer data field to broadcast data using bluetooth's Advertising mechanism and does not offer a proper service for accessing these values. The same is true for this scale and I suspect the reason for this is that this sinocare scale has a rather short timeout of about 10 seconds before it turns off, making it hard for a user to connect and get the data before it turns itself off.

Also from that thread it appears that this is the second scale to operate this way
> ok that would be the first scale without an Bluetooth service but if its works I will merge it.

To my knowledge and the best of my wireshark/deduction skills based on the wiki (which to me seems like it may be a little out of date regarding the process of getting the bluetooth packet capture) this scale only supports advertising weight values. Because i couldnt find any solid evidence for the data format, this (and a checksum) is all i have implemented, but my limited testing seems to suggest that it works well enough.

Because of a lack of reliable indication that the scale is "done" measuring (i.e. users weight has stabilized) this implementation checks for this based on receiving a sequence of many identical weight values in a row. I admit this is not ideal and should be updated once better information about the operation of this scale is known.

Hopefully this explains the bulk of how this scale operates, let me know if you have questions or need me to make any changes!

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothFactory.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothFactory.java
@@ -130,6 +130,9 @@ public class BluetoothFactory {
         if (deviceName.equals("SBF72")) {
             return new BluetoothSanitasSBF72(context);
         }
+        if (deviceName.equals("Weight Scale")){
+            return new BluetoothSinocare(context);
+        }
         return null;
     }
 }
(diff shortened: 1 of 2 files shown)


oliexdev:
Thanks for your awesome PR :+1:

PR #873, opened 2022-07-31
Init commit of AH100 / CH100.

Iliaroz:
It could connect to device, write user info and current time/date. Could read measurement (online, standing on the scale). In theory it should read offline measurements too, but not working on my scale.

In scale datasheet is mentioned that offline measurement will not work if two device are bound to same scale. Maybe it is the reason, why I cannot get offline measurements (scale do not recognize which user measurement apply to)


oliexdev:
Thanks for your awesome initial support of the Huawei AH100 / CH100 scales. :rocket:

Iliaroz:
Thank you.

In this scale I have to send user weight to scale. What is the correct way to get current user weight?

PR #896, opened 2022-11-06
MiScale2: muscle conversion: use a local variable

manio:


oliexdev:
thx maybe that was the bug with the strange fat values.

PR #911, opened 2023-01-15
Support for Beurer BF915 (based on Sanitas SBF72).

Giesd:
The Beurer BF915 seems to work just like the Sanitas SBF72. I just had to add a constructor parameter for the device name.

I have successfully tested the following functions:

- Connection and initial setup
- Using users who have already been created on the scale (with pin confirmation)
- Data transmission (including history data)
- User auto detection


oliexdev:
thanks for the PR :+1:

PR #916, opened 2023-01-29
statistics: add total weight lost

### LINKED ISSUE #663: Total weight lost
Often happens that I want to check how many kg I've lost from the beginning of my dieting journey, but there is not this type of indication within openScale

I would add in the "Statistics" page a new voice called "Total weight loss" that shows the total amount of kg lost

BoboTiG:
It would fixes #663.

I adapted the CI workflow to ease working for forks, I guess it will work for the main repository too, but let me know if you prefer to revert those changes.

One thing is missing: a nice icon. But I didn't understand how icons are stored/created. If you already have a nice one to propose, and guide me a little bit, I would be more than happy to update the PR.

And one question: is it more logical to use a positive number rather than a negative one? I mean "-0.2 kg" means the user lost 0.2 kg. So maybe "0.2 kg" would be more suitable :thinking: 

Result:
img2


oliexdev:
Thanks for your PR. I need to test it if the CI workflow is still working.

I found my icons on https://www.flaticon.com/search?word=weight%20loss They are mostly free to use. If you found a nice one you can simple right click in Android Studio under `res->drawable` then click in the context menu `New->Vector Assest` to add a new icon.

I think it's better to use negative values for weight loss, some users want to gain weight.

BoboTiG:
For the icon:
1. https://www.onlinewebfonts.com/icon/562680 (it's a reward, because it's good to celeabrate achievements)
2. https://www.onlinewebfonts.com/icon/55183 (same kind, but with a certificate)
3. https://www.onlinewebfonts.com/icon/453823 (handshake, like a contract with yourself is going well)

WDYT?

oliexdev:
I would suggest https://www.flaticon.com/free-icon/weight-loss_4813685?term=weight+loss&page=1&position=1&origin=search&related_id=4813685

I am not sure if everybody understand the reward and the contract idea.

BoboTiG:
Done, I updated screenshots in the issue description.

PR #917, opened 2023-02-01
fix: total weight lost computation

BoboTiG:
:sweat:


BoboTiG:
Ah sorry for the noise, I should get some sleep, and stop thinking.

Issue #816, opened 2022-01-09
Auto backup works only to the Download or Documents folders


marvinwankersteen:
**Describe the bug**
Hello,
for me, the autobackup function has not worked properly since mid 2021. I think since I switched to Android 11.
I have specified `Transfer/Backup/openScale` as the backup folder. That had also worked with it. I was able to look into it now and see where the problem is and it seems that openScale is only allowed to write to the `Download` and `Documents` folder.

**To Reproduce**
1. Go to 'Settings' --> 'Backup'
2. Click on 'Export folder' and set `Transfer/Backup/openScale` (`Transfer` will be synced with syncthing)
3. See error in logfile

Reproduced with: v2.4.5 (60) from F-Droid

**Expected behavior**
Database should be saved.

**Additional context**
Is this the normal behavior, that only the backup may be written to `Download` or `Documents`?
Because if I manually export the backup to `Transfer/Backup/openScale` it works. Only the automatic backup does not work.
If I specify as backup folder for example `Download`, the backup is written successfully:
```
2022-01-09 12:50:10.075 Debug [2] AlarmBackupHandler: openScale Auto Backup to /storage/emulated/0/Download/2022-01-09_openScale.db
```

Logcat:
```
01-09 12:42:07.188  6494 24285 W MediaProvider: Forgot to handle a top level directory in getContentUriForFile?
01-09 12:42:07.191  6494 24285 E MediaProvider: insertFileIfNecessary failed
01-09 12:42:07.191  6494 24285 E MediaProvider: java.lang.IllegalArgumentException: Primary directory Transfer not allowed for content://media/external_primary/file; allowed directories are [Download, Documents]
01-09 12:42:07.191  6494 24285 E MediaProvider: 	at com.android.providers.media.MediaProvider.ensureFileColumns(MediaProvider.java:2707)
01-09 12:42:07.191  6494 24285 E MediaProvider: 	at com.android.providers.media.MediaProvider.ensureUniqueFileColumns(MediaProvider.java:2372)
01-09 12:42:07.191  6494 24285 E MediaProvider: 	at com.android.providers.media.MediaProvider.insertFile(MediaProvider.java:2981)
01-09 12:42:07.191  6494 24285 E MediaProvider: 	at com.android.providers.media.MediaProvider.insertInternal(MediaProvider.java:3533)
01-09 12:42:07.191  6494 24285 E MediaProvider: 	at com.android.providers.media.MediaProvider.insert(MediaProvider.java:3249)
[...]

Issue #867, opened 2022-06-23
Auto Export does not work, "Permission not granted"  


Tzatzikiiii:
**Description**

Tapping the checkbox "Auto Backup"  in Settings/Backup only shows the toast "Permission not granted" despite giving openScale the permission to save files (both via the settings and the dialogue that pops up when tapping the check box). It also won't allow changing the export directory since the option is greyed out. Exporting manually works fine however. (Samsung A22 5G, Android 11, One UI Core 3.1)

1. Tap Settings/Backup/Auto Backup
2. Grant the permission via the dialogue 
3. Observe toast "Permission not granted", check boxes greyed out
4. Further taps on the checkbox result in duplicate toasts unless you manually take its permission away in the android settings (then it displays the initial dialogue again)

Reproduced with [latest dev version](https://github.com/oliexdev/openScale/releases/tag/dev-build): yes, happens on latest devbuild 

**Expected behavior**
The permission should be sufficient to save the backup. The checkbox should also stay activated and not immediately deactivate again. The greyed out options should be usable.

**Debug log**

Removed my data, otherwise as is (I doubt it's useful though)

```
2022-06-24 01:00:44.105 Debug [2] AboutPreferences: Debug log enabled, openScale v2.4.6 (61), SDK 30, samsung SM-A226B
2022-06-24 01:00:44.108 Debug [2] AboutPreferences: Selected user id(3) name() birthday() age() body height() scale unit(kg) gender() initial weight() goal weight() goal date() measure unt(cm) activity level(2) assisted weighing(false)
2022-06-24 01:00:56.564 Debug [2] AboutPreferences: Debug log disabled
```

Tzatzikiiii:
appears to be related to [#816](https://github.com/oliexdev/openScale/issues/816)

Electrenator:
Can confirm this issue since I am also unable to export via the auto backup on my Motorola also running Android 11. The manual export seems to have no problem though.
<details><summary>Made a little GIF of the "Permission not granted" action on app version 2.4.6 (61):</summary> ![output](https://user-images.githubusercontent.com/18311389/176144642-ba1bf892-3cff-4b99-80bf-29e24c247356.gif)</details>

oliexdev:
thanks for the report, please try out the latest dev-build.

Electrenator:
Can confirm that the checkbox can now indeed be activated

pho3nixx:
[...]

Issue #859, opened 2022-05-15
Export dir path prefix


FifiTheReal:
**Describe the bug**

Since last release i can't modify export path, because export path has path prefix:
"/storage/emulated/0/Android/data/com.health.openscale.pro/files/"

I would like use "/storage/emulated/0/Download/openscale-Backup" as before but i can't do it.

**Debug log**
[openScale_2022-05-15_16-40.txt](https://github.com/oliexdev/openScale/files/8695281/openScale_2022-05-15_16-40.txt)

![Screenshot_20220515_163808_com health openscale pro](https://user-images.githubusercontent.com/34412887/168480523-adadad10-d2b0-4c9a-9354-4dd41f0ae0d0.jpg)

oliexdev:
That's not possible anymore because in >= Android 11 they change the storage permissions, see https://developer.android.com/about/versions/11/privacy/storage
You cannot create in the background files that are outside your app scope.

Issue #684, opened 2021-01-11
backup: Add test backup button


obbardc:
I've setup the daily backup function but I can't wait until tomorrow to test that everything is working.

It would be nice to have a "Test backup" or "Backup now" button to test the backup works correctly.

oliexdev:
just click export backup...

PR #924, opened 2023-02-11
Fixed auto backup;

FFDA:
Made auto backup compatible with SAF.

- Dev-build had some fixes for Andriod10+ devices that allowed to check "Auto backup" checkbox in settings. However, it still used [getExternalFilesDir()](https://developer.android.com/reference/android/content/Context#getExternalFilesDir(java.lang.String)) that returns directory that is hard to reach and is deleted when app is removed from the device. Moreover (while I didn't check it on dev-build), there is [a comment](https://github.com/oliexdev/openScale/issues/816#issuecomment-1380063080) that this issues still persisted even with the fix.
- This fix asks user to select a directory where to store the backup database(s) and lets it change later;
- Added "Test auto backup" button.

While all the issues are closes this pull request fixes: https://github.com/oliexdev/openScale/issues/816 https://github.com/oliexdev/openScale/issues/867 https://github.com/oliexdev/openScale/issues/859 https://github.com/oliexdev/openScale/issues/684

Tested on Android12 physical device and Android6, Android9 and Android12 virtual devices.

Screenshot of how backup preferences looks now:
https://user-images.githubusercontent.com/8406967/218267366-39b11c82-7eca-490e-a35e-9e30b767beff.png


oliexdev:
Thanks @FFDA  for your PR. Could you remove the Test auto backup button or move it to the settings->About->Development because I don't want to have any test buttons in the main menus.

FFDA:
> Thanks @FFDA  for your PR. Could you remove the Test auto backup button or move it to the settings->About->Development because I don't want to have any test buttons in the main menus. 

Which option do you prefer?

oliexdev:
then I would prefer to remove it.

FFDA:
I removed the backup test button from the settings.

oliexdev:
Thanks :+1:

FFDA:
Yeah, my bad. I actually tried about it and added this code:
```
DocumentFile exportDir = DocumentFile.fromTreeUri(context, Uri.parse(exportDirString));
// Check if it is possible to read and write to auto export dir
// If it is not possible auto backup function will be disabled
if (!exportDir.canRead() || !exportDir.canWrite()) {
    SharedPreferences.Editor editor = prefs.edit();
    editor.putBoolean("autoBackup", false);
    editor.apply();
    return;
}
```
It seems it didn't work :/ And I only tested on new installs.

PR #927, opened 2023-02-19
Add support for Yoda1

hyx0329:
Yoda1 is a simple scale which broadcasts its measurements through advertising packets, with the name "Yoda1". It's LE only and not connectable. Some resellers say that the device can measure body impedance, but from my experience it only has a fake, fixed value equals to 0x1388.

I copied some code from `BluetoothOKOK`.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothFactory.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothFactory.java
@@ -136,6 +136,9 @@ public class BluetoothFactory {
         if (deviceName.equals("CH100")){
             return new BluetoothHuaweiAH100(context);
         }
+        if (deviceName.equals("Yoda1")){
+            return new BluetoothYoda1Scale(context);
+        }
         return null;
     }
 }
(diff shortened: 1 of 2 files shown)


oliexdev:
Thanks for your PR :+1: Do you have a website of the scale and the vendor app which is being used?

hyx0329:
My scale is bought on Taobao, and similar products can be found by searching `huawei hilink body scale` on google(price from 19CNY to 60CNY). [This link](https://detail.1688.com/offer/681659238699.html?spm=a26352.13672862.offerlist.4.76f5c87940SaoH) points to a same product. BTW, there are products with almost identical appearance on walmart, like [this one](https://www.walmart.com/ip/Body-Fat-Scale-Smart-BMI-Digital-Bathroom-Wireless-Weight-Scale-Composition-Analyzer-Smartphone-App-sync-Bluetooth-396-lbs-Rose/210073844).

Most of the resellers claim that the scale is compatible with _Huawei HiLink_(a smart home platform by Huawei). I have never actually used their app([direct link for android](https://appgallery.cloud.huawei.com/appdl/C10414141)), but I found project [yoda1-scale](https://github.com/denysvitali/yoda1-scale) which inspired me.

PR #931, opened 2023-02-21
Support for additional OKOK/Chipsea-BLE based scale

rrooggiieerr:
Support for additional OKOK/Chipsea-BLE based scale


oliexdev:

PR #953, opened 2023-04-16
Mi Scale: Set scale time when wrong

ondrej-zary:
Set time when hour/minute is differrent, not only year/month/day. Useful for DST changes or clock drift.

I'm not able to build APKs so it's not tested.


oliexdev:
Would be better if somebody could test that because I don't know if the mi scale returns the hour in 12-hour clock or 24 hour clock. This depends what you have to compare against Calendar.HOUR or Calendar.HOUR_OF_DAY. Might also  depending on the some kind of scale settings?

oliexdev:
As nobody had the change to test it, I will close it but thanks for your PR.

Issue #940, opened 2023-03-12
Add SBF73 its same SBF72


davvarrr:
Hello, I just bought a scale SBF73 it and not support, but seen that the SBF72 is well recognized I tested by changing just the SBF72 in SBF73 and it works very well, the user already created on the scale is well recognized and all the information on the body are also.

if you need pictures of the app said the month, for the moment I will keep the one I created to support the SBF73

I don't know how to extract the log ble and sorry for my english I translate them by deepl

2spirit:
Hello, where do you change the name SBF72 -> SBF73 ?

2spirit:
OK, I found it , @oliexdev , in ./android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothFactory.java
line 130 , you can add "deviceName.equals("SBF73")"

Thanks !

PR #959, opened 2023-06-02
Add Sanitas SBF73 ( same as SBF72)

2spirit:
Hello, could you add please the Sanitas SBF73 , it is the same as SBF72 #940


2spirit:
#940

oliexdev:
sure thx for PR.

Issue #716, opened 2021-03-22
Make Week & Month (and year) average come back (Please)


SuzukaDev:
Older versions of OpenScale had the option to **see the average for each week (and month (and year))**. It was removed at some point the last year, IIRC, and now is **missing**. I waited for a few updates, and I thought it was "deprecated" by error and it would be added again, since is such a trivial and useful feature. But it didn't happen.

Another user opened another similiar issue:
https://github.com/oliexdev/openScale/issues/670

It was told to "just change the view in the graph to week". But, no. Changing the graph view to week (or month), only changes the labels on the Y-axis to week (or months).
But **the graph is exactly the same, there is no week/month/year average**.

Could it be added again, please?
I'm not asking for a new feauture, just to incorporate again that missing functionality that was removed.
Personally, it was my favourite feature, and the thing that helped me the most to see and track my progress. Daily measures are not as precise as week averages to track the progress, and since it was removed, I'm making the average via pen&paper... wich is weird, because thats why I use(d) the app, to track those things for me.

Thank you so much!

john-bill:
don't know when it as been removed, but its really missing.
But I'm also missing the (very-)old Statistics-View, when we had statistics instead of useless (and from day one wrong!) diagrams...

oliexdev:
Thanks @undefiened we have a new option which use a simple moving average on the chart.

PR #968, opened 2023-07-21
Simple moving average on the chart

undefiened:
This code adds a simple moving average trend line with an option to select the time period over which the line should be averaged.

It is different from the currently implemented exponential smoothing because the simple moving average does not account for the old measurements: even though the old measurements are supposed to contribute little, they still contribute, and therefore the current state of things isn't represented properly (at least for me).

There are no future predictions in this implementation, only the moving average.

Also, my simple moving average should work properly with irregular measurements, while the current implementation of the exponential smoothing works incorrectly when there are long periods without data since it works measurement-wise, not time-wise. I.e., if I have the following list of measurements (an arbitrary example with random numbers)

January 1st 2022 - 100 kg
December 1st 2023 - 50 kg
December 2nd 2023 - 50 kg
...
December 29th 2023 - 50 kg

then the whole thing would be skewed by the January 1st 2022 measurement, even though technically it was more than a year ago compared to the December 1st 2023 measurement.

I tried locally implementing interpolation to fix exponential smoothing when there are gaps in data by filling all missing days with simple linear interpolation, it works better that way, but still looks different from the simple moving average. (It works slowly, so I decided to not put my attempt at the interpolation fix into this pull request)

Also, this code should somewhat address https://github.com/oliexdev/openScale/issues/716 although not directly.

I tried this code on my data and it looks fine to me.

Please let me know if something doesn't look good or needs to be done, I am not familiar with the rules on how one should contribute to this repository.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/gui/measurement/ChartMeasurementView.java
+++ b/android_app/app/src/main/java/com/health/openscale/gui/measurement/ChartMeasurementView.java
@@ -698,7 +698,7 @@ public class ChartMeasurementView extends LineChart {
 
 
     private void addMeasurementLineSimpleMovingAverage(List<ILineDataSet> lineDataSets, List<Entry> lineEntries, FloatMeasurementView measurementView) {
-        LineDataSet measurementLine = new LineDataSet(lineEntries, measurementView.getName().toString() + "-" + getContext().getString(R.string.label_simple_moving_average));
+        LineDataSet measurementLine = new LineDataSet(lineEntries, "");
         measurementLine.setLineWidth(1.5f);
         measurementLine.setValueTextSize(10.0f);
         measurementLine.setColor(measurementView.getColor());
@@ -713,6 +713,7 @@ public class ChartMeasurementView extends LineChart {
         measurementLine.setHighLightColor(Color.RED);
         measurementLine.setDrawCircles(false);//prefs.getBoolean("pointsEnable", true));
         measurementLine.setDrawValues(prefs.getBoolean("labelsEnable", false));
+        measurementLine.setForm(Legend.LegendForm.NONE);
 
         if (measurementView.isVisible()) {
             if (isInGraphKey) {
--- a/android_app/app/src/main/java/com/health/openscale/gui/measurement/ChartMeasurementView.java
+++ b/android_app/app/src/main/java/com/health/openscale/gui/measurement/ChartMeasurementView.java
@@ -398,6 +398,10 @@ public class ChartMeasurementView extends LineChart {
             addTrendLine(lineDataSets);
         }
 
+        if (prefs.getBoolean("simpleMovingAverage", false)) {
+            addSimpleMovingAverage(lineDataSets);
+        }
+
         if (!lineDataSets.isEmpty()) {
             LineData data = new LineData(lineDataSets);
             setData(data);
@@ -429,8 +433,8 @@ public class ChartMeasurementView extends LineChart {
         measurementLine.setDrawCircles(prefs.getBoolean("pointsEnable", true));
         measurementLine.setDrawValues(prefs.getBoolean("labelsEnable", false));
         measurementLine.setMode(LineDataSet.Mode.HORIZONTAL_BEZIER);
-        if (prefs.getBoolean("trendLine", false)) {
-            // show only data point if trend line is enabled
+        if (prefs.getBoolean("trendLine", false) || prefs.getBoolean("simpleMovingAverage", false)) {
+            // show only data points if trend line or simple moving average is enabled
             measurementLine.enableDashedLine(0, 1, 0);
         }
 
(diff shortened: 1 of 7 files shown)


oliexdev:
Takes a little bit more to review it but somehow we have now double lines of each measurements!? Was that intended? (One normal and one average line but I see no difference)

undefiened:
That is weird, I am not entirely sure what you mean. Probably I messed up something.

This is how it looks to me.

Without moving average:



With moving average:

oliexdev:
please go the chart fragment and enable the legend.

undefiened:
> please go the chart fragment and enable the legend.

Got it, you mean that it is shown two times in the legend. This is the same behavior as for the trendline, so not sure how I should deal with it.

undefiened:
I removed the second entry for the moving average from the legend. Would that be a good solution?

oliexdev:
My fault no please add the legend again. What's only now confusing is the naming of "Trendline" and "Simple moving average" in the options / code now. :/
Nobody knows what method is behind "Trendline"  (problem not new) and "Simple Moving Average" could be also said as a "Trendline".
Maybe its better to add in the graph settings an option to enable/disable the trendline in general and add a second option to select the method of calculation!? (none, moving  average, exponentially smoothed moving average and future ones). Would also solve the confusion with many parallel visible trendlines. But would be more complex and a lot of renaming have to be done.
What do you think? Do you have a better idea?

undefiened:
You are right. I did the changes in the settings, does it look better now?

android_app/app/src/main/java/com/health/openscale/gui/measurement/ChartMeasurementView.java
```diff
@@ -395,7 +399,18 @@ public void refreshMeasurementList() {
         }
 
         if (prefs.getBoolean("trendLine", false)) {
-            addTrendLine(lineDataSets);
+            switch (prefs.getString("trendlineComputationMethod", "Exponentially Smoothed Moving Average")) {
```
oliexdev, line 402: would be better to work with ENUM variable.

android_app/app/src/main/res/values/arrays.xml
```diff
@@ -118,4 +118,14 @@
...
         <item>@string/label_time_period_set_custom_range</item>
     </string-array>
+
+    <string-array name="trendline_computation_methods_entries">
+        <item>@string/simple_moving_average</item>
+        <item>@string/exponentially_smoothed_moving_average</item>
+    </string-array>
+
+    <string-array name="trendline_computation_methods_values">
+        <item>Simple Moving Average</item>
```
oliexdev, line 128: Ok I see what you did but maybe remove the spaces so its looks more like a variable

android_app/app/src/main/res/xml/graph_preferences.xml
```diff
@@ -24,10 +25,41 @@
...
+        <SeekBarPreference
+            android:id="@+id/simpleMovingAverageNumDaysSeekbar"
+            android:defaultValue="7"
+            android:key="simpleMovingAverageNumDays"
+            android:dependency="trendLine"
+            android:max="30"
+            app:min="2"
+            app:showSeekBarValue="true"
+            android:title="@string/label_simple_moving_average_num_days" />
+        <CheckBoxPreference
```
oliexdev, line 57: maybe its better to move that under the trendline category?

android_app/app/src/main/java/com/health/openscale/gui/measurement/ChartMeasurementView.java
```diff
@@ -76,6 +76,10 @@ public enum ViewMode {
     private boolean isInGraphKey;
     private ProgressBar progressBar;
 
+    private interface TrendlineComputationInterface {
+        public List<ScaleMeasurement> operation(List<ScaleMeasurement> measurementList);
```
oliexdev, line 80: what does operation do in Java?

oliexdev (commented):
Wow you are really fast, looks really good.

android_app/app/src/main/java/com/health/openscale/gui/measurement/ChartMeasurementView.java
undefiened: I am not sure how to connect enums and Android preferences, can you give any advice or example for that? Currently, I added a separate class with two static Strings which contain the values I specified in the `arrays.xml` file.

android_app/app/src/main/res/xml/graph_preferences.xml
undefiened: I moved it, but there is now a small issue (at least on my emulator) that when the user selects Simple Moving Average, the seeker bar is below the screen, so the user doesn't know it appears.

android_app/app/src/main/java/com/health/openscale/gui/measurement/ChartMeasurementView.java
undefiened: An interface with a single method can be used to pass lambda into a method. https://www.w3schools.com/java/java_lambda.asp
It doesn't matter which name to use, so I renamed it to reduce confusion.

undefiened:
@oliexdev thank you again for the quick reviews!

android_app/app/src/main/java/com/health/openscale/gui/measurement/ChartMeasurementView.java
oliexdev: I am not sure either. Anyway I just delete the separate class and put the two static string to ChartMeasurementView.

android_app/app/src/main/res/xml/graph_preferences.xml
oliexdev: thanks fixed that.

android_app/app/src/main/java/com/health/openscale/gui/measurement/ChartMeasurementView.java
oliexdev: Yeah indeed sorry I was a little bit confused I thought this were a new lambda function of Java 11 or something. :laughing:

oliexdev:
Really big thanks for your awesome implementation @undefiened :fireworks:

undefiened:
Thank you very much for all the help and patience and especially for this app in the first place :)

PR #969, opened 2023-07-22
Increase the offset so that Y-axis labels with decimals places fit

undefiened:
I don't really know how to properly fix this, but experimentally I have found the value of the offset multiplier which fixes the issue of labels being cut out. It eats a little bit more space when no decimal places are present, but at least it works fine otherwise. I understand that this is a dirty hack, so no hard feelings from my side if it doesn't make it upstream.

Screenshot before:

Screenshots after:


undefiened:
Actually, after creating this PR, I think I've found a less dirty hack by just adding one symbol worth of space similar to how the charting library does it.

oliexdev:
thanks for your PR :+1:

PR #970, opened 2023-07-22
Settings for overview Y-axes visibility

undefiened:
Add a new setting that allows enabling or disabling Y-axes on the overview screen. It allows selecting left, right, both, or no axes on the overview screen. When any of the two axes are enabled, it also automatically enables grid lines (I think it makes sense to do so).


undefiened:
Sorry, I've accidentally messed up branches, I will reopen this PR.

PR #971, opened 2023-07-22
Settings for overview Y-axes visibility

undefiened:
Add a new setting that allows enabling or disabling Y-axes on the overview screen. It allows selecting left, right, both, or no axes on the overview screen. When any of the two axes are enabled, it also automatically enables grid lines (I think it makes sense to do so).

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/gui/measurement/ChartMeasurementView.java
+++ b/android_app/app/src/main/java/com/health/openscale/gui/measurement/ChartMeasurementView.java
@@ -629,4 +629,5 @@ public class ChartMeasurementView extends LineChart {
             }
         }
     }
+    
 }
--- a/android_app/app/src/main/java/com/health/openscale/gui/measurement/ChartMeasurementView.java
+++ b/android_app/app/src/main/java/com/health/openscale/gui/measurement/ChartMeasurementView.java
@@ -630,8 +630,13 @@ public class ChartMeasurementView extends LineChart {
         }
     }
 
-    public long getNumberOfMeasurementViewsInOverview() {
-        return measurementViews.stream().filter(view -> view instanceof FloatMeasurementView && view.isVisible() && view.getSettings().isInOverviewGraph()).count();
+    public long getNumberOfMeasurementViewsInOverview(YAxis.AxisDependency axisDependency) {
+        boolean isOnRight = axisDependency == YAxis.AxisDependency.RIGHT;
+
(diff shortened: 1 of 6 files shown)


oliexdev:
Thanks for your PR. I think the y-axis option in the settings are not the right place because this option is specific for the overview graph and not for all graphs. I would prefer to have an option in the already existing popup menu (bottom right corner of the graph) of the overview graph. It could be simplify to only show/hide the y-axis. I don't think to have two more options with only left and only right is necessary. What do you think?

undefiened:
@oliexdev Yeah, I think it makes sense. The reason why I wanted to have control of the left and right axes separately is that the right one is often not needed (I only really care about the weight, so the right axis just eats up space). But this probably can be solved separately. I can imagine two possible solutions:
1. add a single selector like the one I had before which would work for both overview and full graph views
2. automatically disable right y-axis if there is only one measurement

Would any of them make sense to you?

Otherwise, I think I changed things according to your comments. I am not sure how to call the button though. Is "Y-axis" fine?

oliexdev:
thanks @undefiened for the quick redesign. The second option would be better because there is no user interaction necessary. Yeah the label is fine. I renamed just the internal label name to a more general name (removed the word "overview")

undefiened:
Thanks for the quick reviews!

I added autohiding of the right y-axis in the overview. I decided to not make `measurementViews` in `ChartMeasurementView` public, so I added a method that returns the number of measurement views activated in overview. Is that fine? Or should I just make it public?

Should I also add autohiding in the full graph view?

undefiened:
Sorry, I just discovered that there is also some logic regarding that in `isOnRightAxis` method in `MeasurementViewSettings`. How should I deal with it? Should I hide left axis instead when the only measurement view is one which should be on the right axis?

oliexdev:
True, well I put the weight on the right axis and the rest measurements on the left axis by default. The best would be to count how many are on the left side and how many are on the right side and depending of the counting enable/disable the left/right axis but I don't know if that's easy to implement. If not than remove the "auto hide one axis" feature. ;)

undefiened:
I implemented autohiding of the appropriate axis, it is not difficult, but I am not sure if it looks better. I feel like having y axis at least on the left of the chart is more natural. I would rather remove the "auto hide one axis" feature then unless you like the current behavior.

oliexdev:
I agree with you. Thanks for your effort :1st_place_medal:

PR #990, opened 2023-09-02
Goal line length and width modified

### LINKED ISSUE #659: The goal color should be more distinguished

Also using the Overview graph as an example:

The `Goal line` color is too similar to other data.

It's should have a unique color/style to better differentiate the goal from other information.

Arshjeet2003:
Fixes #659
WhatsApp Image 2023-09-02 at 8 20 16 PM (1)


oliexdev:
thanks for your PR but I prefer to keep them smaller.

PR #1041, opened 2024-04-20
Initial support for icomon BleBroadcastScales

Duncaen:
This adds support for scales as described in https://github.com/oliexdev/openScale/issues/593.

The model I have is a "Foryond FG263LB", with the device name "AAA002", for which I added the support, other models described in the linked issue probably work too if their name is added.

I'm not very familiar with android or java so this is pretty basic and hacked together. I had to somehow get the `BluetoothConnection` class to not be connection orientated, so I added an empty constuctor and overwrote the
connect and disconnect methods to instead just scan for bluetooth le advertisement messages from the know mac address.

The other ignored bluetooth packet contains the impedance measurement, but without the full algorithms its useless

`BroadcastScale` is named after the internal name in the original app, couldn't come up with anything better, since there are a lot of different brandings for this OEM.

### DIFF
--- a/android_app/app/build.gradle
+++ b/android_app/app/build.gradle
@@ -2,7 +2,7 @@ apply plugin: 'com.android.application'
 apply plugin: "androidx.navigation.safeargs"
 
 android {
-    compileSdkVersion 34
+    compileSdk 34
     defaultConfig {
         applicationId "com.health.openscale"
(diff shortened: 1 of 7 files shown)


nanomad:
Confirmed working with device name AAA007, nice job!

oliexdev:
Thanks I merged it into the repository.

PR #1046, opened 2024-05-06
Fix #1038 dark mode dialog text color issue

### LINKED ISSUE #1038: Minor dark mode issue
**Describe the bug**

Buttons "Ok" and "Cancel" are barely visible in openScale Pro's dark mode.

**To Reproduce**
Steps to reproduce the behaviour:

1. When creating a new user the datepicker for the birthday has black buttons in dark mode

Screenshot_20240420_102835_openScale

**Expected behaviour**

The buttons should be properly visible instead

crvdgc:
Fixed two places where the button colors are hard-coded. Also changed `AlertDialog`'s style so that the positive and negative buttons have proper text colors, instead of black.

Closes #1038.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/gui/MainActivity.java
+++ b/android_app/app/src/main/java/com/health/openscale/gui/MainActivity.java
@@ -273,7 +273,7 @@ public class MainActivity extends AppCompatActivity
 
                 // ask the user once for feedback on the 15th app launch
                 if(launchCount == 15){
-                    AlertDialog.Builder builder = new AlertDialog.Builder(this);
+                    AlertDialog.Builder builder = new AlertDialog.Builder(this, R.style.AppTheme_Dialog);
 
                     builder.setMessage(R.string.label_feedback_message_enjoying)
                             .setPositiveButton(R.string.label_feedback_message_yes, new DialogInterface.OnClickListener() {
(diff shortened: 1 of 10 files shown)


oliexdev:
First of all thx for the PR. On which Android API did you tested it? I tried it with API level 33 and 23 and unfortunately it  the text on the birhtday date picker on #1038 is still the same.

crvdgc:
> On which Android API did you tested it?

I'm sorry, due to a change of computer, I lost the previous configuration for this project. I think it's automatically picked up by Android Studio, so it's very likely to be the latest one. (Sorry, I'm not very familiar with Android development as well.)

One thing I used to test the PR is to add a manual measurement without creating a user, which triggers [this `showNoSelectedUserDialog` error](https://github.com/oliexdev/openScale/pull/1046/files#diff-34be5c9890dda3a5f060ca49f8868ca9403af2a52c91d65b4a6af6dfcb58de77R378). Before this PR, the button in dark mode is black. This PR should at least fix buttons in that dialog, can you test if that is working? If so, the birthday picker error may be a result of a missing change.

oliexdev:
Thanks, you PR works only for the dialog color. I have finally find a solution to fix every dialog color button problem by only changing the theme xml file, see commit https://github.com/oliexdev/openScale/commit/8b4326687c91d33f496a0f37dc5fc7e14388aca9
But thanks for your effort, your PR gave me the right way.

PR #1057, opened 2024-08-14
Remove location permision not needed

### LINKED ISSUE #937: Bluetooth requires location to be turned on (Mi Smart Scale 2 / OnePlus NordCE 5G)
Mi Smart Scale 2 requires location to be enabled on OnePlus NordCE 5G. This was not the case a few weeks ago.

If I press the bluetooth icon (or have it enabled to auto turn on) it shows this screen, opens the 'enable location' screen and doesn't turn on bluetooth unless location is already on.

openscale

This should only be required for *finding* a bluetooth scale for the first time, no? If I click on my already-paired scale in the settings, it opens the same window, with the small difference that it has a 'no' button too.

So seems like a bug?

FAQ info which I believe was mentioned in https://github.com/oliexdev/openScale/issues/887 :

> **Why do openScale needs permission to the coarse location?**
> The answer is simple because the Android API >= 6.0 needs access to the coarse location to search for Bluetooth devices. The App openScale doesn't read your location at any time. Read the official [Android Bluetooth Developer Guide](https://developer.android.com/guide/topics/connectivity/bluetooth-le.html#permissions) or read the [Android 6.0 Changelog](https://developer.android.com/about/versions/marshmallow/android-6.0-changes.html#behavior-hardware-id) for a more technical description.

Doesn't really make sense to me. Especially because in android app settings, location is not allowed for openscale. And if I enable location permission, the exact same thing happens - unless it's actually turned on.

xshyne:
The location permission is not needed when SDK >= 31.

Should fix #937

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/gui/MainActivity.java
+++ b/android_app/app/src/main/java/com/health/openscale/gui/MainActivity.java
@@ -600,31 +600,6 @@ public class MainActivity extends AppCompatActivity
             return;
         }
 
-        // Check if GPS or Network location service is enabled
-        LocationManager locationManager = (LocationManager) getSystemService(LOCATION_SERVICE);
-        if (!(locationManager.isProviderEnabled(LocationManager.GPS_PROVIDER) || locationManager.isProviderEnabled(LocationManager.NETWORK_PROVIDER))) {
-            Timber.d("No GPS or Network location service is enabled, ask user for permission");
-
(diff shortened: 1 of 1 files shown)


oliexdev:
Thanks for your PR.  If I remember right, some strange scales needs the Bluetooth search before they actually send the actual measurements but I have to look more into it.

oliexdev:
should be solved in openScale 3.0 beta version.

PR #1081, opened 2024-10-18
Support for nameless OKOK scales (Myria MY4836)

### LINKED ISSUE #950: Another OKOK scale
**Scale name**
Nameless chinese stuff - okok international
https://play.google.com/store/apps/details?id=com.chipsea.btcontrol.en&hl=en&gl=US

I tried the latest version (v2.5.1 (63)) and clicked the "detect blueooth" button before and after stepping on the scale, but it never seems to recognize it. 

**Step 2: Acquiring some Bluetooth traffic**
_Attach 3 log files with the corresponding true values, read [here](https://github.com/oliexdev/openScale/wiki/How-to-reverse-engineer-a-Bluetooth-4.x-scale#1-acquiring-some-bluetooth-traffic) for further information._

reports weight only (i think)
Weight values (oldest to newest):
81.50 kg ( 1FD6 )
50.75 kg ( 13D3 )
84.15 kg ( 20DF )

1. Bluetooth HCI Snoop log file
user settings in the vendors app:
```
sex (male/female), body height, age, activity level
male, 180cm, 37, active (but I don't think any of this is relevant with this device)

_--> Attach the 1. HCI Snoop log file here <--_
[hci_snoop20230329140200.cfa.zip](https://github.com/oliexdev/openScale/files/11097992/hci_snoop20230329140200.cfa.zip)



_--> Attach the openScale debug log file here <--_
openscale did not see the scale at all, I used the BLE Scannner, it showed 10FFC05B000017700A01245D0249A63DE8 as "raw data" for the MAC address of the scale

Mac address to search for in Bluetooth snoop: 5d:02:49:a6:3d:e8

I don't think any data is sent from the phone to the scale, the scale just advertises the weight while it's working.

### PR DESCRIPTION
**Pull request by Florin9doi**
Closes #950

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/gui/preferences/BluetoothPreferences.java
+++ b/android_app/app/src/main/java/com/health/openscale/gui/preferences/BluetoothPreferences.java
@@ -17,6 +17,7 @@ package com.health.openscale.gui.preferences;
 
 import android.content.SharedPreferences;
 import android.os.Bundle;
+import android.text.TextUtils;
 import android.view.LayoutInflater;
 import android.view.Menu;
 import android.view.MenuInflater;
@@ -41,6 +42,9 @@ public class BluetoothPreferences extends PreferenceFragmentCompat {
     private Preference btScanner;
 
     private static final String formatDeviceName(String name, String address) {
+        if (TextUtils.isEmpty(name) && !address.isEmpty()) {
+            return String.format("[%s]", address);
+        }
         if (name.isEmpty() || address.isEmpty()) {
             return "-";
         }
--- a/android_app/app/src/main/java/com/health/openscale/gui/preferences/BluetoothSettingsFragment.java
+++ b/android_app/app/src/main/java/com/health/openscale/gui/preferences/BluetoothSettingsFragment.java
@@ -42,6 +42,7 @@ import android.preference.PreferenceManager;
 import android.provider.Settings;
 import android.text.SpannableStringBuilder;
 import android.text.Spanned;
+import android.text.TextUtils;
 import android.text.style.ForegroundColorSpan;
 import android.text.style.RelativeSizeSpan;
 import android.view.Gravity;
@@ -228,6 +229,9 @@ public class BluetoothSettingsFragment extends Fragment {
     }
 
     private static final String formatDeviceName(String name, String address) {
+        if (TextUtils.isEmpty(name) && !address.isEmpty()) {
+            return String.format("[%s]", address);
+        }
         if (name.isEmpty() || address.isEmpty()) {
             return "-";
         }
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by TRex22** [comment]
@Florin9doi Thank you for this 🙏🏻 

It helped me use my scale. 

I did notice that the impedance readings were not coming through. I took a look at `core/bluetooth/BluetoothOKOK.java` and see that the code appears incomplete. My scale uses the fallback conditional branch which does not extract impedance.

Ive started building out a processing module for OKOK compatible scales but Im not sure 100% which calculations to be using. Kind of guessing, trying to reverse engineer the app, and also looking at how the other scales calculate their data. My work is forked off your PR since my scale is one of the ones with no name.

I don't have examples of the other type of OKOK scales. Ill try my best to update the code in a way that helps everyone who uses these scales. 

If anyone has example payloads I could use them to test my work

File: android_app/app/src/main/java/com/health/openscale/gui/preferences/BluetoothPreferences.java [hunk 1]
```diff
@@ -41,6 +42,9 @@ public class BluetoothPreferences extends PreferenceFragmentCompat {
     private Preference btScanner;
 
     private static final String formatDeviceName(String name, String address) {
+        if (TextUtils.isEmpty(name) && !address.isEmpty()) {
```
oliexdev, line 45: Why you using TextUtils.isEmpty(name) instead of name.is.Empty()?

android_app/app/src/main/java/com/health/openscale/gui/preferences/BluetoothSettingsFragment.java
```diff
@@ -228,6 +229,9 @@ public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
     }
 
     private static final String formatDeviceName(String name, String address) {
+        if (TextUtils.isEmpty(name) && !address.isEmpty()) {
```
oliexdev, line 232: Why you using TextUtils.isEmpty(name) instead of name.is.Empty()?

oliexdev:
Thanks for your PR, yeah some scales has no Bluetooth names. Might be problem if we have different vendors using scales with no Bluetooth names.

Could you explain why you using TextUtils.isEmpty instead of name.isEmpty()? If there is no specific reason I would change it to name.isEmpty() to make the code more consistent.

Florin9doi:
`TextUtils.isEmpty` is used to check also for null in a single operation, instead of `name==null || name.isEmpty`

oliexdev:
Thanks for your quick reply and your explanation, I have just merged your PR into the master branch. Thanks again! :1st_place_medal:

oliexdev:
I am sorry I had to revert it as it appears that of course every Bluetooth address appears now as the nameless scale. We need to figure out how to filter these out.
Unfortunately I cannot reopen this PR again :( My fault sorry.

oliexdev:
I created a new PR https://github.com/oliexdev/openScale/pull/1088 for reopen this PR

PR #1083, opened 2024-11-06
Adds compatibility with Renpho (QN Scale) scales that use the 2nd set of characteristics

EDharmasiri:
- I'm not able to test how this affects anyone with the older/currently working scales

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothQNScale.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothQNScale.java
@@ -53,6 +53,14 @@ public class BluetoothQNScale extends BluetoothCommunication {
     private final UUID CUSTOM5_MEASUREMENT_CHARACTERISTIC = UUID.fromString("0000ffe5-0000-1000-8000-00805f9b34fb"); // write-only
     /////////////
 
+    // 2nd Type Service and Characteristics (2nd Type doesn't need to indicate, and 4th characteristic is shared with 3rd.)
+    private final UUID WEIGHT_MEASUREMENT_SERVICE_ALTERNATIVE = UUID.fromString("0000fff0-0000-1000-8000-00805f9b34fb");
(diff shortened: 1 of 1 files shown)


ndurchx:
Regarding Renpho ES-CS20M: The app doesn´t crash anymore, but the measurements seem to randomly use the different types. Sometimes the real values are used, sometimes wrong values and the weight/10

oliexdev:
Thanks for your PR @EDharmasiri

@ndurchx do you mean by wrong values that the weight is sometimes divided by 10? Could you explain it more in detail when what happen?
Would be good if you solve this problem.

android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothQNScale.java
```diff
@@ -87,12 +95,29 @@ public String driverName() {
...
+                    timestamp -= SCALE_UNIX_TIMESTAMP_OFFSET;
+                    byte[] date = new byte[4];
+                    Converters.toInt32Le(date, 0, timestamp);
+                    writeBytes(WEIGHT_MEASUREMENT_SERVICE, CUSTOM4_MEASUREMENT_CHARACTERISTIC, new byte[]{(byte) 0x02, date[0], date[1], date[2], date[3]});
+                } catch (NullPointerException e) {
+                    useFirstType = false;
+                }
+
                 // set notification on for custom characteristic 1 (weight, time, and others)
-                setNotificationOn(WEIGHT_MEASUREMENT_SERVICE, CUSTOM1_MEASUREMENT_CHARACTERISTIC);
```
oliexdev, line 91: maybe we have here a timing problem as the above test with the catch ... statement will take some time!? Maybe somebody could try to add an extra step to setup the notifcations on?

fbievan:
Seems to be working for me, the body fat % does seem to be reporting differently (although I'm 99% sure that rephno app is underreporting it massively)

I'll have to test more later, but seems to work fine for me.

oliexdev:
thanks for testing, I will merge this PR, it seems at least that the app doesn't crashed anymore. Any updates are welcome.

g4rb4g3:
I see the exact same /10 issue here. I dropped 82 kg in a single day 😆. Anything I can help out to get it fixed?
I used the dev buld from 15th of March.

PR #1091, opened 2024-11-23
Support for Active Era BS-06 scales

NecroKote:
PR contains rudimentary support for Active Era BS-06 scales.
Scales report weight, feet balance (weight distribution), raw impedance readout from the ADC, and heart rate.
All of those are logged, but only weight is reported in the app. 

Historical measurements are also parsed, but not stored.

The BIA is done inside a native library, which I haven't looked at yet.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothActiveEraBF06.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothActiveEraBF06.java
@@ -1,0 +1,326 @@
+/* Copyright (C) 2024  olie.xdev <olie.xdev@googlemail.com>
+*                2024  Duncan Overbruck <mail@duncano.de>
+*
+*    This program is free software: you can redistribute it and/or modify
+*    it under the terms of the GNU General Public License as published by
+*    the Free Software Foundation, either version 3 of the License, or
+*    (at your option) any later version.
+*
+*    This program is distributed in the hope that it will be useful,
+*    but WITHOUT ANY WARRANTY; without even the implied warranty of
+*    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+*    GNU General Public License for more details.
+*
+*    You should have received a copy of the GNU General Public License
+*    along with this program.  If not, see <http://www.gnu.org/licenses/>
+*/
+
+package com.health.openscale.core.bluetooth;
+
+import android.content.Context;
+
+import com.health.openscale.R;
+import com.health.openscale.core.OpenScale;
+import com.health.openscale.core.datatypes.ScaleMeasurement;
+import com.health.openscale.core.datatypes.ScaleUser;
+import com.health.openscale.core.utils.Converters;
+
+import java.nio.ByteBuffer;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Calendar;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+import timber.log.Timber;
+
+/**
+ * Support for Active Era BS-06 scales
+ *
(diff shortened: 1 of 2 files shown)


oliexdev:

android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothActiveEraBF06.java
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    private void parseHistoricalPacket(byte[] pkt) {
+        Instant time = Instant.ofEpochSecond(Converters.fromUnsignedInt24Be(pkt, 3));
+        float weight = (Converters.fromUnsignedInt24Be(pkt, 0x08) & 0x03FFFF) / 1000.0f;
+        float weightLeft = Converters.fromUnsignedInt16Be(pkt, 0x0b) / 100.0f;
+        int hr = pkt[0x0d] & 0xff;
+        int adc = Converters.fromUnsignedInt16Be(pkt, 0x0f);
+        Timber.i("Historical measurement: %.3f kg, Weight Left: %.2f kg, HR: %d, ADC: %d", weight, weightLeft, hr, adc);
+        // TODO: store historical results
```
oliexdev, line 324: maybe just a simple

```
ScaleMeasurement historyMeasurement = new ScaleMeasurement();
historyMeasurement.setWeight(weight);
addScaleMeasurement(historyMeasurement);
```

would do it?

android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothActiveEraBF06.java
```diff
@@ -0,0 +1,326 @@
...
+
+                    // calculate BIA using measure weight and impedance
+                    if (impedance > 0.0) {
+                        final ScaleUser selectedUser = OpenScale.getInstance().getSelectedScaleUser();
+                        int height = (int) Math.ceil(selectedUser.getBodyHeight());
+                        int age = selectedUser.getAge();
+                        int gender = selectedUser.getGender() == Converters.Gender.FEMALE ? 0 : 1;
+
+                        calculateBIA(height, impedance, stableWeightKg, age, gender);
+                        // TODO: report results
```
oliexdev, line 258: you could set scaleData with the results?

android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothActiveEraBF06.java
NecroKote: The problem is - I'm not the sole user of this scales ) So unless we have a proper "select and claim" mechanic, I would rather not claim everything automatically

android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothActiveEraBF06.java
NecroKote: Here I've used example BIA calculation, just to see how far off the result whould be from what the scales are showing. The results are way off, so this is why I'm not storing the results

android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothActiveEraBF06.java
oliexdev: This should be solved with the settings->Bluetooth->Intelligence user assignment which automatically assign the user which have the nearest weight. You could also enable then ignore out of range measurements to ignore weights that out of the user weights.

android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothActiveEraBF06.java
oliexdev: ok yeah the vendors calculations are sometimes mysterious. I also implemented some estimates calculations, see https://github.com/oliexdev/openScale/wiki/Body-metric-estimations
You can enable them under settings->measurements->Body fat / Body water / Lean body mass

PR #1092, opened 2024-11-24
Add in a workflow for building locally

TRex22:
# What does this PR do?
I found it helpful to compile APKs locally but I did not have the correct Java and Android SDK setup so I wanted to use the GitHub workflow and `docker`. However the current workflow did not work with `act` and `docker`. 

Ive created a new workflow designed to work with `act` and `docker` so that you can generate apks locally. Also added a quick start to the readme to execute `act`. 

I felt this would be helpful to add to the main repo for anyone else wanting to do something similar and not having the correct Android SDK setup.


oliexdev:
Sorry, I think to setup the right envorinment using for e.g. Android Studio isn't so complicated.

PR #1096, opened 2024-12-20
add app ID suffix for the debug build

Strubbl:
With this change it is possible to have the debug version of the app installed in parallel to the release version from an app store.

I hope this does not interfere with any CI like travis.

### DIFF
--- a/android_app/app/build.gradle
+++ b/android_app/app/build.gradle
@@ -97,6 +97,7 @@ android {
                     }
                 }
             }
+            applicationIdSuffix '.debug'
         }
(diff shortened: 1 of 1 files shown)


oliexdev:
Thanks for your PR but before I can merge that I need some adjustments for openScale sync

Strubbl:
Yes, no problem. I used this to install & test the debug version of the app in parallel to the production version of the app.

oliexdev:
I added it in the openScale 3.0 beta build.

PR #1098, opened 2025-01-01
Added preference to vibrate when receiving data from scale

### LINKED ISSUE #1097: Add feature to vibrate or make a sound when logged data
No indication the data is logged when the user is on the scale without looking at the app 

A notification or haptic feedback to let the user know data is logged

jontyms:
Defaults to true
User configurable in general preferences
Added 2 strings info_bluetooth_retrieve_data_successful and label_vibrate_on_measurement
Added permission android.permission.VIBRATE
Fixes #1097

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/gui/MainActivity.java
+++ b/android_app/app/src/main/java/com/health/openscale/gui/MainActivity.java
@@ -35,6 +35,8 @@ import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.Message;
+import android.os.VibrationEffect;
+import android.os.Vibrator;
 import android.preference.PreferenceManager;
 import android.provider.Settings;
 import android.text.Editable;
@@ -115,6 +117,7 @@ public class MainActivity extends AppCompatActivity
     private NavController navController;
     private NavigationView navigationView;
     private BottomNavigationView navigationBottomView;
+
 
     private boolean settingsActivityRunning = false;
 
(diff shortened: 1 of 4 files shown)


android_app/app/src/main/java/com/health/openscale/gui/MainActivity.java
```diff
@@ -715,7 +718,7 @@ public void handleMessage(Message msg) {
                 case RETRIEVE_SCALE_DATA:
                     setBluetoothStatusIcon(R.drawable.ic_bluetooth_connection_success);
                     ScaleMeasurement scaleBtData = (ScaleMeasurement) msg.obj;
-
+                    Toast.makeText(getApplicationContext(), getResources().getString(R.string.info_bluetooth_retrieve_data_successful), Toast.LENGTH_SHORT).show();
```
oliexdev, line 721: On some scales you will retrieve 100 data point from the scales. Therefore its not good to show everytime a toast message. If a toast message is shown it is handled in `addScaleMeasurement` in `OpenScale.java`. If the parameter silent is true then now toast message is shown. An idea is to implement the vibrate feature to add it after ` runUiToastMsg(infoText);` in `addScaleMeasurement` but beware that it must be run in a ui thread as the  `addScaleMeasurement` will also be triggered during synchronization without a ui.

android_app/app/src/main/res/xml/general_preferences.xml
```diff
@@ -26,4 +26,10 @@
         android:summaryOff="@string/info_is_not_enable"
         android:summaryOn="@string/info_is_enable"
         android:title="@string/label_delete_confirmation" />
+    <CheckBoxPreference
+        android:defaultValue="true"
```
oliexdev, line 30: default should be false

oliexdev:
Thanks for your PR but currently I cannot merge this PR see the comments above.

oliexdev:
I added it in openScale 3.0 beta with commit ac8e27fbfb914a82f3f3ee501d76a1b773750ae9

PR #1108, opened 2025-01-26
Add RenphoScale/ES-WBE28

eronconi:
Add support for Renpho ES-WBE28.

Not sure how to better match the scale, since its reported device name is quite generic ("RenphoScale").

Only weight measurement works, I was not able to fully decode the body composition data (although some fields such as "Water [%]", "Subcutaneous Fat [kg]" and others have been decoded (see line 221)).


oliexdev:
Thanks for your PR :rocket:  We could use some body composition data if you are able to decode them.

PR #1110, opened 2025-02-01
Initial support for Beurer BF500

ghtyrant:
Beurer's BF500 seems to be a mix of BF105 and BF600. Characteristics determined by using BLE Scanner and trial & error. I did not verify that `CHARACTERISTIC_BEURER_BF500_SCALE_SETTING` and `CHARACTERISTIC_BEURER_BF500_REFER_WEIGHT_BF` are correct or do what they are supposed to do as they were not implemented by other scales.

What works:
- Transferring a user from the app to the scale (e.g. deleting all users on the scale before connecting with the app)
- Taking measurements (Weight, BMI, Body water, Body fat)
  - Body Water required the same calculations as with BF105 before working correctly, the rest was correct out-of-the-box

What doesn't work:
- I've set up a user with my correct birth year (1988), but on the scale the user was created with birth year 1995 - day and month were correct though. I don't know if this is a bug in the app or something went wrong when sending the data via Bluetooth.
When reading the user list from the scale, the birth year is transferred and parsed correctly, so updating the year on the scale worked.
- Automatic user detection doesn't seem to work. Someone else got on the scale and the scale still assigned the measurement to my user (even with vastly different weights). I don't know if this is an issue with the app (Smart User assignment is active) or the scale.

Let me know if you need any other data/characteristics from the scale to improve this first draft.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothBeurerBF500.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothBeurerBF500.java
@@ -1,0 +1,121 @@
+/* Copyright (C) 2019  olie.xdev <olie.xdev@googlemail.com>
+ *
+ *    This program is free software: you can redistribute it and/or modify
+ *    it under the terms of the GNU General Public License as published by
+ *    the Free Software Foundation, either version 3 of the License, or
+ *    (at your option) any later version.
+ *
+ *    This program is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


oliexdev:
Thanks for your PR :+1:  Did you add two different users in openScale with the initial weight that is near the "real" values before you use the "smart" user assignment?

ghtyrant:
No, I did not. Is this a necessary step for "smart" user assignment to work? For now, I only have one user (me) and when somebody else steps on the scale the measurement is assigned to me, independent of weight difference etc.

Thanks for merging!

oliexdev:
Yes, this is necessary because otherwise openScale has no other user to whom the measurement can be assigned.

SKrywinski:
When will the BF 500 be available in the apk?
I tried the latest development-version and still can't find it. Can you give me any hint?

PR #1115, opened 2025-03-11
Add support for Renpho ES-CS20M

DeerMaximum:
This pr adds support for the Renpho ES-CS20M scale with the firmware version v1.

The scale only sends the impedance, but when I use the 'YunmaiLib' to calculate the body composition data, I get similar results to the original app.

The scale does support historical data, but it is sent via the same characteristic as the normal measurements. This, and the fact that I couldn't find a date in the data, meant that I couldn't add support for historical data. To avoid messing up the measurements, the historical data is removed from the scale at the beginning (hint for the wiki?).


oliexdev:
Thanks for your awesome PR :+1: , I added it to the wiki with " o : supported by the scale but still needs to be reverse engineered" at the history data column

Issue #1126, opened 2025-05-21
Crashes when trying to connect to Renpho scale


rdnielsen:
**Describe the bug**
An error message is displayed with diagnostics when stepping on the scale.

**To Reproduce**
1. Open app.
2. Enable Bluetooth
3. Step on scale

Reproduced with [latest dev version](https://github.com/oliexdev/openScale/releases/tag/dev-build): 2.5.3 (65)

**Expected behavior**
App acquires data from scalee

**Additional context**
Diagnostics:
Build version: 2.5.3 
Build date: 1981-01-01 01:01:02 
Current date: 2025-05-21 07:11:39 
Device: Samsung SM-X210 
OS version: Android 14 (SDK 34) 
 
Stack trace:  
java.lang.NullPointerException: no valid characteristic provided
	at j$.util.Objects.requireNonNull(Unknown Source:5)
	at com.welie.blessed.BluetoothPeripheral.writeCharacteristic(BluetoothPeripheral.java:1224)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.writeBytes(BluetoothCommunication.java:303)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.writeBytes(BluetoothCommunication.java:291)
	at com.health.openscale.core.bluetooth.BluetoothQNScale.onNextStep(BluetoothQNScale.java:111)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.nextMachineStep(BluetoothCommunication.java:610)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.nextMachineStep(BluetoothCommunication.java:612)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.nextMachineStep(BluetoothCommunication.java:612)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.resumeMachineState(BluetoothCommunication.java:210)
	at com.health.openscale.core.bluetooth.BluetoothCommunication$1.onServicesDiscovered(BluetoothCommunication.java:464)
	at com.welie.blessed.BluetoothPeripheral$1$1.run(BluetoothPeripheral.java:213)
	at android.os.Handler.handleCallback(Handler.java:958)
	at android.os.Handler.dispatchMessage(Handler.java:99)
	at android.os.Looper.loopOnce(Looper.java:230)
	at android.os.Looper.loop(Looper.java:319)
	at android.app.ActivityThread.main(ActivityThread.java:8934)
[...]

Issue #1151, opened 2025-07-27
crashing on Renpho es-cs20m scale


TonyApuzzo:
**Describe the bug**
Click the Bluetooth button then step on my scale and the app crashes with a stack Trace.

**To Reproduce**
1. Add the scale as a device in open scale beta
2. Click on the Bluetooth icon
3. Step on the scale

Reproduced with [latest dev version](https://github.com/oliexdev/openScale/releases/tag/dev-build): ???

**Expected behavior**
Weight should be loaded from the scale 

**Additional context**

**Debug log**

Build version: 2.5.4-oss 
Build date: 1981-01-01 01:01:02 
Current date: 2025-07-27 08:36:47 
Device: Google Pixel 9 Pro 
OS version: Android 16 (SDK 36) 
 
Stack trace:  
java.lang.NullPointerException: no valid characteristic provided
	at j$.util.Objects.requireNonNull(SourceFile:244)
	at com.welie.blessed.BluetoothPeripheral.writeCharacteristic(BluetoothPeripheral.java:1224)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.writeBytes(BluetoothCommunication.java:303)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.writeBytes(BluetoothCommunication.java:291)
	at com.health.openscale.core.bluetooth.BluetoothRenphoScale.onNextStep(BluetoothRenphoScale.java:111)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.nextMachineStep(BluetoothCommunication.java:610)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.nextMachineStep(BluetoothCommunication.java:612)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.nextMachineStep(BluetoothCommunication.java:612)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.nextMachineStep(BluetoothCommunication.java:612)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.nextMachineStep(BluetoothCommunication.java:612)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.resumeMachineState(BluetoothCommunication.java:210)
	at com.health.openscale.core.bluetooth.BluetoothCommunication$1.onServicesDiscovered(BluetoothCommunication.java:464)
	at com.welie.blessed.BluetoothPeripheral$1$1.run(BluetoothPeripheral.java:213)
[...]

Issue #1104, opened 2025-01-11
App crashing when connecting to QN-Scale (Slimpal)


rosscullen:
**Describe the bug**
When I try to connect in OpenScale App to the "QN-Scale" Bluetooth broadcast, the app shows the ladybird crash icon and a copy of log is below.

**To Reproduce**
Reproduced with 2.5.3 (65)

**Debug log**
Build version: 2.5.3 
Build date: 1981-01-01 01:01:02 
Current date: 2025-01-11 22:55:44 
Device: Samsung SM-G973F 
OS version: Android 12 (SDK 31) 
 
Stack trace:  
java.lang.NullPointerException: no valid characteristic provided
	at j$.util.Objects.requireNonNull(Unknown Source:5)
	at com.welie.blessed.BluetoothPeripheral.writeCharacteristic(BluetoothPeripheral.java:1224)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.writeBytes(BluetoothCommunication.java:303)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.writeBytes(BluetoothCommunication.java:291)
	at com.health.openscale.core.bluetooth.BluetoothQNScale.onNextStep(BluetoothQNScale.java:111)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.nextMachineStep(BluetoothCommunication.java:610)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.nextMachineStep(BluetoothCommunication.java:612)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.nextMachineStep(BluetoothCommunication.java:612)
	at com.health.openscale.core.bluetooth.BluetoothCommunication.resumeMachineState(BluetoothCommunication.java:210)
	at com.health.openscale.core.bluetooth.BluetoothCommunication$1.onServicesDiscovered(BluetoothCommunication.java:464)
	at com.welie.blessed.BluetoothPeripheral$1$1.run(BluetoothPeripheral.java:213)
	at android.os.Handler.handleCallback(Handler.java:938)
	at android.os.Handler.dispatchMessage(Handler.java:99)
	at android.os.Looper.loopOnce(Looper.java:226)
	at android.os.Looper.loop(Looper.java:313)
	at android.app.ActivityThread.main(ActivityThread.java:8751)
	at java.lang.reflect.Method.invoke(Native Method)
[...]

Issue #1013, opened 2023-12-05
Renpho  ES - CS20M not showing up


FlameSilver:
I removed it from the renpho app as advised in the FAQ but even after that it just doesn't show up in the list at all. Tried the 2 most recent versions from fdroid as well which didn't make a difference

jsspen:
Were you ever able to solve this? I'm having the same issue with the ES-CS20M-W. It doesn't come up in the scan at all.

oliexdev:
Since there has been no progress on this issue for a long time, I am closing it for now.  

Please note that implementation of new scale support requires access to the **actual scale**.

If someone wants to continue working on this, feel free to **reopen this issue** or submit a working implementation via a **pull request**.

For guidance on how to add a new scale handler into openScale, see the updated wiki page:  **[How to support a new scale](https://github.com/oliexdev/openScale/wiki/How-to-support-a-new-scale)**

PR #1152, opened 2025-07-29
Add support for new protocol on Renpho ES-CS20M

keeperofdakeys:
Add support for new protocol on Renpho ES-CS20M by modifying BluetoothQNScale. Marked as a draft since I made a lot of changes that could break QN Scales, the code needs some clean up, and I still need to work out the best way to add scale selection.

My Renpho ES-CS20M identifies itself as "Renpho-Scale", but uses a modified QN Scale protocol. I've successfully used my changes to get weight and resistance values from the scale. However I've had to temporarily map "Renpho-Scale" to BluetoothQNScale, which breaks existing Renpho-Scale units.

I can see 4 ways to fix this:

1. Provide bluetooth device UUIDs to BluetoothFactory, so they can be considered when selecting a device driver.
Unfortunately this looks a little difficult, as the device driver appears to be selected before connecting to the bluetooth device in most code paths. This would need to be refactored so we can connect to the bluetooth device before calling BluetoothFactory, to get access to UUIDs. (Maybe MAC-Address? Need more data from users for this).

2. Merge BluetoothQNScale and BluetoothRenphoScale, or modify BluetoothRenphoScale to side-load BluetoothQNScale if required. This may be the easiest option.

3. Confirm the new BluetoothQNScale works on existing devices advertising "Renpho-Scale". (Just ES-WBE28?) Then BluetoothRenphoScale can be removed.

4. Force-override bluetooth driver in settings? Second easiest option, but bad UX.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothFactory.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothFactory.java
@@ -107,9 +107,6 @@ public class BluetoothFactory {
                 || deviceName.equals("042FatScale01")) {
             return new BluetoothInlife(context);
         }
-        if (deviceName.startsWith("QN-Scale")) {
-            return new BluetoothQNScale(context);
-        }
         if (deviceName.startsWith("Shape200") || deviceName.startsWith("Shape100") || deviceName.startsWith("Shape50") || deviceName.startsWith("Style100")) {
             return new BluetoothSoehnle(context);
         }
@@ -155,13 +152,16 @@ public class BluetoothFactory {
         if (deviceName.equals("AE BS-06")) {
             return new BluetoothActiveEraBF06(context);
         }
-        if (deviceName.equals("Renpho-Scale")) {
-            /* Driver for Renpho ES-WBE28, which has device name of "Renpho-Scale".
-               "Renpho-Scale" is quite generic, not sure if other Renpho scales with different
-               protocol match this name.
-             */
-            return new BluetoothRenphoScale(context);
+        if (deviceName.startsWith("QN-Scale") || deviceName.equals("Renpho-Scale")) {
+            return new BluetoothQNScale(context);
         }
+        // if (deviceName.equals("Renpho-Scale")) {
+        //     /* Driver for Renpho ES-WBE28, which has device name of "Renpho-Scale".
+        //        "Renpho-Scale" is quite generic, not sure if other Renpho scales with different
+        //        protocol match this name.
+        //      */
+        //     return new BluetoothRenphoScale(context);
+        // }
         if(deviceName.equals("ES-CS20M")){
             return new BluetoothESCS20M(context);
         }
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothQNScale.java
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothQNScale.java
@@ -77,8 +77,12 @@ public class BluetoothQNScale extends BluetoothCommunication {
 
 
     private static long MILLIS_2000_YEAR = 949334400000L;
-    private boolean hasReceived;
-    private float weightScale=100.0f;
+    // Have we received the final weight measurement
+    private boolean finalMeasureReceived;
+    // Weight scale of raw value
+    private float weightScale = 100.0f;
+    // Last seen protocol type, used in reply packets
+    private int seenProtocolType = 0;
 
 
     public BluetoothQNScale(Context context) {
@@ -86,6 +90,7 @@ public class BluetoothQNScale extends BluetoothCommunication {
     }
 
     // Includes FITINDEX ES-26M
+    // Includes some Renpho EC-CS20M
     @Override
     public String driverName() {
         return "QN Scale";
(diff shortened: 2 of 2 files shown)


keeperofdakeys:
Fixes #1126

Maybe related:
#1104

oliexdev:
Thanks for your PR, currently I am working on openScale 3.0 (see the new branch) were UUIDs are passed to BluetoothFactory for the new kotlin devices but still a lot to do. Currently I am trying to implement both handlers java and kotlin.

keeperofdakeys:
Ah that's good to hear, good luck with the work. I've been thinking about this more, and the only solution that seems reasonable for openScale 2 requires some rearchitecting (after scan match driver name with UUIDs, store driver name in config file, pass driver name to BluetoothFactory when connecting to scale).

Do you have any timeline for 3.0? If not I may try working on this. As far as I can see the above solution would only require major changes in BluetoothFactory, the rest would be relatively minor.

oliexdev:
no I don't have a timeline I have complete rewritten openScale in Kotlin but still a lot to do but you can try it https://github.com/oliexdev/openScale/tree/openScale-3.0

keeperofdakeys:
With all the changes present, my scale now works correctly without the hacks that I required before. I've successfully gotten multiple readings from my scale with this code.

However to achieve this I had to make a few changes in how bluetooth drivers are selected. Below is a list of the major changes I made.

1. Moved bluetooth drivers to their own submodule, this may come in handy for future changes.
2. Each driver now has a device ID. This is saved into the settings along with device name when selecting a scale so the correct driver can be loaded.
3. When searching for bluetooth devices, BluetoothFactory is given a list of service UUIDs to assist scale selection. (My scale requires this). This returns a device ID to save in settings.
4. Some scale subtype selection logic was moved from BluetoothFactory to individual drivers, this allowed me to standardise the Bluetooth Driver constructor to be Class(context, deviceName).

With the driver submodule, and standard constructor, this opens the door to more refactoring of BluetoothFactory in the future. Possibly using a ServiceLoader or Registry pattern, then moving all selection logic into the drivers. However I don't see any need to make these changes today.

keeperofdakeys:
I just released I have a few bugs to fix in this. Mainly transitioning config of users, since they won't have device ID saved. Plus we should get some testing of this build.

oliexdev:
Thanks but in openScale 3.0 you get already a list of UUID see https://github.com/oliexdev/openScale/blob/openScale-3.0/android_app/app/src/main/java/com/health/openscale/core/bluetooth/ScaleFactory.kt via the https://github.com/oliexdev/openScale/blob/openScale-3.0/android_app/app/src/main/java/com/health/openscale/ui/screen/bluetooth/BluetoothScannerManager.kt

I would appreciate if you could try to work on openScale 3.0 rather then openScale 2.x as this will be soon deprecated.

amitlevy21:
@oliexdev  Is there a chance to at least merge this to version 2 of the app?
since its a pretty common issue it would help to have it
of course having it for version 3 would be amazing, as this scale appears in the support devices wiki
related issues:
https://github.com/oliexdev/openScale/issues/1166
https://github.com/oliexdev/openScale/issues/1151
https://github.com/oliexdev/openScale/issues/1126
https://github.com/oliexdev/openScale/issues/1013
https://github.com/oliexdev/openScale/issues/769

keeperofdakeys:
I'll look at porting this to OpenScale 3 when I have some free time, hopefully in the next week.

oliexdev:
I have migrated this scale into the modern kotlin version, see https://github.com/oliexdev/openScale/blob/openScale-3.0/android_app/app/src/main/java/com/health/openscale/core/bluetooth/modern/ESCS20MHandler.kt

But please test it via the latest beta build on GitHub https://github.com/oliexdev/openScale/releases/tag/beta-build
Please post the log file if it's not working (which probably won't work as I don't own this scale).

amitlevy21:
Thank you! I have tested with the scale but it didn't work for me
By testing I mean:
1. opening the app
2. granting bluetooth permissions
3. turning on bluetooth
4. adding the scale
5. trying to get measurement in overview by stepping on the scale and pressing the bluetooth icon, at first, I got connecting to... and then a response saying I should step on the scale, but when I did. nothing happen
6. enable debug via Bluetooth -> Enable debug for the scale
7. enabling the debug file in general
8. repeating 5 but I only get trying to connect without responses

I don't see too much info in the log file, this is the first im trying the app so let me know if I didn't do the steps as expected.
attaching log file
[openScale_current_log (3).txt](https://github.com/user-attachments/files/22185712/openScale_current_log.3.txt)

amitlevy21:
I've done another testing, this time cleared storage
1. enable debug file
2. find the scale via bluetooth
3. connect to the scale
4. got the message to step barefoot on the scale
5. I step on the scale
6. nothing

here are the logs (I removed the scale mac address):
[openScale_current_log (4).txt](https://github.com/user-attachments/files/22185695/openScale_current_log.4.txt)

keeperofdakeys:
I had a test of this. The scale is successfully detected as QNScale, but no measurements are recorded. I see that my changes from QNScale in this PR were not merged, which explains why it's not working.

I'll rewrite this PR on top of Openscale 3.0 (basically only QNScale.java changes) and test, once everything is working I'll mark this PR as mergeable. Thanks for your help.

oliexdev:
@keeperofdakeys any progress?

amitlevy21:
@oliexdev I have the scale and can help with any kind of testing

oliexdev:
I will close this PR @keeperofdakeys please create a new one if you are able to fix this issue with the scale

PR #1243, opened 2025-12-22
QN-Scale Type 1 fix

### LINKED ISSUE #1190: QN-Scale and Renpho ES-CS20M Unable to Record Measurements
**Describe the bug**
The beta 3.0.0 app is able to pair with the scale, but does not appear to pull any data from the scale when attempting to take a measurement.

**To Reproduce**
1. Pair normally via bluetooth
2. Start connecting to scale via app
3. Step onto scale
4. Wait until measurement settles
5. Step off

Reproduced with [latest dev version](https://github.com/oliexdev/openScale/releases/tag/dev-build): `b440cf9`

**Expected behavior**
- measurements show up in UI

**Additional context**
Attached a debug log.

[openScale_current_log.txt](https://github.com/user-attachments/files/22931832/openScale_current_log.txt)

**Debug log**
Attach a debug log (see Settings -> About -> Save debug log to file) that is captured while reproducing the issue.

Stullemon:
Hi, I tried to come up with something to resolve #1190. As I pointed out there: I wasn't quite smart enough to suddenly learn Kotlin, etc. Yet, there's Anthropic's Claude (free version). So I asked what's the difference between the old Java version and the new Kotlin version. I also asked it to help me understand how the differences relate to the logging I was able to produce. I then proceeded to have it suggest a fix for it and help me build a debug build. Which it did. So I ended up testing it on my phone and - lo and behold - it worked.

What I understood kind of makes sense to me. The flow of messages was altered so it kind of revised it back to how it used to be and now it works. I couldn't test with type 2 so I cannot say if I created a regression. I am also not sure if the use of an LLM helped me comply with standards you otherwise observe. So all considered... well... I hope you can pull as is or suggest something to help me improve things to a point where you can do the pull. I would really like to continue weighing my children sometime soon. :-D

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/QNHandler.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/QNHandler.kt
@@ -34,6 +34,8 @@ import kotlin.experimental.and
  * - We subscribe to both layouts. Writes to a missing characteristic are harmless (adapter logs a warning).
  * - Weight and two resistance values are broadcast in notifications of 0xFFE1/0xFFF1 (opcode 0x10).
  * - We derive body composition with [TrisaBodyAnalyzeLib] from weight + impedance.
+ *
+ * FIXED: Protocol type race condition - now waits for 0x12 frame before sending configuration.
  */
 class QNHandler : ScaleDeviceHandler() {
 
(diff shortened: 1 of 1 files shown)


Stullemon:
Oh, one thing I also noticed is that Claude was suggesting the following change. The snippet

```
        if (weightKg <= 5f || weightKg >= 250f) {
            weightKg = weightKg / 10.0f
```
was suggested to be replaced by 

```
        if (weightKg <= 5f || weightKg >= 250f) {
            weightKg = raw / 10.0f
            logD("QN: weight adjustment applied, new weight=$weightKg kg")
```
Feels wrong judging by the fact that it pretty much went unchanged during the conversion from Java to Kotlin. Just thought I'd still share. :-)

oliexdev:
Thanks for the fix, I can't test it as I don't own this scale but if it works.

auspiciouslink:
Does this mean a dev version should have this fix in it? I have the most recent version available in F-Droid, 3.0 from 2025-12-01, and as of 2026-01-01, it was not working with the QN-Scale.

PR #1267, opened 2026-01-05
update bmi agebands

Montyr:
source for kids:
  since ageband does not allow for months, i took the lowest and highest value for that year
  https://www.who.int/tools/growth-reference-data-for-5to19-years/indicators/bmi-for-age

definition for adults:
  https://www.cdc.gov/bmi/adult-calculator/bmi-categories.html
  
and fixed a typo


oliexdev:
Thanks for your PR!
Could you please add a short comment in the code referencing the source/specification?
That would be awesome and help others understand where this behavior comes from later on.

oliexdev:
Thanks :+1:

Issue #671, opened 2020-12-22
US Navy body fat estimate


jimr1603:
The formulas for estimating body fat percentage (BF%) from age, sex and weight are good, but I would like to improve accuracy by including waist (& hip, female) & neck measurements. 

I prefer this measure to calipers since I can get the same measurement twice in a row! 

The formulas aren't complicated - [this site](https://www.calculatorpro.com/calculator/us-navy-body-fat-calculator/) has them in inches & lbs. If you like the concept I'm happy to dig out the paper the USA DoD wrote about the method for the citations in the project. 

I'd offer to code it but I've never managed to get something to compile in Android Studio.

oliexdev:
Currently I used the caliper measurement formula by Jackson, Polloc but another formula could be implemented as an option in the caliper measurement settings in openScale like for the body water measurement. If you want to do it your are welcome to send a PR.

jimr1603:
I got the latest commit of the app to build on my PC, so I'll give it a
try.

>

oliexdev:
Since there has been no progress on this issue for a long time, I am closing it for now.

PR #1311, opened 2026-02-19
Implementing body fat percentage calculation with US Navy method

JPFrancoia:
This PR implements the US Navy body fat % measurement method, [docs](https://www.wikihow.com/Measure-Body-Fat-Using-the-US-Navy-Method). It relies on:

For males:
- Neck measurement
- Waist measurement

For females:
- Same as for men + hip measurement

I use this method because I do a lot of sports and the impedance measurement of my scale is wildly off (8-10% higher than reality). The US Navy method is closer to reality.

Someone already asked for this in the past: https://github.com/oliexdev/openScale/issues/671.

The metric is auto-calculated if the neck and waist measurements (and hips) exist. I tested it on myself (male) and my partner (female). It works more or less the same way as "Fat caliper".

Screenshots:
Screenshot_20260219-192404_openScale

### DIFF
--- a/android_app/app/schemas/com.health.openscale.core.database.AppDatabase/15.json
+++ b/android_app/app/schemas/com.health.openscale.core.database.AppDatabase/15.json
@@ -1,0 +1,411 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 15,
+    "identityHash": "2c6d48ddefaa85707983de54030ef0e6",
+    "entities": [
+      {
+        "tableName": "User",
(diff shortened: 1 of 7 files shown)


oliexdev:
Thanks for your effort but openScale has an easier method for that, see commit https://github.com/oliexdev/openScale/commit/91a72db2c0c64aeee7192cc362030b96b446c03c 
Please try it out you have just change the calculation method for body fat in the settings->measurement types->body fat formula to US Navy

JPFrancoia:
Thank you! I'm away on holiday for now but I'll have a look when I have the chance. Thanks for the help

PR #1316, opened 2026-02-25
realme Smart Scale model RMH2011

qumuq-til:
Adding support for RMH2011 scale in OpenScale.


oliexdev:
Thanks for your PR :+1:

Issue #1307, opened 2026-02-18
Setup issues with BF600


caoilte:
### Preflight Check (Required)

- [x] I tested the latest dev version https://github.com/oliexdev/openScale/releases/tag/dev-build
- [x] The issue still occurs in the dev version
- [x] I searched the existing open and closed issues
- [x] This issue has not already been reported or fixed

### Bug Description

I got a BF600 working but there were some challenges that are sub-optimal.

**Problem 1 - OpenSync cannot setup the first profile** 

I was unable to get the device to work out of the box. OpenSync finds it in bluetooth settings, asks you to stand on the scales and then does nothing.

The only way I got OpenSync to work was by setting up a profile using Beurer Health Manager (possibly I could also have done it on device). Once a profile exists OpenSync pops up a dialog asking if you want to use the existing profile or to create a new one. 

I have included the debug log for connecting to a wiped device

**Problem 2 - Using an existing profile is non-intuitive**

In order to use an existing profile you need to enter a pin number. Finding this pin number is non-intuitive and it would be better if the app told you to get it by pressing the "Set" button on the scale.

**Problem 3 - Failing to enter a correct pin puts you in an unusable state**

If you don't enter the pin correctly you get a blue informational popup each time you connect telling you you need to enter the pin number, but are not provided with a way to do so. Resetting the app data seems to be the only way to move forward.

**But it appears to work**

Despite these problems I do appear to have got the device working. There's are small dev fixes for all of these problems I'm sure and just having an open issue is bound to be useful to anyone struggling with this or other similar devices.

### Steps to Reproduce

1. Factory Reset
2. Pair device
3. Create User
4. Connect via bluetooth

At this point you are asked to step onto the scales but nothing further happens.

### Debug Log (Required)

```shell
| Field | Value |
|---|---|
| Time | 2026-02-18 12:40:39.674 |
| Timezone | Europe/London (UTC 0h) |
| Session ID | 19c70c44ffc |
| App | openScale |
| Version | 3.0.2-debug (72) |
| Package | com.health.openscale.debug |
| Build Type | debug |
| Device | OnePlus IV2201 |
| Android | 13 (API 33) |
| Build ID | IV2201_11_F.62 |
| Build Time | 2026-02-18T12:14:40Z |
| Git SHA | 6ced0e9 |


+ 2026-02-18 12:40:39.661 I/openScaleLog: File logging preference updated to: true (was: false)
[...]

Issue #1236, opened 2025-12-14
Cannot change wrong consent code


Focshole:
**Describe the bug**
Once a wrong consent code is set, there is no way to reset it

**To Reproduce**
1. Pair a new Beurer 720 with existing users
2. Try to pair it but use the wrong code
3. Remove pairing after wrong code
4. Retry to pair it and see that it says the consent code is wrong but you cannot update it

Reproduced with 3.0

**Expected behavior**
You should be able to change it

oliexdev:
read the FAQ

Focshole:
Sorry, I did not read the FAQ part.

After reading it, I was still unable to reset the code in the app.
In the end I have been able to update the consent code without resettting the scale by:
1. Backing up the app DB
2. Uninstall the app (or at least we should delete the app data from android I think!)
3. Reinstall the app and import the DB backup

In this way I get the window on the app to write the consent code.
Thank you, the FAQ was helpful 👍

kri164:
Thank you @Focshole, your post help me very much. Reading FAQ doesnt help me.
You dont need uninstall app, cleaning app storage is enought.

1. Backup DB
2. Clean app storage
3. Restore DB
4. Enter consent code listed from weight device

Issue #1216, opened 2025-12-03
SBF77 Can't get new measurements from scale on v3


Olangu:
**Describe the bug**
Can't get new measurements from scale on v3

**To Reproduce**
Connect to Scale
Step on scale

Reproduced with [latest dev version](https://github.com/oliexdev/openScale/releases/tag/dev-build): 3.0 (fdroid) 

**Expected behavior**
New measurement should be added to the app. 

**Additional context**

**Debug log**
[openScale_current_log.txt](https://github.com/user-attachments/files/23899175/openScale_current_log.txt)

kri164:
Same problem there.
App say me to Enter conset to slot 1.
What is a consent and how to enter this?

oliexdev:
If you don’t remember the old consent code, it needs to be entered again. To do this, the user must first be deleted on the scale (please follow the instructions in your scale’s manual). After that, you can reconnect the scale to openScale, ideally using a new openScale user.

bsimmo:
That's not quite true @oliexdev 
When OpenScale was promoted to release, for some reason it didn't work.
It does now.
(Which is odd as it worked before that).
My wife couldn't get it to work, neither could I. But all working now.

Anyway.
Consent code you tap the up/down arrows on the scale to select the user, then press the middle button. It will then give you the code.
Type that in.


If anyone knows how to delete a user, that would be great. The only way is a full reset wiping it all, and nobody wants to do that (here).
I now have 6 or so users from all the beta testing, 😆

Olangu:
Where would I type that in?

I think I just tried with 0000 the first time.
And now I can't change it, even if I remove the scale in OpenScale and try to add it anew.

Why is there no setting in config for the scale?

bsimmo:
Once you have connected the scale to Bluetooth on the phone, with that PIN.

When you then go and detect the BT devices and select you scale.
You go back to the main font page with the BT icon, you wait and on first
connect it should ask you to select which user on the scale you want (it
says P01, P02 etc in V3, on V2 it used to give the names on the scale.).
It then asks for a confirmation code, that when you go and get it from the
user on the scale and type it in.
It took me months to find that and many questions on here back in the V2
era of OpenScale.

Ben.

Olangu:
I never get to choose a pin. I think it remembers my old faulty pin.

oliexdev:
[...]

Issue #1187, opened 2025-10-11
Beurer BF 720 in 3.0.0 (69)


helioshk:
When connecting to the scale using version 2.5.4, I get asked for the pin and everything works.
In 3.0.0, it connects successfully to the scale, but I don't get asked for a pin and I don't get any measurements.

b9chomps:
> When connecting to the scale using version 2.5.4, I get asked for the pin and everything works.
In 3.0.0, it connects successfully to the scale, but I don't get asked for a pin and I don't get any measurements.

I have the same issue after the latest update. I was not prompted for my PIN. No connection with my user on the scale is established and no measurements are synced.
I reverted from version code 69 to 68 (Aurora Store) and was prompted for my PIN after the initial setup.

oliexdev:
please provide the log files

helioshk:
I don't see much in the log:

```diff
? 2025-10-11 10:34:16.532 D/openScaleLog: Initial markdown-diff headers written to: /storage/emulated/0/Android/data/com.health.openscale.oss/files/logs/openScale_current_log.txt
+ 2025-10-11 10:34:16.534 I/openScaleLog: File logging enabled during runtime – started fresh session log.
? 2025-10-11 10:34:16.534 D/openScaleLog: Initial markdown-diff headers written to: /storage/emulated/0/Android/data/com.health.openscale.oss/files/logs/openScale_current_log.txt

```

oliexdev:
you have to connect to your scale while logging is enabled

oliexdev:
you are using the debug handler and not posted the whole log file...

bsimmo:
@helioshk @b9chomps 
Is this working for you yet?
Currently not on the SBF77 but working through getting it working, so it might be by the time you read.

The logs needed are just turn writing to a file on in Settings.

<img width="1220" height="844" alt="Image" src="https://github.com/user-attachments/assets/11a1581f-366d-4ba4-8ce0-910552ca72af" />
BUT _do not_ turn on the 'debug' in the BT scale select, that is for something else.

Once enabled, close the scale program.
Start it again, go through the method you used, when it fails....Wait a bit then go back to settings and save, then copy/paste the code (I think you can also attach with Add File on the post if wanted)

helioshk:
Is this the correct one?
(I replaced height with "hhh.h" and age with "aa")

[openScale_current_log _1_.txt](https://github.com/user-attachments/files/22989526/openScale_current_log._1_.txt)

[...]

PR #1317, opened 2026-02-26
Fix consent code not being cleared on rejection or device removal. 

Olangu:
When a scale rejects a wrong consent code, the stored code is now reset and the user is re-prompted immediately. Removing a saved device now also clears all persisted BLE driver settings (consent codes, user-slot mappings) so re-pairing starts fresh.

Issues:
#1216 #1236 #1307 #1187 #1157


oliexdev:
Thanks for your PR :+1:

PR #1300, opened 2026-01-31
Renpho ES-CS20M communication fixes (with Claude help)

davidhildebrand:
This fixes issue #1190 for Renpho ES-CS20M scales, similar to the PR #1243 fix for QN-Scale type 1 scales.  Relied heavily on assistance from Claude Code.

PR #1330, opened 2026-03-11
Add impedance/resistance measurements in Renpho ES-CS20M Handler

davidhildebrand:
These changes (with Claude help) fix impedance/resistance measurements from the Renpho ES-CS20M in addition to the working weight measurements.  
Should have though to test this before PR #1300 but didn't think of it.  

The body fat % calculation in this case uses the formula in TrisaBodyAnalyzeLib because it most closely—though not perfectly—matched my measurements with what the scale display reported.  Unfortunately, Renpho Customer Support <support@renpho.com> refused to share their actual formula: "We apologize, but our proprietary formulation cannot be disclosed."


oliexdev:
Thanks for your PR, did you test it probably? Does it work smoothly as it has been heavily refactored by Claude.

davidhildebrand:
Yes, tested several times.  

In addition to recording impedance/resistance, there were some fixes to correct assumptions previously made based on using a single user profile to debug.  Varying the user parameters (male vs female, height changes, athlete mode, etc) to understand the impact they had on body fat percentage calculation revealed these errors.  

I also tried to get the actual formula used by Renpho for the body fat calculation but support wouldn't share it.

Thanks for reviewing and for the great project!

oliexdev:
Thanks for your PR :+1:

Issue #1061, opened 2024-08-23
Support for "Crenot GoFit S2"


sroemer:
I am looking into getting openScale to work with a "Crenot GoFit S2" scale, which I recently bought on [Lazada](https://www.lazada.vn/products/can-suc-khoe-dien-tu-thong-minh-crenot-gofit-ket-noi-bluetooth-do-luong-phan-tich-chi-so-co-the-body-fat-scale-can-dien-tu-suc-khoe-thong-minh-do-mo-bao-hanh-12-thang-i2059623990-s12266002911.html) in Vietnam.
The [Crenot website](https://crenot.com/) currently only shows an older version of the S2 scale, which has a simpler display.
The scale uses the [Fitdays](https://play.google.com/store/apps/details?id=cn.fitdays.fitdays&hl=en_US) app.

Here pictures of my scale and its packaging:
![crenot_gofit_s2](https://github.com/user-attachments/assets/d7fbb931-f1d2-4964-9789-acb2a63b3551)
![crenot_gofit_s2_box](https://github.com/user-attachments/assets/d4dd521c-57f5-495a-9402-faa9964e0f39)

Today I gathered some initial information with BLE Scanner and looked if I find any scale sharing similarities, but so far didn't see one (I am totally new to bluetooth stuff and might have missed that, so if anybody sees similarities or has additional helpful input I am very happy to hear about it). After the weekend I will set up my development environment for this and also gather Bluetooth HCI logs and OpenScale logs.

Here my BLE Scanner screenshots (not on the screenshots, but please note that the manufacturer of the scale is "ICOMON" and the model name is "FG2210LB" - maybe that can be useful for supporting similar scales with different branding?):

![crenot_gofit_s2_ble_device_information](https://github.com/user-attachments/assets/4838a9e1-978b-4c83-b1bf-9fa6012ac98c)
![crenot_gofit_s2_ble_generic_access](https://github.com/user-attachments/assets/48ca16de-57ee-479a-a09b-d110fcfd4836)
![crenot_gofit_s2_ble_generic_attribute](https://github.com/user-attachments/assets/b16cb4fb-5680-4d93-b294-bc43349ee7e5)
![crenot_gofit_s2_ble_custom_service_1](https://github.com/user-attachments/assets/5b775df4-8f20-4fb7-bf53-a95b2cd6facf)
[...]

PR #1337, opened 2026-03-23
Add support for Runstar R5

babbitt:
Hi there! Wanted to reopen #1061 with support for the RunStar R5 (which appears to be a rebrand of the same scale). 

> Note: 
> While I am a full time software engineer I don't usually write Kotlin; this PR is entirely vibe-coded (but has been tested on a real device with a real RunStar R5 scale and works as expected).  I'm happy to take any feedback or changes that should be made.


oliexdev:
Thanks for your PR :+1:

PR #1345, opened 2026-04-06
feat(bluetooth): add broadcast handler for newer Icomon/e-volve scales

BramB-1952444:
Fixes an issue where newer Icomon boards (broadcasting as "AAA002") fail with an UNKNOWN_STATUS_CODE during GATT connection attempts. This commit bypasses the connection entirely and passively reads the BLE advertisement.

- Added `IcomonBroadcastHandler` using LinkMode.BROADCAST_ONLY.
- Implemented the 0xA0 XOR cipher to decrypt weight and impedance from the 0xA0AC manufacturer data payload.
- Placed the new handler at the top of `ScaleFactory.kt` so it intercepts the AAA002 broadcast before the legacy GATT handler attempts to connect.

Note: The protocol reverse-engineering and Kotlin implementation for this commit were developed with the assistance of generative AI.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/ScaleFactory.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/ScaleFactory.kt
@@ -61,6 +61,7 @@ import com.health.openscale.core.bluetooth.scales.StandardBeurerSanitasHandler
 import com.health.openscale.core.bluetooth.scales.TrisaBodyAnalyzeHandler
 import com.health.openscale.core.bluetooth.scales.TuningProfile
 import com.health.openscale.core.bluetooth.scales.YunmaiHandler
+import com.health.openscale.core.bluetooth.scales.IcomonBroadcastHandler
 import com.health.openscale.core.facade.MeasurementFacade
 import com.health.openscale.core.facade.SettingsFacade
 import com.health.openscale.core.facade.UserFacade
@@ -89,6 +90,7 @@ class ScaleFactory @Inject constructor(
 
     // List of modern Kotlin-based device handlers.
     private val modernKotlinHandlers: List<ScaleDeviceHandler> = listOf(
+        IcomonBroadcastHandler(),
         RealmeScaleHandler(), // Added Realme handler here
         YunmaiHandler(isMini = false),
         YunmaiHandler(isMini = true),
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/IcomonBroadcastHandler.main.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/IcomonBroadcastHandler.main.kt
@@ -1,0 +1,79 @@
+package com.health.openscale.core.bluetooth.scales
+
+import android.bluetooth.le.ScanResult
+import com.health.openscale.core.bluetooth.data.ScaleMeasurement
+import com.health.openscale.core.bluetooth.data.ScaleUser
+import com.health.openscale.core.service.ScannedDeviceInfo
+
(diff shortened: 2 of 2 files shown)


oliexdev:
Thanks for your PR but what about the already implemented AAAxHandler? Does it not work? I would prefer to fix that implementation.

valenipes:
Hi @oliexdev, some days ago I opened the issue #1341 because when I try to pair the scale I see this
pair
and when I pair and try the scale, nothing happens. 

I noticed later that a while ago a pr was done #1041, and if I understand correctly the measures should go, but they don't. So yes, probably there is some issue with the already implemented AAAxHandler (or maybe I'm missing something on how to use the app properly)

oliexdev:
@valenipes did you try out this PR?

valenipes:
Just tried, it doesn't work. Btw yesterday I noticed that the handler "SWAN / Icomon (MGB)" is probably not correct for the scale, because in `app/src/main/java/com/health/openscale/core/bluetooth/scales/` there is a specific handler for AAA-Family, so the handler is somehow evaluated incorrectly in `app/src/main/java/com/health/openscale/core/bluetooth/scales/MGBHandler.kt` before it's even evaluated by `AAAxHandler.kt`. 

So what I did is: I removed previous existing MGBHandler.kt and IcomonBroadcastHandler.main.kt by @BramB-1952444, and I used the same function `supportFor` written by @BramB-1952444, and the handler was now recognized correctly. The problem is that, even after that, body weight values aren't displayed in the app, so probably there is something else wrong in the `AAAxHandler.kt` that will require time to fix because I'm not a developer.

To summarize: 
1. there is some issue in the `MGBHandler.kt`, because it tries to handle AAAx scales and fails
2. `AAAxHandler.kt` have some issue because the app don't show body weight measures even when `AAAxHandler.kt` handles the AAA002 scale

oliexdev:
It seems your AAA002 scale has as a bluetooth name that only **starts** with AAA002. Could you do Step 1.1 of the [wiki](https://github.com/oliexdev/openScale/wiki/How-to-support-a-new-scale#part-1-analyzing-the-bluetooth-protocol) and post the result here. But it seems you have a newer protocoll version of AAA002 which is currently not supported if this PR is also not working.
The MGBHandler shouldn't be active for the AAAx scales.

valenipes:
Here is the log file:
[openScale_current_log (3).txt](https://github.com/user-attachments/files/26964711/openScale_current_log.3.txt)
Regarding the name, it seems to me that the name is exactly "AAA002" and doesn't only starts with it, my suspect is that in [MGBHandler.kt ](https://github.com/oliexdev/openScale/blob/master/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/MGBHandler.kt ), for some reason the boolean variable `serviceMatch` is true for my scale (obviously that can't be true for the variable `nameMatch`) and this condition fails:
`if (!nameMatch && !serviceMatch) return null`

Another strange thing is that the scale's model (FG263LB) is the same of the one [Duncaen](https://github.com/Duncaen) had and made the pr #1041 with, and thus it isn't so new (it should've been bought in 2021/2022).

oliexdev:
unfortunately the log didn't list any GATT services and the name of your scale. Try to find your scale with https://play.google.com/store/apps/details?id=com.macdom.ble.blescanner and post the screenshots here.

oliexdev:
Since there has been no progress on this PR for a long time, I am closing it for now.  

Please note that implementation of new scale support requires access to the **actual scale**.

If someone wants to continue working on this, feel free to **reopen this PR** or submit a working implementation via a **pull request**.

For guidance on how to add a new scale handler into openScale, see the updated wiki page:  **[How to support a new scale](https://github.com/oliexdev/openScale/wiki/How-to-support-a-new-scale)**

PR #1346, opened 2026-04-06
Implemented StandardImpedanceLib and switched Etekcity ESF-551 to use that lib

wkornewald:
The measurements at least look plausible for my body. I've linked to the respective research and they all use the same H^2/R factor, so that's at least consistent. The unit test also checks if we're in the right ballpark of TBW / FFM = 0.732.

Only bone mass is an estimate based on FFM percentage, which should work well enough, but in a future PR I'd like to find a better formula which takes age into account.

Visceral fat is missing completely since it's hard to estimate the fat distribution in the body based on the few values we measure. It might make sense to skip that completely, at least until we find a good formula.

The BMR value is still ignored because openScale hard-codes its own calculation. The Katch-McArdle formula is 5kcal lower for me than the hard-coded formula. The difference is bigger with higher body fat.

Could you please also test this lib with your scale and see if the values make sense? I also still plan to take my scale to a few friends and do measurements with them at least to double-check for plausibility.

Also, the fat percentage is based on FFM instead of LBM. I originally wanted to include the following code which is based on EtekcityLib, but couldn't find a respective paper for LBM. The FFM based value for my own measurement falls more nicely within openScale's value range, but most online sources and the U.S. Navy value ranges at least work with lower values which seem to be LBM based instead of FFM. Not sure if we should add LBM as an estimate based on FFM, but simply subtract 2-3 percent points for men (if I understand the literature correctly that's the expected difference? and for women it was around twice as high).

```kotlin
    /**
     * Body fat percentage based on [EtekcityLib].
     *
     * This formula is not really scientific, but currently the best we have.
     *
     * It seems to align with the U.S. Navy value ranges and these sources:
     * https://www.newhealthadvisor.org/Body-Fat-Percentage-Chart.html
     * https://www.medicalnewstoday.com/articles/body-fat-percentage-chart#men
     * https://www.calculator.net/body-fat-calculator.html
     *
     * Also, see this paper for details about the confusion around FFM (fat-free mass) vs. LBM (lean body mass):
     * https://pmc.ncbi.nlm.nih.gov/articles/PMC11625996/
     */
    val bodyFatPercentageWithoutEssentialFats: Double by lazy {
        val ageFactor = if (isMale) 0.103 else 0.097
        val bmiFactor = if (isMale) 1.524 else 1.545
        val constant = if (isMale) 22.0 else 12.7
        ageFactor * age + bmiFactor * bmi - 500.0 / impedance - constant
    }

    /**
     * LBM / lean body mass.
     *
     * Note: Based on the assumption that [bodyFatPercentageWithoutEssentialFats] doesn't contain essential lipids.
     * Also, see this paper for details about the confusion around FFM (fat-free mass) vs. LBM (lean body mass):
     * https://pmc.ncbi.nlm.nih.gov/articles/PMC11625996/
     */
    val leanBodyMass: Double = weightKg * (1 - bodyFatPercentageWithoutEssentialFats / 100)
```

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/libs/StandardImpedanceLib.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/libs/StandardImpedanceLib.kt
@@ -1,0 +1,133 @@
+/*
+ * openScale
+ * Copyright (C) 2025 olie.xdev <olie.xdeveloper@googlemail.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package com.health.openscale.core.bluetooth.libs
+
+import com.health.openscale.core.data.GenderType
+
+/**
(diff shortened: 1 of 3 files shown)


oliexdev:

Issue #1314, opened 2026-02-22
Xiaomi Scale 2 still can't connect since update to 3.x


QuadCorei8085:
### Preflight Check (Required)

- [x] I tested the latest dev version https://github.com/oliexdev/openScale/releases/tag/dev-build
- [x] The issue still occurs in the dev version
- [x] I searched the existing open and closed issues
- [x] This issue has not already been reported or fixed

### Bug Description

Issue got closed for Xiaomi Scale 2 but I do not see the resolution.
Since 3.x it stopped working.

### Steps to Reproduce

Start the app
Hit the Bluetooth icon
Step on xiaomi scale 2

No measurement reported

### Debug Log (Required)

```shell

```diff
? 2026-02-22 22:26:48.214 D/openScaleLog: Markdown diff block closed.
? 2026-02-22 22:26:48.219 D/openScaleLog: Initial markdown-diff headers written to: /storage/emulated/0/Android/data/com.health.openscale/files/logs/openScale_current_log.txt
? 2026-02-22 22:26:48.220 D/openScaleLog: Initial markdown-diff headers written to: /storage/emulated/0/Android/data/com.health.openscale/files/logs/openScale_current_log.txt
+ 2026-02-22 22:26:48.221 I/openScaleLog: File logging enabled during runtime – started fresh session log.
+ 2026-02-22 22:26:51.263 I/GattScaleAdapter: Disconnected 5C:64:F3:6D:58:A9: REMOTE_USER_TERMINATED_CONNECTION
? 2026-02-22 22:26:51.264 D/MiScaleHandler: handleDisconnected()
? 2026-02-22 22:26:51.265 D/MiScaleHandler: detach()
? 2026-02-22 22:26:51.266 D/BleConnector: BluetoothEvent received: Disconnected(deviceAddress=5C:64:F3:6D:58:A9, reason=REMOTE_USER_TERMINATED_CONNECTION) for MI SCALE2
+ 2026-02-22 22:26:51.266 I/BleConnector: Event: Disconnected from 5C:64:F3:6D:58:A9. Reason: REMOTE_USER_TERMINATED_CONNECTION
? 2026-02-22 22:26:51.267 D/BleConnector: Disconnected event: Releasing active communicator: GattScaleAdapter
? 2026-02-22 22:26:51.268 D/BleConnector: Disconnected event: Active communicator released and set to null.
[...]

PR #1349, opened 2026-04-10
Fix for Mi Scale 2 measurements not being considered

KonradIT:
Since v3.x.x measurements from Mi Scale 2 were silently discarded due to the magic byte being considered. Issue https://github.com/oliexdev/openScale/issues/1314 was closed prematurely without a proper fix. Issue persisted as of today on 3.0.3.


oliexdev:
Thanks for your PR :+1:

Issue #1308, opened 2026-02-18
ES-CS20M: broadcast-only variant (non-connectable) findings


KristianP26:
### Cross-project findings: ES-CS20M hardware variants

I maintain [ble-scale-sync](https://github.com/KristianP26/ble-scale-sync) (Node.js/TypeScript alternative for BLE scales) and we've been debugging the same ES-CS20M problems in [our issue #34](https://github.com/KristianP26/ble-scale-sync/issues/34). Sharing our findings here since they're relevant to openScale users.

#### The ES-CS20M ships in at least 3 hardware variants

| Variant | Advertised name | Protocol | Services | GATT connectable |
|---------|----------------|----------|----------|-----------------|
| 1 | "QN-Scale" | QN/Qingniu | 0xFFE0 / 0xFFF0 | Yes |
| 2 | "ES-CS20M" or unnamed | Yunmai | 0x1A10 | Yes |
| 3 | "Renpho Scale" or unnamed | QN/Qingniu (broadcast) | none | **No** (ADV_NONCONN_IND) |

Variant 1 works with the QN handler. Variant 2 was fixed by PR #1300. Variant 3 is the problematic one.

#### Variant 3: broadcast-only (non-connectable)

Some ES-CS20M units advertise as **non-connectable** (`ADV_NONCONN_IND`). No BLE stack on any OS can establish a GATT connection to these devices. This explains the "unable to connect" reports from users where the scale is visible but connection always fails, on both Windows and Android.

These devices use QN/Qingniu manufacturer data advertisements with the `AABB` marker:

```
FF FF AA BB [MAC 6 bytes] [data bytes...] [stability byte near end]
```

**Impedance is not available** in broadcast mode (confirmed independently by two users).

**Weight byte position is still unconfirmed.** Our initial assumption (bytes [10-11] as big-endian uint16 / 100) produced incorrect readings: 135 kg for a user who actually weighs ~70 kg. We are currently collecting controlled data samples (known weights + raw hex) from a user with this variant to determine the correct byte positions and divisor. Will update once the format is confirmed.

Two raw samples captured so far:

```
Idle (stable):      FF FF AA BB ED 67 39 DC 60 60 08 D6 00 00 FF FF FF 02 00 18 1A 4C 09 03 FC 01
With weight (meas): FF FF AA BB ED 67 39 DC 60 60 13 00 00 00 FF FF FF 02 00 55 05 4C 09 03 27 00
```

[...]

PR #1301, opened 2026-02-02
Fix Renpho ES-30M (QN-Scale) support

8bitgentleman:
I recently got an ES-30M scale and since it wasn't working with openScale decided to do some investigating. The ES-30M scale was connecting but not recording measurements due to incomplete protocol handshake and incorrect frame parsing.

Fixes #986 

### Problem                                                                       
  The ES-30M scale (advertises as "QN-Scale") was connecting successfully but not   
  recording measurements. The app showed "step on barefoot" but weight data never   
  appeared.                                                                         
                                                                                    
  ### Root Cause                                                                    
  1. **Incomplete handshake**: The scale requires specific responses to 0x21 opcode 
  before sending weight data                                                        
  2. **Wrong frame format**: ES-30M uses a different byte layout than other QN      
  scales                                                                            
                                                                                    
  ### Solution                                                                      
                                                                                    
  #### 1. Fixed Protocol Handshake                                                  
  - **0x21 handler**: Now sends TWO 0xA0 responses (was sending one incorrect       
  response)                                                                         
  - **0x14 handler**: Sends proper 0x20 time sync with real timestamp (was sending  
  fixed bytes)                                                                      
  - **Added 0x22 query**: After 0xA0 responses, queries stored data                 
                                                                                    
  #### 2. Added Format Detection for 0x10 Weight Frames                             
                                                                                    
  ES-30M uses a different byte layout:                                              
  ES-30M format:                                                                    
    byte[3] = unit (kg/lb)                                                          
    byte[4] = stable flag (0x00/0x01/0x02)                                          
    bytes[5,6] = weight (big-endian)                                                
    bytes[7-10] = resistances                                                       
                                                                                    
  Original QN format:                                                               
    bytes[3,4] = weight (big-endian)                                                
    byte[5] = stable flag (0x01)                                                    
    bytes[6-9] = resistances                                                        
                                                                                    
  **Detection logic**: If `byte[4] <= 0x02` AND `weightScaleFactor == 10.0`, use    
  ES-30M format.                                                                    
                                                                                    
  #### 3. Added Opcode Handlers                                                     
  - 0x23 (user data frames)                                                         
  - 0xA1, 0xA3 (acknowledgments)                                                    
                                                                                    
  ### Backward Compatibility                                                        
  Format detection ensures existing QN scales continue working with the original    
  format. ES-30M is auto-detected based on:                                         
  - Byte pattern (stable flag at byte[4])                                           
[...]

PR #1353, opened 2026-04-17
Fix QNHandler: enable live weight for Renpho ES-CS20M

Ebonhawk3829:
# Renpho ES-CS20M: Fix QNHandler to deliver live weight (0x10) frames

## Problem
The Renpho ES-CS20M scale connects via QNHandler and completes the BLE handshake, but no live weight measurement frames (opcode 0x10) ever arrive. The scale goes silent after the 0x23 stored-data response.

## Root Cause (from btsnooz log analysis)
Four issues in QNHandler.kt prevent the scale from entering live measurement mode:

1. **Extra 0xA0 sub-opcode 0x04 frame**: The handler sends two 0xA0 frames (sub-opcodes 0x04 then 0x02). The extra 0x04 frame is not in the working protocol and prevents the scale from transitioning to measurement mode.

2. **Premature 0x22 query**: The 0x22 stored-data query was sent immediately after 0xA0, before receiving the 0xA1 acknowledgment. The scale requires 0xA1 ACK before accepting a 0x22 query.

3. **Wrong 0x22 payload**: Byte[4] of the 0x22 frame was 0x03 instead of 0x01 per the working reference implementation.

4. **Weight scale factor**: Type-2 (FFF0 service) scales with protocol byte 0xFF report `weightScaleFactor=10` in the 0x12 frame, but the actual conversion is `raw / 100` (i.e., an additional `/10` is needed).

### Fix 1: Send only a single 0xA0 with sub-opcode 0x02
In the `0x21` handler, replace the two-frame 0xA0 sequence with a single 0xA0 frame using sub-opcode 0x02 and payload bytes from the working capture (`FE FF EE 01 1C 06 86 03 02`).

### Fix 2: Move 0x22 query to 0xA1 handler
Instead of sending 0x22 immediately after 0xA0, add an `0xA1` handler that sends the 0x22 stored-data query only after receiving the scale's acknowledgment.

### Fix 3: Correct 0x22 payload byte[4]
Change byte[4] from `0x03` to `0x01` per the working implementation.

### Fix 4: Weight correction for type-2 protocol 0xFF
After `weightKg = raw / weightScaleFactor`, add an additional `/10` when `!likelyUseType1 && weightScaleFactor == 10.0f`, giving the correct `/100` total for kg.

## Reference
Based on keeperofdakeys' working Java implementation from PR #1152 and verified with BLE btsnooz log captures.

## Testing
Tested on a physical Renpho ES-CS20M (firmware V10.0, BLE name "Renpho-Scale"). After the fix, 0x10 frames arrive and weight is recorded correctly.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/QNHandler.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/QNHandler.kt
@@ -203,34 +203,88 @@ class QNHandler : ScaleDeviceHandler() {
             }
             0x12 -> handleScaleInfoFrame(data)         // scale factor setup
             0x21 -> {
-                // Reply with a single 0xA0 user-profile frame (sub-opcode 0x02).
-                // Matches the proven working sequence from keeperofdakeys (PR #1152).
-                // NOTE: Do NOT send the 0x04 sub-opcode — the extra frame prevents
-                // the scale from entering live measurement mode.
-                // NOTE: Do NOT send 0x22 here — wait for the 0xA1 ACK first.
-                logD("QN: received 0x21 frame, sending single 0xA0 response")
+                // ES-30M needs two 0xA0 frames (sub-opcodes 0x04 then 0x02) plus
+                // an immediate 0x22 query.  Other QN scales (e.g. ES-CS20M) need
+                // only one 0xA0 (sub-opcode 0x02) and must wait for the 0xA1 ACK
+                // before sending 0x22.
+                // Detection: reuse the same heuristic from #1301's 0x10 parser.
+                // The 0x12 scale-info frame sets weightScaleFactor; ES-30M reports
+                // weightScaleFactor == 10.  If 0x12 hasn't arrived yet the default
+                // (1.0) will keep us on the original path, which is safe.
+                val isES30M = weightScaleFactor == 10.0f
 
-                val msg = byteArrayOf(
-                    0xa0.toByte(), // Opcode
-                    0x0d,          // Length (13 bytes)
-                    0x02,          // Sub-opcode type
-                    0xfe.toByte(), // 0xFE = no specific user slot
-                    0xff.toByte(), // Payload bytes from working capture
-                    0xee.toByte(),
-                    0x01,
-                    0x1c,
-                    0x06,
-                    0x86.toByte(),
-                    0x03,
-                    0x02,
-                    0x00           // Checksum placeholder
-                )
-                msg[msg.lastIndex] = checksum(msg, 0, msg.lastIndex - 1)
+                if (isES30M) {
+                    // ---- ES-30M path (from PR #1301) ----
+                    logD("QN: received 0x21 frame, sending TWO 0xA0 responses (ES-30M path)")
 
-                if (hasCharacteristic(SVC_T2, CHR_T2_WRITE_SHARED)) {
-                    writeTo(SVC_T2, CHR_T2_WRITE_SHARED, msg, true)
-                } else if (hasCharacteristic(SVC_T1, CHR_T1_WRITE_CONFIG)) {
-                    writeTo(SVC_T1, CHR_T1_WRITE_CONFIG, msg, true)
+                    // Response 1: sub-opcode 0x04
+                    val msg1 = byteArrayOf(
+                        0xa0.toByte(), 0x0d,
+                        0x04,                   // sub-opcode
+                        0xfe.toByte(),
+                        0x00, 0x00, 0x00, 0x00,
+                        0x00, 0x00, 0x00, 0x00,
+                        0x00                    // checksum placeholder
+                    )
+                    msg1[msg1.lastIndex] = checksum(msg1, 0, msg1.lastIndex - 1)
+
+                    // Response 2: sub-opcode 0x02
+                    val msg2 = byteArrayOf(
+                        0xa0.toByte(), 0x0d,
+                        0x02,                   // sub-opcode
+                        0x01, 0x00, 0x08, 0x00,
+                        0x21, 0x06,
+                        0xb8.toByte(), 0x04, 0x02,
+                        0x00                    // checksum placeholder
+                    )
+                    msg2[msg2.lastIndex] = checksum(msg2, 0, msg2.lastIndex - 1)
+
+                    if (hasCharacteristic(SVC_T2, CHR_T2_WRITE_SHARED)) {
+                        writeTo(SVC_T2, CHR_T2_WRITE_SHARED, msg1, true)
+                        writeTo(SVC_T2, CHR_T2_WRITE_SHARED, msg2, true)
+                    } else if (hasCharacteristic(SVC_T1, CHR_T1_WRITE_CONFIG)) {
+                        writeTo(SVC_T1, CHR_T1_WRITE_CONFIG, msg1, true)
+                        writeTo(SVC_T1, CHR_T1_WRITE_CONFIG, msg2, true)
+                    }
+            
+                    // ES-30M: send 0x22 stored-data query immediately (no 0xA1 wait)
+                    val queryMsg = byteArrayOf(
+                        0x22, 0x06,
+                        seenProtocolType,
+                        0x00, 0x03,
+                        0x00                    // checksum placeholder
+                    )
+                    queryMsg[queryMsg.lastIndex] = checksum(queryMsg, 0, queryMsg.lastIndex - 1)
+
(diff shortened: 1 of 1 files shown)


mr-wewlad:
I'd like to add my 2 cents to this: recording weight (or other data) on my Renpho ES-CS20M is currently not working at all in 3.1.0.

After capturing the logs and walking through the code myself I think the two `0xA0` responses are indeed causing issues (I cannot verify any of the other identified issues in this PR). They were introduced by https://github.com/oliexdev/openScale/pull/1301 to support the ES-30M. That PR also claims that it's backwards compatible, and that it tries to detect the ES-30M by checking byte[4] and the weightScaleFactor, however that detection logic is not used for sending back two responses. In other words, the two responses are always sent, regardless of the scale model.

This PR removes those two responses again, which probably fixes the issue for the ES-CS20M, but probably breaks it again for the ES-30M.

Perhaps a conditional needs to be introduced in the logic for opcode `0x21` to send two responses in the case of a ES-30M, but only one in the case of a ES-CS20M?

Mind you: I haven't actually tested any of this, only got my scale yesterday. This analysis is just from me looking at the code, the debug logs and PRs/issues on Github, trying to figure out why my scale doesn't work. :)

Edit:

Ping @8bitgentleman, since this PR is (probably) about to break the support you builtin, I figured it'd be worth looping you in on this one.

8bitgentleman:
Hi, author of PR #1301 here, thanks for the ping @mr-wewlad  . I agree, I think this PR will break the ES-30M support.
PR #1301 added the dual 0xA0 response in the 0x21 handler specifically because the ES-30M requires it. This PR removes that second frame unconditionally, which would regress ES-30M behavior.
Like said above the underlying issue send to be that both our PRs modify the same 0x21 handler without device-specific branching. Adding some conditional logic here might be needed. Send two 0xA0 frames for ES-30M and one for ES-CS20M. The `isES30MFormat` detection `(byte4Value <= 0x02 && weightScaleFactor == 10.0f)` already exists in the 0x10 handler from #1301 and might be able to be reused here.

Ebonhawk3829:
Thanks @mr-wewlad and @8bitgentleman — you're both right, and I appreciate the catch. My PR removes the dual 0xA0 unconditionally, which would indeed break the ES-30M.

The underlying issue is that #1301 and this PR both modify the 0x21 handler without device-specific branching. As @8bitgentleman suggested, I think the cleanest solution is to reuse the isES30MFormat detection that #1301 already introduced in the 0x10 handler.

The tricky part is that isES30MFormat currently relies on weightScaleFactor, which gets set in the 0x12 handler — and 0x12 may or may not have arrived before 0x21 depending on the scale. However, we can also key off byte[4] of the 0x21 frame itself, or more simply, we can check weightScaleFactor with a safe default: if 0x12 hasn't arrived yet, weightScaleFactor will still be at its initial value (likely 1.0f or whatever the default is), which is != 10.0f, so the detection should still discriminate correctly in practice. If you both concur I could update the PR to do something like this in the 0x21 handler:

```0x21 -> {
    // ES-30M needs two 0xA0 frames (sub-opcodes 0x04 then 0x02)
    // ES-CS20M (and original QN scales) need only one 0xA0 frame (sub-opcode 0x02)
    val isES30M = /* reuse detection logic — e.g. weightScaleFactor == 10.0f 
                     or inspect the 0x21 payload itself */
    
    if (isES30M) {
        // Send dual 0xA0: sub-opcode 0x04 then sub-opcode 0x02 (per #1301)
        // Then send 0x22 immediately (ES-30M doesn't require waiting for 0xA1)
    } else {
        // Send single 0xA0: sub-opcode 0x02 only
        // Do NOT send 0x22 here — wait for 0xA1 ACK first (per working ES-CS20M capture)
    }
}
```

The 0xA1 handler would then only send the 0x22 query for the non-ES-30M path (since the ES-30M path already sent it inline). Do we know whether the ES-30M actually receives an 0xA1 ACK after the dual 0xA0, or does it go straight to 0x10 frames?

mr-wewlad:
> Thanks @mr-wewlad and @8bitgentleman — you're both right, and I appreciate the catch. My PR removes the dual 0xA0 unconditionally, which would indeed break the ES-30M.
> 
> The underlying issue is that #1301 and this PR both modify the 0x21 handler without device-specific branching. As @8bitgentleman suggested, I think the cleanest solution is to reuse the isES30MFormat detection that #1301 already introduced in the 0x10 handler.
> 
> The tricky part is that isES30MFormat currently relies on weightScaleFactor, which gets set in the 0x12 handler — and 0x12 may or may not have arrived before 0x21 depending on the scale. However, we can also key off byte[4] of the 0x21 frame itself, or more simply, we can check weightScaleFactor with a safe default: if 0x12 hasn't arrived yet, weightScaleFactor will still be at its initial value (likely 1.0f or whatever the default is), which is != 10.0f, so the detection should still discriminate correctly in practice. If you both concur I could update the PR to do something like this in the 0x21 handler:
> 
> ```
>     // ES-30M needs two 0xA0 frames (sub-opcodes 0x04 then 0x02)
>     // ES-CS20M (and original QN scales) need only one 0xA0 frame (sub-opcode 0x02)
>     val isES30M = /* reuse detection logic — e.g. weightScaleFactor == 10.0f 
>                      or inspect the 0x21 payload itself */
>     
>     if (isES30M) {
>         // Send dual 0xA0: sub-opcode 0x04 then sub-opcode 0x02 (per #1301)
>         // Then send 0x22 immediately (ES-30M doesn't require waiting for 0xA1)
>     } else {
>         // Send single 0xA0: sub-opcode 0x02 only
>         // Do NOT send 0x22 here — wait for 0xA1 ACK first (per working ES-CS20M capture)
>     }
> }
> ```
> 
> The 0xA1 handler would then only send the 0x22 query for the non-ES-30M path (since the ES-30M path already sent it inline). Do we know whether the ES-30M actually receives an 0xA1 ACK after the dual 0xA0, or does it go straight to 0x10 frames?

It sounds like a workaround that would work! I can imagine that the ES-30M would maybe need some retry logic if the first attempt fails, but we can try it out and see if it works. :)

Ebonhawk3829:
Updated the PR with device-specific branching in the 0x21 handler. It now uses weightScaleFactor == 10.0f (the same heuristic from #1301's 0x10 parser) to detect the ES-30M and preserve the dual 0xA0 + immediate 0x22 sequence for it, while keeping the single 0xA0 + wait-for-0xA1 path for the ES-CS20M and other QN scales.

oliexdev:
Thanks for your effort :+1: 
The whole QN and ESCS-20M scales and derivatives seems a bit complex, see https://github.com/oliexdev/openScale/issues/1308

Anyways if this PR resolve your problems with the ES-CS20M and ES-CS30M I am happy to merge it. Could you please give me a ping this PR works with your scales or not? Thanks in advance.

mr-wewlad:
@Ebonhawk3829 's branch unfortunately doesn't work for me with an ES-CS20M. The `weightScaleFactor` that gets set by the device by opcode 0x12 is also 10.0, so the ES-CS30M path gets triggered:

```
? 2026-04-22 20:27:35.548 D/QNHandler: QN: set weightScaleFactor=10.0 from opcode 0x12
```

Ebonhawk3829:
Hi all — I'm going to close this PR as it stands.

Some context: I opened this purely as a "works on my device" fix for the ES-CS20M after debugging its BLE handshake from snoop logs. I figured I'd contribute it upstream in case it helped others, but I wasn't aware of the full landscape of QN-protocol scales and the differences between them.

The core issue with Commit 2's branching approach is that the weightScaleFactor == 10.0f heuristic doesn't actually distinguish the ES-CS20M from the ES-30M — both scales report the same value from their 0x12 frame, so the ES-CS20M always falls into the ES-30M path and breaks. I've reached the limit of both my available time and my knowledge of these devices, so I don't think I'm the right person to solve the compatibility problem properly.

For anyone who wants to pick this up, the first commit in this PR (the four-fix ES-CS20M change) does work in isolation. The challenge is finding a reliable runtime discriminator between the ES-30M and ES-CS20M. Some avenues that might help:

- Compare seenProtocolType (byte[2] of vendor packets) between the two scales — the ES-CS20M reports 0xFF, but the ES-30M's value is unknown to me:
- Compare the full contents of the 0x12 scale-info frame beyond byte[10]
- Compare the 0x21 frame payload itself — the scales may send different data in their handshake initiation
- Check if the advertised BLE device name differs in a parseable way
- As a last resort, expose a user-facing device variant selector in the app settings

A BLE snoop log from an ES-30M compared against the ES-CS20M capture would probably settle it quickly.

I'll keep a clean branch with just the working ES-CS20M fix on my fork for anyone who needs it.
https://github.com/Ebonhawk3829/openScale/tree/es-cs20m-fix

PR #1355, opened 2026-04-19
Add support for Eufy C20

cvoinf:
Addresses issue [https://github.com/oliexdev/openScale/issues/1352](https://github.com/oliexdev/openScale/issues/1352)

PR #1356, opened 2026-04-22
Add support for Eufy C20

cvoinf:
Continuation of [https://github.com/oliexdev/openScale/pull/1355](https://github.com/oliexdev/openScale/pull/1355)


cvoinf:
It works now quite well. Only one problem persists: If I update s.dateTime while keeping scanning, a new measurement is generated, what I want to avoid. But if it is not updated, an updated weight is not shown in the app. When weight, heartrate and impedance have been all sent by the scale, I update dateTime, and all information is transmitted.
Is there another possibilty to publish a measurement preliminarily (Live-Scan)?

oliexdev:

PR #1359, opened 2026-05-01
Fix QN stored measurements for RENPHO scale

PerkinsWork:
## Summary
- Parse QN `0x23` stored measurement frames used by RENPHO/QingNiu-style scales such as `Renpho-Scale` / ESCS20MA2.
- Retry the stored-data query when the scale initially returns empty stored slots.
- Reject stale stored records from before the current connection session so old scale memory is not imported as a fresh weigh-in.
- Reuse the existing QN body composition calculation path for both live and stored measurements.

## Test Plan
- Built locally with `cd android_app && .\gradlew.bat assembleDebug`.
- Installed debug APK on a Samsung Android phone via ADB.
- Verified successful fresh measurements from a RENPHO ES CS20M / hardware ID ESCS20MA2 scale advertising as `Renpho-Scale`.
- Confirmed empty `0x23` records are rejected and retried until a fresh stored record is received.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/QNHandler.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/QNHandler.kt
@@ -17,6 +17,8 @@
  */
 package com.health.openscale.core.bluetooth.scales
 
+import android.os.Handler
+import android.os.Looper
 import com.health.openscale.R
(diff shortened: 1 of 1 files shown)


oliexdev:
Thanks for your PR, hopefully it doesn't destroy another QN implementation, this device is a little bit tricky.

PR #1361, opened 2026-05-06
Data export to influxdb

h4rib0:
This PR adds the ability to export weight and body metrics data to an InfluxDB instance.

## Changes
- Added `InfluxDbExportUseCases.kt` with the core export logic
- Added InfluxDB connection settings to `SettingsFacade.kt` and `SettingsViewModel.kt`
- Added export option to the Data Management settings screen (`DataManagementSettingsScreen.kt`)
- Added webhook support for triggering exports
- Added localization strings for the new UI elements (de, el, es, eu, fi, fr, it, and more)
- Updated README with InfluxDB export documentation

## Supported versions
Both InfluxDB v1 and v2 (including InfluxDB Cloud) are supported.

## How it works
Users can configure an InfluxDB connection (URL, token, bucket, org) in the app settings and export their measurements directly to InfluxDB.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,8 @@ fastlane/screenshots
 fastlane/test_output
 fastlane/readme.md
 fastlane/README.md
+
+# VS Code local config (machine-specific, not for repo)
+.vscode/launch.json
+.vscode/tasks.json
+.vscode/noop.js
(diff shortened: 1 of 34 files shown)


oliexdev:
Thanks for your PR but for any third party synchronization please implement that in [openScale sync](https://github.com/oliexdev/openScale-sync)

Issue #1371, opened 2026-05-16
perf(TableScreen): reduce recompositions via derivedStateOf, stable LazyColumn keys, and primitive state specialization


SamuelRT123:
# Summary
This issue tracks three micro-optimizations identified in `TableScreen.kt` that 
reduce unnecessary recompositions, GC pressure, and GPU overdraw during normal 
table interaction (scrolling, selection mode, record editing).

These were surfaced through Android Studio's Layout Inspector (recomposition 
counts), Perfetto GPU rendering traces, and Android Lint warnings.

## Problem 1 — Unstable `LazyColumn` key forces full item recomposition on updates

**File:** `TableScreen.kt`  
**Location:** `LazyColumn { items(...) }` block

The current item key is composed as:

```kotlin
items(tableData, key = { "${it.measurementId}_${it.timestamp}" }) { ... }
```

`measurementId` is already globally unique per row. Appending `_${timestamp}` 
makes the key redundant and causes Compose's diffing algorithm to treat 
re-ordered or refreshed items as entirely new nodes rather than moved ones, 
triggering full recomposition of every visible item after any table update 
(e.g. returning from the edit screen).

**Expected:** Only items whose data actually changed should recompose.

## Problem 2 — `resolvedSelectionCount` recalculates on every recomposition

**File:** `TableScreen.kt`  
**Location:** `resolvedSelectionCount` and `resolvedCount` inside 
`showDeleteConfirmDialog`

Both use `remember(selectedKeys, tableDataSnapshot) { ... }`, which re-executes 
the lambda on every recomposition of the surrounding scope — not only when 
`selectedKeys` or `aggItemByPeriodStart` change. In selection mode with many 
rows, this `sumOf` runs on every frame touching any state in the composable.

**Expected:** The count should only recalculate when the inputs it actually 
reads change.

---

## Problem 3 — `mutableStateOf<Int?>` autoboxes on every highlight change

**File:** `TableScreen.kt`  
**Location:** `highlightedItemId` state declaration

```kotlin
var highlightedItemId by remember { mutableStateOf(null) }
```

Android Lint flags this pattern: *"State\<T\> will autobox values assigned to 
this state. Use a specialized state type instead."*  
Every assignment to `highlightedItemId` allocates a new `Integer` wrapper on 
the heap, adding GC pressure during the post-edit scroll/highlight animation 
where timing is frame-critical.

**Expected:** Use `mutableIntStateOf` with a sentinel value to eliminate the 
[...]

PR #1372, opened 2026-05-16
Issue #1371

SamuelRT123:
Fix for problems reported in the issue #1371. Additionally, how Androir works with proccess and taks in the applications it's better to use SingleTask for the launch mode in AndroidManifest file.

### DIFF
--- a/android_app/app/src/main/AndroidManifest.xml
+++ b/android_app/app/src/main/AndroidManifest.xml
@@ -25,6 +25,7 @@
     <application
         android:name=".OpenScaleApp"
         android:allowBackup="true"
+        android:launchMode="singleTask"
         android:dataExtractionRules="@xml/data_extraction_rules"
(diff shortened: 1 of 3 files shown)


oliexdev:
Thanks for the PR.

After review, I’m closing this PR.

The `singleTask` launch mode change introduces unintended navigation and lifecycle side effects and is not appropriate as a performance optimization.

The LazyColumn key change is not justified, as the current key is already stable and unique. Changing it does not provide a measurable benefit.

The `derivedStateOf` change adds complexity without clear benefit, since the existing `remember` usage already handles recomposition efficiently.

The `mutableIntStateOf` and `stringResource` changes are valid, but not sufficient to justify the broader changes in this PR.

Overall, the changes mix minor optimizations with architectural and behavioral modifications that are out of scope for this PR.

Issue #1280, opened 2026-01-10
Since Update to Version 3.0.x the values are incorrect for Huawei CH100


dbielz:
### Preflight Check (Required)

- [x] I tested the latest dev version https://github.com/oliexdev/openScale/releases/tag/dev-build
- [x] The issue still occurs in the dev version
- [x] I searched the existing open and closed issues
- [x] This issue has not already been reported or fixed

### Bug Description

Since the already existing 3 Issues are closed without resolving and I am not aware, how closed issues are handled, I apparently have to create a new one. Long story short: CH100 worked flawlessly with all versions 2.x. Then the first upgrade to 3.x came, 3.0.0, and the values became ridiculously high. The same with version 3.0.1 and now with 3.0.2, although in the already closed issue tickets this issue is claimed as resolved. At least the datetime difference seems now to be resolved, the timestamp now is correct.

So the issue still is: 

Value displayed differ vastly.
in Scale | in App
97.0 kg |  138 kg
28.00 %|  180 % 

### Steps to Reproduce

1. Install any 3.x version of Openscale
2. add the Scale CH100
3. step on the scale

Unfortunately version 2.5.4 isn't available on F-Droid anymore, so I can't create a debug log of a working version.

### Debug Log (Required)

```shell
| Field | Value |
|---|---|
| Time | 2026-01-10 08:03:39.640 |
| Timezone | Europe/Berlin (UTC 1h) |
| Session ID | 19ba6b7c378 |
| App | openScale |
| Version | 3.0.2 (72) |
| Package | com.health.openscale |
| Build Type | release |
| Device | samsung SM-S921B |
| Android | 16 (API 36) |
| Build ID | BP2A.250605.031.A3.S921BXXSBCYK8 |
| Build Time | 2026-01-09T00:56:12Z |
| Git SHA | 7a76b06f |


? 2026-01-10 08:03:39.639 D/openScaleLog: Initial markdown-diff headers written to: /storage/emulated/0/Android/data/com.health.openscale/files/logs/openScale_current_log.txt
? 2026-01-10 08:03:39.641 D/openScaleLog: Initial markdown-diff headers written to: /storage/emulated/0/Android/data/com.health.openscale/files/logs/openScale_current_log.txt
+ 2026-01-10 08:03:39.641 I/openScaleLog: File logging enabled during runtime – started fresh session log.
[...]

Issue #1276, opened 2026-01-07
Fixed Huawei AH100/CH100 measurement parsing


jphansen:
### Preflight Check (Required)

- [x] I tested the latest dev version https://github.com/oliexdev/openScale/releases/tag/dev-build
- [x] The issue still occurs in the dev version
- [x] I searched the existing open and closed issues
- [x] This issue has not already been reported or fixed

### Bug Description

As other has reported, the Huawei CH100 scale, or off measurements on both weight and body-fat. 

This is not mere a bugreport, but also include a workable "patch". I prefer this to a pull request, as my code right now is filled up with a lot of lovely debug information :-)

Anyway, what I have done is revert the incorrect measurement parsing changes from commit 2e7e708f that
caused wrong weight and body fat calculations.

Issues fixed:
- Weight was using incorrect formula: (1457 - byte[1]) / 10
- Body fat was reading from wrong position (20) as whole percent
- Data was being treated as XOR-only instead of AES-CTR encrypted
- CH100 device name was not recognized (CH100 is same hardware as AH100)

The December 2025 commit was based on incorrect reverse engineering, I assume. My scale data is AES-CTR encrypted (not just XOR obfuscated) and uses the original v2.5.4 format.

Restored correct parsing:
- User ID: position 0
- Weight: positions 1-2 (little-endian 16-bit, divided by 10)
- Body fat: positions 3-4 (little-endian 16-bit, divided by 10)
- Timestamp: positions 5-12 (year, month, day, hour, minute, second)
- Resistance: positions 13-14 (for future body composition calculations)

Added CH100 device name support since CH100 and AH100 are the same
hardware with different advertised names.

I will put the patch in the "how-to-recreated" session.. /JPH

### Steps to Reproduce

```
---
 .../bluetooth/scales/HuaweiAH100Handler.kt    | 89 +++++++++++--------
 1 file changed, 52 insertions(+), 37 deletions(-)

diff --git a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/HuaweiAH100Handler.kt b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/HuaweiAH100Handler.kt
index f14abe4d..8c9e3f12 100644
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/HuaweiAH100Handler.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/HuaweiAH100Handler.kt
[...]

PR #1374, opened 2026-05-19
Fix Huawei AH100 / CH100 measurement parsing (closes #1206, #1280)

### LINKED ISSUE #1206: Huawei CH100, New Framework
Hi there, thanks a lot for this very useful lightweight app, I use it daily now for over 1 1/2 years for tracking my weight with a Huawei CH100 scale. 

Up to now it worked perfectly, without any issue, in the last version 2.5.4. Now on F-Droid an update to version 3.0 was offered, which I installed instantly. Here's my feedback:
- the new version is _**8 times bigger**_ than the last one ( **53MB** vs **7MB** )
- I found it harder to (re-)add my scale than before, it took me some time to find the right place in the apps settings
- the Bluetooth behavior seems to have changed: before apparently, when starting the app the BT-connection was initiated, when I stepped on my scale, the connection was established and when the scale had finished collecting data, the connection was terminated. Now, the connection seems to be established all the time, whereas I don't know, how the CH100 handles Bluetooth in standby mode. I can't find this very intuitive.
- Now for the most predominant issue, my motivation for this feedback: the data read from the **CH100** is now **completely wrong**: the date read is something like **April 3084** (or similar, can't recall the exact value and have downgraded again), the weight is something about **295Kg** (instead of 95) and the body fat measurement now takes ages to finish and comes out on something like **300%** .

Ok, there seems to be some interpretation factor wrong for the CH100, which could probably be fixed quite easy. But as for the rewrite to a new platform: I'm sure it was quite an effort to do this, but is there any benefit from it? As I never had the feeling, that the old version was slow or sluggish in any way, I can't see any performance improvement. So, sorry for having to say, that for me all that remains, is a bigger size and a more complex interface (and some new problems with data acquirement at least for my CH100).

By the way: for the downgrade I had to uninstall the new version, which cost me my complete data from the last 18 months. But this was my own fault, since I didn't verify if the backup I had configured already had saved data. I somehow had the expectation that configuring the backup already initiates the first backup. Maybe this could be improved to prevent other users from making the same experience.

danieldorigatti:
# Fix Huawei AH100 / CH100 measurement parsing

Closes #1206 and #1280. Supersedes #1276.

## Summary

- Restore the v2.5.4 measurement decode for the Huawei AH100 / CH100 scales (Chipsea CST34M97, BLE service `0xFAA0`).
- Split the handler into `HuaweiAH100Handler` (advert `AH100`) and `HuaweiCH100Handler` (advert `CH100`). Until now `HuaweiAH100Handler` matched `CH100` only, so real AH100 hardware was never picked up.
- Move all byte-level primitives into a pure-Kotlin object so they can be unit-tested on the JVM.

## The bug

`HuaweiAH100Handler` in 3.x stopped AES-CTR-decrypting the encrypted measurement frame, used invented field offsets, and computed weight as `(1457 - byte) / 10`. Reports from #1206 and #1280:

| scale display | app shows  |
| ---           | ---        |
| 95.0 kg       | 138.6 kg   |
| 28 % fat      | 180 % fat  |
| today         | year 3084  |

First fixture from the new `HuaweiAhCh100ProtocolTest` (matches the capture in #1280):

```
Notification on 0xFAA2:
  bc100edef702c9bbf2176c755b6eccdd5b92b0

3.x master parser:
  weight   = (1457 - 0xF7) / 10        = 122.6 kg     (wrong)
  fat      = byte[20] (out of bounds)  = 0 / crash    (wrong)
  datetime = "now", no parse                          (lost)

This patch (same numbers v2.5.4 produced):
  aesCtr(deobfuscate(payload), magicKey, INITIAL_IV)
    = 01 ca 03 18 01 ea 07 05 13 08 0c 21 02 1c 02
  weight    = u16le[1..2] / 10 = 0x03ca / 10 = 97.0 kg
  fat       = u16le[3..4] / 10 = 0x0118 / 10 = 28.0 %
  datetime  = 2026-05-19 08:12:33 (Tuesday)
  user      = 1
  impedance = 540 ohm
```

Reconstructed from the v2.5.4 source and verified against captures from #1280.

- Service `0xFAA0` with TX `0xFAA1` (write) and RX `0xFAA2` (notify).
- Frames are MAC-XOR obfuscated; encrypted frames additionally use AES-128-CTR with `magicKey` derived from the auth token.
- The measurement is split across two notifications (`op=0x0E` then `op=0x8E`). AES-CTR-decrypt the deobfuscated payload of the first half only; the second half carries derived BIA fields that v2.5.4 did not decode.
- Decrypted measurement layout: `[user, weightLE, fatLE, yearLE, month, day, hour, min, sec, dow, impedanceLE]`.
- Plain `0xDB` header: `lengthByte = payload.size + 1`. Encrypted `0xDC` header: `lengthByte = payload.size`. The 3.x rewrite had the `0xDB` length off by one.
- `USER_INFO` payload (CMD `0x09`) is the v2.5.4 14-byte layout: `auth(7) || age | sexBit | height | 0x00 | weightLE | resistanceLE | 0x1C 0xE2`.

## Changes

- **New `core/bluetooth/libs/HuaweiAhCh100Protocol.kt`** - pure-Kotlin object holding `obfuscate`, `aesCtr`, `xorChecksum`, `buildAuthToken`, `deriveMagicKey`, `buildPlainCommand`, `buildEncryptedCommand`, and `parseMeasurement`. `INITIAL_KEY` / `INITIAL_IV` are exposed via getters returning defensive copies.
- **New abstract `HuaweiAhCh100ScaleHandler`** with the BLE state machine.
- `HuaweiAH100Handler` rewritten as a thin subclass matching advert `AH100`.
- **New `HuaweiCH100Handler`** matching advert `CH100`. Both products use the same Chipsea CST34M97 hardware (confirmed in #1276); splitting just lets AH100 hardware be detected and leaves room for firmware-specific divergence later.
- Name matching strips trailing NULs, whitespace, and `HUAWEI ` / `HONOR ` prefixes so OEM rebrands still resolve to the right handler.

## Notes

If the first half already contains at least 15 plaintext bytes (i.e. the scale shipped the full record in one notification under the negotiated 185-byte MTU), the handler decodes immediately and ignores any later `0x8E` half. This means the default Balanced tuning works as well as the legacy 23-byte ATT MTU; no tuning profile needs to be pinned.


android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/HuaweiAH100Handler.kt
```diff
@@ -17,465 +17,14 @@
...
-            if (i > off) sb.append(' ')
-            sb.append(String.format("%02X", b[i]))
-        }
-        return sb.toString()
-    }
-
-    private fun ts(d: Date) = SimpleDateFormat("yyyy-MM-dd HH:mm:ss", Locale.US).format(d)
+class HuaweiAH100Handler : HuaweiAhCh100ScaleHandler() {
+    override val supportedAdvertName: String = "AH100"
+    override val displayName: String = "Huawei AH100"
```
oliexdev: No need for that class anymore should be handled in HuaweiAhCh100Handler

android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/HuaweiAhCh100ScaleHandler.kt
```diff
@@ -0,0 +1,466 @@
...
+ *    notification at that MTU instead of the paired 16-byte halves
+ *    v2.5.4 saw at MTU 23. [tryDecodeFirstHalfImmediately] handles both
+ *    shapes, so we don't need to pin the conservative profile.
+ *  - Real-world note: when migrating from openScale 2.5.x or Huawei Health
+ *    on the same phone, users may need to "Forget" the scale once in the
+ *    Android Bluetooth settings. The link-layer encryption keys from the
+ *    previous bond don't survive this protocol's re-pair flow and otherwise
+ *    cause `CONNECTION_TERMINATED_MIC_FAILURE` after the first command.
+ */
+abstract class HuaweiAhCh100ScaleHandler : ScaleDeviceHandler() {
```
oliexdev: Please rename the class name for Consistency to HuaweiAhCh100Handler

android_app/app/src/main/java/com/health/openscale/core/bluetooth/libs/HuaweiAhCh100Protocol.kt
```diff
@@ -0,0 +1,353 @@
...
+ * - **Decrypt only the first half's deobfuscated tail** (this is what
+ *   v2.5.4 does and what the captures confirm). The keystream of AES-CTR
+ *   is contiguous so decrypting the merged buffer would give the same
+ *   bytes for the first 16, but the second half then contains junk; we
+ *   keep things simple and decrypt just what we need.
+ * - The decrypted layout is documented on [parseMeasurement].
+ *
+ * @see HuaweiAhCh100ProtocolTest
+ */
+object HuaweiAhCh100Protocol {
```
oliexdev: Please remove this protocol class from the libs and integrate them into the HuaweiAhCh100Handler

android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/HuaweiCH100Handler.kt
```diff
@@ -0,0 +1,35 @@
...
+ * Same Chipsea CST34M97-based hardware as the [HuaweiAH100Handler]; the two
+ * products only differ in the BLE advertisement name. All wire-protocol
+ * logic lives in [HuaweiAhCh100ScaleHandler].
+ *
+ * This handler restores the v2.5.4 measurement-decoding behaviour after the
+ * 3.x rewrite regressed it (see issues #1206, #1276, #1280). Tests in
+ * `HuaweiAhCh100ProtocolTest` lock the protocol so the regression cannot
+ * recur silently.
+ */
+class HuaweiCH100Handler : HuaweiAhCh100ScaleHandler() {
```
oliexdev: Should be better handled in HuaweiAhCh100Handler

android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/HuaweiAH100Handler.kt
danieldorigatti: HuaweiAH100Handler and HuaweiCH100Handler are removed; the single HuaweiAhCh100Handler now matches both the AH100 and CH100 advertisement names.

android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/HuaweiAhCh100ScaleHandler.kt
danieldorigatti: HuaweiAhCh100ScaleHandler is renamed to HuaweiAhCh100Handler and is now a concrete class.

android_app/app/src/main/java/com/health/openscale/core/bluetooth/libs/HuaweiAhCh100Protocol.kt
danieldorigatti: HuaweiAhCh100Protocol is removed from libs/; its constants and wire-protocol functions are now an internal companion object inside HuaweiAhCh100Handler. It was in libs/ so it could be JVM-unit-tested without a real scale; HuaweiAhCh100ProtocolTest is retargeted as HuaweiAhCh100HandlerTest against the companion object, so the #1206/#1280 regression stays locked.

android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/HuaweiCH100Handler.kt
danieldorigatti: HuaweiCH100Handler is removed; HuaweiAhCh100Handler handles the CH100 advertisement name.

oliexdev:
Thanks for your PR :+1:

PR #1377, opened 2026-05-23
Fix EEPROM overflow, char_to_int validation, dead debug function and simplify loop/decode_seg

BhaveshDave09:
- write_scale_data(): add EEPROM bounds check to prevent silent corruption when storage is full
- char_to_int(): validate input range so garbled display chars don't produce garbage measurements
- print_debug_output(): was defined but never called, wired to Serial command '4' for hardware debugging
- loop(): simplify active-channel detection from 25-line if/else chain to 7-line loop
- decode_seg(): replace 80-line if-chain with compact lookup table

### DIFF
--- a/arduino_mcu/openScale_MCU/openScale_MCU.ino
+++ b/arduino_mcu/openScale_MCU/openScale_MCU.ino
@@ -40,6 +40,14 @@
 #define MAX_SAMPLE_SIZE 6
 #define MAX_NO_ACTIVITY_CYCLES 32
 
+// FIX 1: EEPROM bounds constants.
+// The I2C_eeprom at address 0x50 is typically a 32KB (32768-byte) chip.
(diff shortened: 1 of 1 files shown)


oliexdev:
Thanks for your PR but did you test that on a real hardware?

oliexdev:
Thanks for the PR, but I can't merge this.

 The decode_seg() lookup table is incorrect — several masks don't match the original segment patterns, '8' and '9' map to the same value, and there are unfinished "wait, let me re-derive" comments left in the committed code. Since this function decodes every measurement, the firmware wouldn't work. The loop() rewrite also changes behavior for multi-LOW states that were previously ignored.

These changes clearly haven't been verified on real hardware. Please test everything on an actual device first and only open a PR once it's confirmed working. Thanks!

PR #1378, opened 2026-05-24
security: restrict READ_WRITE_DATA permission to signature level

jaisukh:
## Summary
- Changes `android:protectionLevel` on the `READ_WRITE_DATA` custom permission from `dangerous` to `signature`
- Prevents any installed third-party app from requesting and receiving this permission at install time
- Access is now limited to apps signed with the same certificate (official sync companions)

## Security Impact
The exported `DatabaseProvider` ContentProvider was previously accessible to **any** installed app that declared the permission in its manifest — no user interaction required. With `signature`, only apps sharing the openScale signing certificate can access health data via the provider.

## Test plan
- [ ] Verify official sync companion app still connects to the ContentProvider normally
- [ ] Verify a test app with a different signing key is denied access to the provider

### DIFF
--- a/android_app/app/src/main/AndroidManifest.xml
+++ b/android_app/app/src/main/AndroidManifest.xml
@@ -12,7 +12,7 @@
         android:name="${applicationId}.READ_WRITE_DATA"
         android:description="@string/permission_read_write_data_description"
         android:label="@string/permission_read_write_data_label"
-        android:protectionLevel="dangerous" />
+        android:protectionLevel="signature" />
 
     <uses-feature android:name="android.hardware.bluetooth_le" android:required="false"/>
 


oliexdev:
I'm declining this one. With minSdk 31, the dangerous permission isn't granted at install — a third-party app must request it at runtime and the user must explicitly approve it via a prompt whose text we control. The permission has no permissionGroup, so the group-based silent auto-grant doesn't apply either. Access is already gated by explicit user consent.
More importantly, signature would restrict access to apps signed with our certificate, which breaks openScale's multi-channel distribution (Play, F-Droid) and locks out the OSS sync companion and any third-party integration — the exact use case the exported provider exists for.

PR #1381, opened 2026-05-25
feat: add support for Dr.Trust Smart Scale 532

dharmendra-gupta:
## Scale

Dr. Trust Smart Scale SSW532 (ICOMON FG2211WB, 8-electrode BIA)

🔗 https://drtrust.in/products/dr-trust-usa-532-eight-electrode-digital-body-weight-bmi-machine-fat-analyzer-with-ios-android-app-sync-29-essential-metrics-composition-analysis-weighing-machine-for-health-monitoring
🔗 https://www.amazon.in/dp/B0DV3JDC57

---

## What's working

- Live weight streaming with stability detection (only locks value when scale blinks/stabilises)
- Full body composition when handles are held: fat%, water%, muscle%, bone mass, LBM, BMR, impedance, visceral fat
- Weight-only fallback when scale is used without handles
- Step-off detection: publishes weight immediately when user steps off without picking up handles, rather than waiting for BLE timeout
- Visceral fat derived from trunk segment impedance Z3 - scientifically more accurate than anthropometric-only estimates

---

## Protocol summary

Custom GATT service `0xFFB0` with three characteristics:

- `0xFFB1` — command write (App → Scale, 20 bytes)
- `0xFFB2` — live weight NOTIFY: stability flag at byte[4] (0x03 = locked/blink), weight BE uint16÷1000 at bytes[7-8]
- `0xFFB3` — result INDICATE: setup frames (byte[1]=0x18) and measurement frames (byte[1]=0x23)

**Startup sequence (state machine):**
1. Subscribe INDICATE on FFB3 → scale sends setup frame with dynamic session ID
2. Subscribe NOTIFY on FFB2 → scale sends second setup frame
3. Write user profile to FFB1 (3 packets: init, demographics, app-id)

**Measurement frames keyed by byte[2]:**
- `0x00` — weight + whole-body impedance channels A/B
- `0x01` — 8 segmental impedances: Z3 (trunk), Z4 (right leg), Z5 (left leg), plus arm and cross-body channels
- `0x02` — end marker; triggers teardown ACK and publish decision

Whole-body impedance is reconstructed as the foot-to-foot electrical path: **Z3 (trunk) + Z4 (right leg) + Z5 (left leg)**, consistent with how the ICOMON Sensun BLE SDK reports `zuKang` (total impedance) internally.

---

## Why this is better than the official Dr. Trust app

The official Dr. Trust app (`com.nurecausa.drtrustscaleconnect`) was reverse-engineered (APK decompiled via JADX, native `.so` disassembled) and revealed two critical findings:

1. **The AiFit path hardcodes impedance to 551Ω** regardless of what the scale actually measures (`AicareBleConfig.getBodyFatData(..., 551)`), and overrides water% with a purely anthropometric formula that ignores impedance entirely.

2. **The ICOMON/SSW532 path** (used by this scale) does use real impedance from the BLE SDK, but its body composition formulas are locked inside the proprietary `libICBleProtocol.so` native library (`CouStru.CountGetXxx3()`), which cannot be audited or adapted.

This openScale implementation uses raw segmental impedance data directly from the scale's BLE packets and passes it through `StandardImpedanceLib`, giving users full transparency and control over the formula pipeline.

---

## Visceral fat approach

Rather than using an anthropometric-only proxy (weight, height, age), visceral fat is estimated from **trunk impedance Z3** - the segment with the strongest correlation to visceral adipose tissue in BIA literature. Calibrated against a real measurement (Z3 = 42.4 Ω, 54.8 kg male, 82 cm waist):

```
VF% = 2.87 × (weight / Z3) + 0.01 × age  [+0 male / −1.0 female]
```

---

## Files changed

| File | Change |
|---|---|
| `DrTrustSSW532Handler.kt` | New handler — full BLE protocol, state machine, BIA pipeline, visceral fat formula |
| `ScaleFactory.kt` | Registered handler in `modernKotlinHandlers` list |
| `EvaluationReferenceTables.kt` | Fixed visceral fat evaluation band: `low = -1f` → `1f` (was triggering false "no matching age band" UI error for any visceral fat reading) |
| `.gitignore` | Added `kotlin` auto-generated folder |

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,6 @@ fastlane/screenshots
 fastlane/test_output
 fastlane/readme.md
 fastlane/README.md
+
+#kotin
+.kotlin
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/ScaleFactory.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/ScaleFactory.kt
@@ -42,6 +42,7 @@ import com.health.openscale.core.bluetooth.scales.HuaweiCH100SHandler
 import com.health.openscale.core.bluetooth.scales.IHealthHS3Handler
 import com.health.openscale.core.bluetooth.scales.InlifeHandler
 import com.health.openscale.core.bluetooth.scales.LinkMode
+import com.health.openscale.core.bluetooth.scales.DrTrustSSW532Handler
 import com.health.openscale.core.bluetooth.scales.MGBHandler
 import com.health.openscale.core.bluetooth.scales.MedisanaBs44xHandler
 import com.health.openscale.core.bluetooth.scales.MiScaleHandler
@@ -113,6 +114,7 @@ class ScaleFactory @Inject constructor(
         MiScaleS400Handler(),
         MiScaleHandler(),
         RunstarR5Handler(),
+        DrTrustSSW532Handler(),
         MGBHandler(),
         MedisanaBs44xHandler(),
         InlifeHandler(),
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/DrTrustSSW532Handler.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/DrTrustSSW532Handler.kt
@@ -1,0 +1,361 @@
+/*
+ * openScale
+ * Copyright (C) 2025 olie.xdev <olie.xdeveloper@googlemail.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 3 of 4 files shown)


dharmendra-gupta:
I further tested Visceral fat approach with 03 real female user, snooped hci logs and tested against dr. trust app values; formula is working perfectly fine and obviously better.

dharmendra-gupta:
Hey, 

tested with users having weight more than 65kg, and it was showing incorrect weight.

Root cause: The live weight frame (FFB2) and measurement frame (FFB3 pkt0) encode weight as a 3-byte big-endian gram value. The original code read only bytes
  [7-8] / [10-11] as a uint16, which silently overflows for anyone heavier than 65.535 kg.

  Example: a 70 kg user sends 70000 g = 0x011170. Reading only the lower 2 bytes gives 0x1170 = 4464. 4464 ÷ 1000 = 4.46 kg — reported as 4 kg, then disconnects.

  This went undetected because the original HCI captures were from a 54.8 kg user - well within uint16 range, no overflow.

  Fix: Added readBE24() helper and changed both weight reads to start one byte earlier (offsets 6 and 9 respectively). For users under 65.5 kg the high byte is
  0x00 so the result is identical to before.

  Confirmed: Reproduced on multiple phones and user profiles. Fixed by this change.

dharmendra-gupta:
Hey, 
Sorry, there are issues with scale, for some reason dr trust app keeping the connection from different phone and scale is behaving weirdly.
i will open a different PR after finding all the bugs.

PR #1384, opened 2026-05-28
feat: add support for Dr.Trust Smart Scale 532

dharmendra-gupta:
## Scale
  
  Dr. Trust Smart Scale SSW532 (ICOMON FG2211WB, 8-electrode BIA)

  🔗 https://drtrust.in/products/dr-trust-usa-532-eight-electrode-digital-body-weight-bmi-machine-fat-analyzer-with-ios-android-app-sync-29-essential-metrics-com
  position-analysis-weighing-machine-for-health-monitoring
  🔗 https://www.amazon.in/dp/B0DV3JDC57

  ---
  What's working

  - Live weight streaming with stability detection (only locks value when scale stabilises/blinks)
  - Full body composition when handles are held: fat%, water%, muscle%, bone mass, LBM, BMR, impedance, visceral fat
  - Weight-only fallback when scale is used without handles
  - Step-off detection: publishes weight immediately when user steps off without picking up handles
  - Historical measurement replay detection: cached measurements from scale memory are silently discarded, only live measurements are published
  - 5-second BIA wait timer: after weight locks, waits up to 5 seconds for handle pickup before publishing weight-only and disconnecting cleanly
  - Visceral fat derived from trunk segment impedance Z3 — scientifically more accurate than anthropometric-only estimates

  ---
  Protocol summary

  Custom GATT service 0xFFB0 with three characteristics:

  - 0xFFB1 — command write (App → Scale, 20 bytes)
  - 0xFFB2 — live weight NOTIFY: stability flag at byte[4] (0x03 = locked/blink), weight 3-byte BE gram value at bytes[6-8] divided by 1000
  - 0xFFB3 — result INDICATE: setup frames (byte[1]=0x18) and measurement frames (byte[1]=0x23)

  Startup sequence (state machine):
  1. Subscribe INDICATE on FFB3 → scale sends setup frame with dynamic session ID
  2. Subscribe NOTIFY on FFB2 → scale sends second setup frame 
  3. Write user profile to FFB1 (3 packets: init, demographics, app-id)

  Measurement frames keyed by byte[2]:
  - 0x00 — weight + whole-body impedance channels A/B; byte[3] command byte distinguishes live (0xA3/0xA7) from cached historical replays
  - 0x01 — 8 segmental impedances: Z3 (trunk), Z4 (right leg), Z5 (left leg), plus arm and cross-body channels
  - 0x02 — end marker; triggers teardown ACK and publish decision

  Whole-body impedance is reconstructed as the foot-to-foot electrical path: Z3 (trunk) + Z4 (right leg) + Z5 (left leg), consistent with how the ICOMON Sensun
  BLE SDK reports zuKang (total impedance) internally.

  ---
  Bugs found and fixed during testing
  
  1. Incorrect weight for users weighing >65.5 kg
  The live weight frame and result frame both encode weight as a 3-byte (24-bit) big-endian gram value, not 2-byte. Reading only 2 bytes causes an integer overflow above 65,535 grams (65.5 kg), producing incorrect readings. Fixed by adding readBE24() and reading bytes[6-8] on the live frame and bytes[9-11] on the result frame.

  2. Historical measurement replay causing false publishes
  When reconnecting, the scale replays its last stored measurement before sending the live one. The command byte at pkt0[3] distinguishes live (0xA3/0xA7) from cached — cached frames are now detected and silently discarded, preventing stale data from being published.

  3. Premature disconnect on weight-only flow
  Previously, a writeTeardownAck() was sent immediately when weight locked but BIA had not arrived yet, which shut down the scale and prevented handle pickup.
  Fixed: teardown ACK is now withheld while waiting for handles. A 5-second coroutine fallback fires if no BIA arrives, publishing weight-only and disconnecting cleanly.
  
  4. Double-publish / weight-only publish from cached BLE replay on reconnect
  onDisconnected() was unconditionally publishing pendingWeightKg if body comp had not been published, but pendingWeightKg could be populated from a cached historical frame. Fixed by requiring both isLiveWeightLocked (set only when the live frame confirms stability) and savedWeightKg > 0f before any fallbackpublish.
  
  ---
  Why this is better than the official Dr. Trust app
  
  The official Dr. Trust app (com.nurecausa.drtrustscaleconnect) was reverse-engineered (APK decompiled via JADX, native .so disassembled) and revealed two critical findings:
  
  1. The AiFit path hardcodes impedance to 551 Ω regardless of what the scale actually measures (AicareBleConfig.getBodyFatData(..., 551)), and overrides water% with a purely anthropometric formula that ignores impedance entirely.
  2. The ICOMON/SSW532 path does use real impedance from the BLE SDK, but its body composition formulas are locked inside the proprietary libICBleProtocol.so native library (CouStru.CountGetXxx3()), which cannot be audited or adapted.
  
  This openScale implementation reads raw segmental impedance directly from the BLE packets and passes it through StandardImpedanceLib, giving users full transparency and control over the formula pipeline.
  
  ---
  Visceral fat approach
  Rather than using an anthropometric-only proxy (weight, height, age), visceral fat is estimated from trunk impedance Z3 - the segment with the strongest
  correlation to visceral adipose tissue in BIA literature. Calibrated against real measurements:

  ``` VF% = 2.87 × (weight / Z3) + 0.01 × age  [+0 male / −1.0 female] ```

  Validated against HCI snooped logs from 3 real female users and cross-checked against Dr. Trust official app values - formula output is accurate and consistent.

  ---
  Note on offline/historical measurements
  
  The scale stores historical measurements internally and replays them on reconnect. These replays are now detected and discarded. The scale does not transmit date/time metadata with historical entries, so accurate time-stamping of past measurements is not possible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,6 @@ fastlane/screenshots
 fastlane/test_output
 fastlane/readme.md
 fastlane/README.md
+
+#kotin
+.kotlin
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/ScaleFactory.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/ScaleFactory.kt
@@ -61,6 +61,7 @@ import com.health.openscale.core.bluetooth.scales.SenssunHandler
 import com.health.openscale.core.bluetooth.scales.SinocareHandler
 import com.health.openscale.core.bluetooth.scales.SoehnleHandler
 import com.health.openscale.core.bluetooth.scales.SppScaleAdapter
+import com.health.openscale.core.bluetooth.scales.DrTrustSSW532Handler
 import com.health.openscale.core.bluetooth.scales.StandardBeurerSanitasHandler
 import com.health.openscale.core.bluetooth.scales.TrisaBodyAnalyzeHandler
 import com.health.openscale.core.bluetooth.scales.TuningProfile
@@ -134,6 +135,7 @@ class ScaleFactory @Inject constructor(
         BeurerSanitasHandler(),
         AAAxHandler(),
         ActiveEraBF06Handler(),
+        DrTrustSSW532Handler(),
     )
 
     /**
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/DrTrustSSW532Handler.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/DrTrustSSW532Handler.kt
@@ -1,0 +1,421 @@
+/*
+ * openScale
+ * Copyright (C) 2025 olie.xdev <olie.xdeveloper@googlemail.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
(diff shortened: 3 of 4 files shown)


oliexdev:
Thanks for your PR :+1:

PR #1385, opened 2026-05-29
enhancement: chart toggle, scrollable chart, and measurement cards

dharmendra-gupta:
# PR: Overview Chart & Measurement Card UX Improvements

---

## Overview

This PR introduces three user-facing improvements to the overview screen.

---

### 1. Chart show/hide toggle

Added a toggle in **Settings → Appearance** to show or hide the chart on the overview screen.

- Default: **on**
- When disabled, the measurement list takes full screen height
- Setting persisted via DataStore key `show_overview_chart`

---

### 2. Scrollable chart mode

Added a second toggle (only visible when chart is enabled) to switch between two chart modes:

| Mode | Behaviour |
|------|-----------|
| **Fixed** (default) | Chart stays pinned above the list with a draggable splitter to resize |
| **Scrollable** | Chart becomes the first item in the measurement list and scrolls with it |

**Technical note:** when scrollable, the chart is rendered with a fixed height of `260dp` because `LazyColumn` items have unbounded height and `MeasurementChart` uses `weight()` modifiers internally that require a bounded parent.

Setting persisted via DataStore key `overview_chart_scrollable`.

---

### 3. Measurement card collapse

Reworked expand/collapse behaviour on measurement cards:

- **Collapsed state** now hides all parameters — previously pinned values (e.g. weight) always showed regardless of expand state, making cards appear never fully collapsed
- **Latest measurement** (index 0) auto-expands on load — gives an immediate summary of the most recent entry
- **All previous measurements** start collapsed — keeps the list clean and scannable
- Expand/collapse controlled by:
  - Chevron icon in the card header (always visible)
  - Collapse button at the bottom of an expanded card

---

## Files Changed

| File | Change |
|------|--------|
| `SettingsFacade.kt` | Added `overviewChartScrollable` Flow + setter + DataStore key |
| `OverviewScreen.kt` | Chart toggle, scrollable chart item, card collapse fix, copyright |
| `GeneralSettingsScreen.kt` | Two new toggles under Appearance, copyright |
| `strings.xml` | Added `settings_chart_scrollable` string |

---

## Settings Keys Added

| Key | Type | Default | Description |
|-----|------|---------|-------------|
| `show_overview_chart` | Boolean | `true` | Show/hide chart on overview |
| `overview_chart_scrollable` | Boolean | `false` | Fixed vs scrollable chart mode |

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/facade/SettingsFacade.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/facade/SettingsFacade.kt
@@ -77,6 +77,8 @@ object SettingsPreferenceKeys {
     // Settings for specific UI components
     val SELECTED_TYPES_TABLE = stringSetPreferencesKey("selected_types_table") // IDs of measurement types selected for the data table
     val MY_GOALS_EXPANDED_OVERVIEW = booleanPreferencesKey("my_goals_expanded")
+    val SHOW_OVERVIEW_CHART = booleanPreferencesKey("show_overview_chart")
+    val OVERVIEW_CHART_SCROLLABLE = booleanPreferencesKey("overview_chart_scrollable")
 
     // Saved Bluetooth Scale
     val SAVED_BLUETOOTH_DEVICE_ADDRESS        = stringPreferencesKey("saved_bluetooth_device_address")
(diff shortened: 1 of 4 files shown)


oliexdev:
Thanks for the PR! 

 The chart show/hide toggle is redundant: you can already hide the chart by dragging the splitter all the way up — that position is persisted, so the list gets the full height on the next launch. No extra setting needed.

 The scrollable mode I don't want either. By your own note it requires a hardcoded 260dp height, which throws away the resizable splitter and fixes the chart size. Worse, the chart scrolls out of view as soon as you browse the list — so you lose exactly what the chart is for: comparing it against the entries while scrolling. That's a worse UX behind an extra toggle.

The card collapse change isn't needed either: which values stay visible on a collapsed card is already controlled by pinning/unpinning measurement types in Settings → Measurement types. Pinned values are shown by design, so users can already decide what a collapsed card displays. For an overview you can use the table view or aggregate the items.

In general I'd rather avoid piling on settings for things the splitter and the existing pin mechanism already solve.

PR #1392, opened 2026-06-08
feat: Add support for 1BODY CONNECT  (Transtek family)

LucasDumont:
Trying to add [Terraillon BODY CONNECT](https://terraillonhelp.zendesk.com/hc/en-us/articles/9709558797596-BODY-CONNECT-Instruction-Manual) to supported scale.

Tested for a few days, pairing and data receiving seems OK. Except when stepping out of the scale before body measurement are done, but i don't really know how to handle it correctly.

I added `USER_SYNC` in capability (but not implemented) as the scale should handle multiples user. But again, i don't really know how to handle it correctly.

I also added `HISTORY_READ` but i don't know if i should have done it : it's kind of a one shot when connecting to the scale. And then the scale "lose" it history, once send. Don't know if it match the real "history read" capacity :/

Never really used Kotlin or Java, so i hope the code isn't too bad


oliexdev:
Thanks for your PR

PR #1393, opened 2026-06-09
feat: add support for Taylor 5331891 

chropic:
# Add support for Taylor 5331891 BIA Scale

## Scale
**Taylor Body Composition Bathroom Scale (model 5331891)** — shows up on Bluetooth as `5331891 BIA Scale`.

## Summary
This PR adds a dedicated `TaylorBIAHandler`. Because this scale uses the same Bluetooth service (`0xFFB0`) as the MGB scale family, our app used to accidentally let the `MGBHandler` claim it. That caused the connection to time out without getting a measurement, as Taylor's data format is entirely different.

**What works:** Live and stable **weight** (tested against real measurements).
**Not implemented:** Body composition (explained below).

## Protocol (reverse-engineered)
- Communicates over service `0xFFB0`.
- Requires a quick initial handshake to tell the scale to start streaming data.
- Data arrives in 20-byte frames. The weight is extracted using this formula: 
  `weight_kg = (((chan & 0x01) << 16) | (hi << 8) | lo) / 1000.0`
- Verified decodes: `80 8D ...` → 68.200 kg = 150.4 lb; `80 8D ...` → 69.500 kg = 153.2 lb (which perfectly matches the official app's readout).

## Publishing the final reading
Most of the time, the scale doesn't send a specific "done" signal, it just continuously streams live weight values until the user stands still and the numbers plateau. To handle this, I added a **stability check**: once the weight stays exactly the same for 4 readings in a row (about 1 second), it's locked and published. The handler will still honor explicit "stable" signals if they ever arrive, and there's a fallback timer just in case the weight never fully settles.

## No body composition
The scale only ever sends weight, there is **no body impedance data sent over BLE**. The official Precision Hub app "calculates" body fat, water, and muscle by itself, which probably means that the stats are bullshit anyway. Since there's no access to those formulas or the raw impedance data over Bluetooth, body comp cannot be reproduced here.

## Changes
- Added a new handler: `core/bluetooth/scales/TaylorBIAHandler.kt`.
- Updated `ScaleFactory.kt` to check for the Taylor scale *before* the MGB scale, ensuring the right handler picks it up first.
- Added unit tests in `TaylorBIAHandlerTest.kt` to verify the weight decoding logic (including idle states and bit-boundary cases).

## Testing
- All unit tests pass locally.
- **Tested on a physical device** (Pixel 10 Pro, Android 16): The app now correctly identifies the scale (instead of calling it "SWAN / Icomon (MGB)"), connects smoothly, and locks in the weight using the stability check just like the official app.

## AI Disclaimer
- I used generative AI (Claude) to help write this solution.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/ScaleFactory.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/ScaleFactory.kt
@@ -61,6 +61,7 @@ import com.health.openscale.core.bluetooth.scales.SenssunHandler
 import com.health.openscale.core.bluetooth.scales.SinocareHandler
 import com.health.openscale.core.bluetooth.scales.SoehnleHandler
 import com.health.openscale.core.bluetooth.scales.SppScaleAdapter
+import com.health.openscale.core.bluetooth.scales.TaylorBIAHandler
 import com.health.openscale.core.bluetooth.scales.DrTrustSSW532Handler
 import com.health.openscale.core.bluetooth.scales.StandardBeurerSanitasHandler
 import com.health.openscale.core.bluetooth.scales.TrisaBodyAnalyzeHandler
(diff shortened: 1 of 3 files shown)


chropic:
[btsnoop_hci_sanitized_v2.log](https://github.com/user-attachments/files/28730418/btsnoop_hci_sanitized_v2.log)

oliexdev:
Thanks for your PR

PR #1396, opened 2026-06-11
feat: Added support for BF450 scale

VulcanoHex:
I just added the values for my Beurer BF450 scale in StandardBeurerSanitasHandler.kt and it worked.


oliexdev:
Thanks for your PR

PR #1408, opened 2026-06-25
Add minimal Huawei Hagrid WSP support

shisoratsu:
## Summary
- add a minimal Huawei/Honor Hagrid WSP scale handler and product detection
- implement WSP framing, CRC handling, authentication helpers, and explicit CAK/C1/C2 driver settings
- support conservative probe-only behavior when secrets are not configured
- parse basic realtime/history/standard measurement payloads without adding destructive scale-user commands

## Scope
This is the first slice of a stacked series. It intentionally does not include segmental impedance storage, scale-side user inventory/actions, or optional body-metric derivation. Those are split into follow-up branches to keep this review focused on the minimal protocol support.

## Verification
- `JAVA_HOME=/Library/Java/JavaVirtualMachines/temurin-24.jdk/Contents/Home ./gradlew testDebugUnitTest --tests "*HuaweiHagrid*" --tests "*MigrationTest*" --tests "*ImportExportUseCasesTest*"`

## Notes
- CAK/C1/C2 values must be provided explicitly by the user in driver settings. This code does not contain Huawei Health key extraction, Frida, APK, root, or native white-box logic.
- `upstream` is configured read-only locally; this PR is opened from the `shisoratsu` fork.


android_app/app/src/main/java/com/health/openscale/core/bluetooth/libs/HuaweiHagridSecretProvider.kt
```diff
@@ -0,0 +1,84 @@
...
+package com.health.openscale.core.bluetooth.libs
+
+/**
+ * Replaceable source for Huawei Hagrid CAK/C1/C2 key material.
+ *
+ * Implementations must not log raw secret values. The upstream app only uses
+ * explicitly configured settings here; private Huawei Health extraction tools
+ * are outside this boundary.
+ */
+interface HuaweiHagridSecretProvider {
```
oliexdev: Could you please add this into HuaweiHagridWspLib.kt so we have only one lib file.

oliexdev:
Thanks for your awesome PR :+1: , does this version already works?

oliexdev:
I have moved the SecretProvider to the lib and merged your PR. Thanks for your PR :+1:

PR #1417, opened 2026-07-07
Add Metabolic Age and Physique Rating derived values

DanyPM:
Two science-based derived values that Endurain and Tanita-class scales expose, computed automatically whenever a measurement carries a body-composition signal. Both follow the existing derived-value pattern (recomputed on insert/update, backfilled once on upgrade) and are seeded together in a single migration (`MIGRATION_15_16`, DB v16) — one schema hop from `master`.

## Metabolic Age
The age at which a population-average metabolism would match this person's. Derived by inverting the age term of the Mifflin-St Jeor BMR curve so it equals the individual's composition-based Katch-McArdle BMR (from fat-free mass). More lean mass → younger age. Both formulas are peer-reviewed and already used in openScale. Requires a body-fat/LBM signal (otherwise it would trivially return the chronological age).

image

## Physique Rating
Tanita's 1–9 body-type index: body-fat level × muscle level, each banded against the app's existing age/sex reference ranges (body-fat + Janssen 2000 muscle), mapped onto Tanita's 3×3 matrix (hidden-obese … very-muscular).

Because the rating is **ordinal** (9 isn't "more" than 1 on a linear axis), it's excluded from line charts / insights / statistics and shown categorically instead:
- Overview/detail rows show the body-type label; the overview row expands to a compact body-fat × muscle plane with the current point highlighted.
- Insights adds a "Your Physique Type" plane below "Your Body Composition" — the nine zones with band-transition %s, a 90-day-sampled trajectory + trend arrow, selectable past points, and a per-type description. Same fat × muscle axes as the composition card.

image

## Tests
Formula coverage for both (all nine physique zones, sex-specific bands, null-guards), Room round-trip insert/delete, and the migration producing both types.


DanyPM:
Updated with the addition of the "Physique Rating" value : 

https://tanita.eu/understanding-your-measurements/physique-rating

DanyPM:
@oliexdev would you have some time to a have a look at this work ?

Both new metrics are compatible with endurain :-)

oliexdev:
Hi @DanyPM, thanks a lot for the effort you put into this PR — the implementation follows the project architecture nicely and the test coverage is good. However, after reviewing it I've decided not to merge it, for two conceptual reasons:

  Metabolic Age: The formula equates two different BMR estimations (Mifflin-St Jeor and Katch-McArdle) that are based on different reference populations, so the resulting "age" is essentially an artifact of the formula mismatch rather than a validated physiological quantity. Mathematically it boils down to a linear rescaling of lean body mass for a given weight and height — so it  adds no new information beyond the BMR/LBM values openScale already shows, just dressed up in "years". I don't want to persist such a pseudo-metric in the user's data.

  Physique Rating: The 1–9 rating is exactly the cross product of the body-fat and muscle evaluation bands that openScale already displays in the evaluation gauges, so information-wise nothing new is added. The new insights card is also visually very close to the existing body-composition plane card, which would leave two near-identical fat × muscle planes on the same screen. If anything, I'd rather see the named zones integrated into the existing body-composition card as a display-only enhancement — computed on the fly, without a new persisted measurement type.

  I appreciate the contribution and hope this doesn't discourage you.

PR #1418, opened 2026-07-08
Add support for Yunmai X (YMBS-M268) broadcast-only scale

pagarsky:
## Summary

Adds a handler for the **Yunmai X (YMBS-M268)**, which was not detected by openScale at all. Unlike the Yunmai SE/Mini (GATT), the Yunmai X **broadcasts** its measurements as manufacturer-specific advertisement data — no GATT connection, and it advertises **no local name** (the vendor app calls this device family "YUNMAI-ADV").

## Protocol

16-byte manufacturer AD (type `0xFF`), reverse-engineered from an advertisement capture and confirmed end-to-end against a live measurement driven through this handler:

| bytes | meaning |
|---|---|
| `[0..5]` | device MAC, byte-reversed |
| `[6..8]` | signature `0B 74 17` |
| `[9]` | per-session byte (varies between weigh-ins; ignored) |
| `[10]` | state: `01` live, `02` stable, `03` final (incl. impedance) |
| `[11..12]` | weight, u16 BE, 0.01 kg |
| `[13..14]` | impedance, u16 BE, Ω (0 until final) |
| `[15]` | XOR checksum over `[9..14]` |

Body composition (fat/water/muscle/bone/LBM/visceral fat) is derived from weight + impedance via the existing `YunmaiLib`, consistent with the Yunmai SE/Mini handler.

## Identification

The scale advertises no name, so matching uses the frame structure — MAC echo + `0B 74 17` signature + checksum — during scans, and the advertised 16-bit service UUID `0x1320` for saved-device snapshots (which carry no manufacturer data). These checks distinguish it from other 14-byte manufacturer frames from nearby devices.

## Changes

- `YunmaiXHandler` (+ `YunmaiXAdv` parser), registered in `ScaleFactory`.
- 15 unit tests with synthetic fixtures built to the observed layout.
- saved-scale panel now falls back to the handler display name when a scale advertises no name (previously rendered a blank line above the MAC).
- README: list the Yunmai X.

## Testing

`./gradlew testDebugUnitTest` passes (15 new tests, 0 failures); `assembleDebug` builds. Validated on real hardware: a full barefoot measurement produced weight + impedance and full derived body composition stored in openScale.

openscale_saved_redacted

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ openScale has built-in support for a number of Bluetooth (BLE or "smart") scales
 - Silvercrest SBF75, SBF77
 - Vigorun
 - Xiaomi Mi scale v1 and v2
-- Yunmai Mini and SE
+- Yunmai Mini, SE and X
 - iHealth HS3
 - Easy Home 64050
 - and many more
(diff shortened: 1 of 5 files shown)


pagarsky:
@oliexdev sorry for the ping, can you take a look? Would appreciate if this can lend in the next release! ❤️

oliexdev:

PR #1419, opened 2026-07-11
feat: add Scale Up broadcast scale protocol

brainrom:
Adds support for BLE Broadcast protocol used in [Scale Up app](https://play.google.com/store/apps/details?id=com.vtrump.qingqing) for many low-cost smart scales. The app is very obfuscated, so I don't know actual name of the protocol and how to properly identify it (even MAC on my scale doesn't have correct vendor). Added heuristic, which should recognize scales with such protocol.


oliexdev:
Thanks for your PR :+1:

PR #1420, opened 2026-07-13
Added scale support for EEBBL scale.

tannermeade:
## Summary
Adds support for an **EEBBL** scale by introducing a **new handler** and wiring it into the device detection/initialization flow. (https://www.amazon.com/dp/B09KCH2BPW)

## Changes
- Added a new handler for EEBBL scales (e.g., `EEBBLScaleHandler`)
- Registered the handler in the scale factory/dispatcher so EEBBL devices are detected and handled correctly
- Implemented the EEBBL-specific protocol parsing/mapping

## Implementation Notes
- Handler is responsible for:
  - Identifying the EEBBL device/protocol signature
  - Decoding incoming payload(s)

## Files
- `src/…/handlers/EEBBLHandler.kt` (new)

Weight is working. Other body composition data is only in impedance value form and the official Fitdays app has the algo for that.


oliexdev:
Thanks for your PR I have just removed "DEBUG" from supported device names.

Issue #1410, opened 2026-06-29
Database import error from v2.X.X to v3.X.X


txanetxarra:
### Preflight Check (Required)

- [x] I tested the latest dev version https://github.com/oliexdev/openScale/releases/tag/dev-build
- [x] The issue still occurs in the dev version
- [x] I searched the existing open and closed issues
- [x] This issue has not already been reported or fixed

### Bug Description

Hello,

I'm trying to migrate from version v2.3.1 to v3.1.1.

I've been trying to export my database from an english configured device (cell phone - Android 9) with openScale v 2.3.1 to an spanish configured one (Tablet - Android 13) with a freshly installed openScale v3.1.1.

The tablet is reporting that the database has been successfully imported and that I must restart the app.

I restart the App but the app does not start.

I've also tried the process from openScale v 2.3.1 but the result is the same.

Thank you in advance for your help.

And thank you so much for developing the openScale project.

Best regards.

### Steps to Reproduce

Problem 1 - Database import error
 
1. Export database from an english configured device (cell phone - Android 9) with openScale v 2.3.1
2. Import database into an spanish configured one (Tablet - Android 13) with a freshly installed openScale v3.1.1
3. The tablet is reporting that the database has been successfully imported and that I must restart the app.
4. Restart the App.
5. Try to open again openScale app but the APP does not start anymore.
6. Delete APPs cahe from APP info.
7. Try to open again openScale app but the APP does not start.
8. Completely delete storage from APP info.

Problem 2 - Problem creating the first user in the APP

1. Try to add a new user (Tablet - Android 13 - openScale v3.1.1).
2. openScale keeps saying "please input correct data"
3. uninstalled and installed v3.
4. The problem persisted for some time until I did something (I don't know what) and suddenly I started being able to add users.

### Debug Log (Required)

```shell
No bug log was possible, as the APP crashes when importing the database.  No way to get to the "Export log file" button.
```

DanyPM:
Hi,

I had a look and you have two issues here, for which I have suggested fixes:

- database migrations before schema version 6 were not transposed from java to Kotlin : #1421
- user-entered decimals only accepted the point separator, unlike the comma that is used in some languages like Spanish : #1422 

I have built an APK (in the attached zip) from these fixes and would advise you to
- install it
- import you db
- export your db
- uninstall it
- install the stable openscale app

Can you confirm it's working ?

[openScale-debug.zip](https://github.com/user-attachments/files/30063768/openScale-debug.zip)

[...]

PR #1422, opened 2026-07-15
Accept comma decimal separator in number input (issue #1410)

DanyPM:
Add one shared helper and use it at the three broken FLOAT sites (INT sites need no change)

Per #1410

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/utils/LocaleUtils.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/utils/LocaleUtils.kt
@@ -181,4 +181,14 @@ object LocaleUtils {
             isGroupingUsed = false
         }.format(cleaned)
     }
+
+    /**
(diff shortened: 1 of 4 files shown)


oliexdev:
Thanks for the PR — the , → . idea is spot on. I went a slightly different route to fix it closer to the root, see https://github.com/oliexdev/openScale/commit/cf1f27efe8294a4309cf591b9a2faf1fbc7b5d4d:

  - Normalize at the source: the shared NumberInputField already accepts a comma while typing, so I just made it emit a dot-normalized value. That fixes every downstream toFloatOrNull() at once, instead of patching each parse site.
  - Real trigger of #1410: the height field used KeyboardType.Number (no decimal key on many keyboards) — switched it to .Decimal + comma normalization.
  - Skipped CSV import: there the comma is the field delimiter, and export always writes dots, so a replace would be a no-op / ambiguous.

  Thanks again — the clear write-up made this easy to reason about!

Issue #1415, opened 2026-07-04
[Feature Request] Add a third option for user gender


Francokab:
I think adding a "Other" or "Not Specified" option for the gender of users would allow to more easily take into account the variety of way people can exist.

As some body metrics uses different formula for male and female, those metrics should be disabled if this third option is taken.

Stephan-Thomas:
Hi, i would like to work on this.

oliexdev:
Thanks for your issue, please read the decline PR #1424

PR #1424, opened 2026-07-17
Add DIVERSE gender option and disable sex-specific metrics (#1415)

anubhav9369:
Implements #1415 — adds a third gender option for users who don't identify as male or female.

## What
- New `GenderType.DIVERSE` option (shown in the user gender dropdown).
- Metrics that rely on male/female-specific formulas are disabled for DIVERSE users, as requested in the issue:
  - Evaluation (in/out-of-range coloring) returns `UNDEFINED` for body fat, body water, muscle, LBM, BMI, WHR, waist circumference and the BMI-derived target-weight range.
  - Derived-value calculation returns `null` (metric not stored) for BMR and the skinfold/caliper body-fat estimate.
  - Body-fat / body-water / LBM estimation formulas are skipped, leaving any device-provided values untouched.
- Non-sex-specific metrics (weight, BMI value, WHtR, visceral fat, circumferences) are unaffected.

## Notes
- A new `GenderType.hasSexSpecificReference()` helper centralizes the rule.
- Gender is stored as enum-name text, so **no database migration** is required.
- BLE scales that compute body composition on-device still use their internal binary reference; wiring DIVERSE through those handlers is left as a follow-up.

## Tests
- Added DIVERSE coverage to `MeasurementEvaluatorTest` and `DerivedValuesCalculatorFormulasTest`.
- `./gradlew testDebugUnitTest` passes (JDK 21, matching CI).

Open to feedback on the behaviour — e.g. using a neutral reference instead of disabling for DIVERSE.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/data/Enums.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/data/Enums.kt
@@ -155,10 +155,22 @@ enum class SupportedLanguage(val code: String, val nativeDisplayName: String) {
 
 enum class GenderType(@param:StringRes val displayNameResId: Int) {
     MALE(R.string.gender_male),
-    FEMALE(R.string.gender_female);
+    FEMALE(R.string.gender_female),
+    DIVERSE(R.string.gender_diverse);
 
     fun isMale(): Boolean {
         return this == MALE}
+
+    /**
+     * Whether this gender maps to a sex-specific reference table / body-composition
+     * formula. Body-composition metrics (body fat, water, muscle, LBM, BMR, ...) rely
+     * on male/female specific formulas and reference ranges. [DIVERSE] has no such
+     * medically defined reference, so those metrics are disabled for it (issue #1415):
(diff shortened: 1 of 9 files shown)


oliexdev:
I'm going to decline it, and I'd rather be clear about why.

  In openScale the gender field is only ever a biological input to the body-composition math — it feeds the BIA formulas, the reference ranges, and the scale protocols (which themselves accept only male/female). It isn't used to represent identity anywhere in the app, for any user; it's purely a calculation parameter, and it means the same narrow thing for everyone.

  So there's no identity layer here for a third value to extend — only the math, and the math needs a male/female reference. That's also why the PR can't really deliver its goal:
  - The ~40 scale handlers all branch on isMale(), so a DIVERSE user still gets fat/water/muscle computed as female, just shown uncolored — not neutral, only unlabeled.
  - The rest is switched off silently, while non-sex metrics (WHtR, visceral fat) stay colored — an incoherent screen with nothing explaining why.

  Since this field is a biological parameter and not an identity setting, a third option doesn't fit here. I really appreciate you raising it and the effort you put in.

PR #1427, opened 2026-07-21
Fix MiScale +2h/+1h timestamp offset regression (revert to local timezone)

jvandenbroek:
## Summary

Xiaomi Mi Scale / Mi Body Composition Scale (`MIBFS`/`MIBCS`) measurements have been recorded with a timestamp offset by exactly the device's local UTC delta (e.g. +2h during CEST, +1h during CET) since commit cd4d5e16 ("Fix: Use UTC for MiScale time synchronization").

That commit changed `MiScaleHandler.writeCurrentTimePreferPrimary` and `MiScaleHandler.parseMinuteDate` from using the device's default (local) timezone to hardcoded UTC, for both writing the BLE Current Time characteristic and parsing the time echoed back by the scale.

The scale itself has no timezone concept - it only stores/echoes plain wall-clock digits (year/month/day/hour/minute/second) and expects those to represent local time. Writing UTC digits while still parsing the reply as UTC produces a timestamp that is off by the local UTC offset, because the round trip no longer matches the convention the firmware assumes.

## Evidence

I cross-referenced years of measurements from a MIBFS scale recorded by openScale against an independent BLE reader of the same physical device (a different companion app that also supports this scale and logs its own timestamps). Every measurement recorded before cd4d5e16 (2021 through Sept 2025) matches the independent reader exactly (0.0h delta, hundreds of samples). Every measurement recorded since cd4d5e16 (Apr 2026 onward, once the scale was used again) is off by exactly the local UTC offset at the time of measurement (+2.00h, matching CEST).

I could not find a linked issue or repro for cd4d5e16's stated rationale ("previously the device's local timezone was used, which could lead to incorrect timestamps") - it appears to have been a direct commit without an associated bug report. The data above contradicts that rationale for this device.

## Change

Reverts the two timezone lines in `MiScaleHandler.kt` to their pre-cd4d5e16 behavior (`Calendar.getInstance()` / `TimeZone.getDefault()`), while keeping the unrelated day-of-week/adjust-reason trailing payload byte fix from the same commit, which is a separate and still-valid protocol correction untouched by the timezone regression.

## Test plan

- [x] `./gradlew :app:compileDebugKotlin` builds successfully with this change
- [x] Verified against real-world measurement history from an affected device (see Evidence above); no unit test exists for `MiScaleHandler`'s time parsing (only `MiScaleLibTest.kt`, unrelated body-composition math) so this relies on the historical data comparison rather than an automated regression test
- [ ] Would appreciate a maintainer/other MIBFS-MIBCS user confirming on their own device, since I can't verify behavior across all Mi Scale v1/v2 firmware variants

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/MiScaleHandler.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/MiScaleHandler.kt
@@ -202,7 +202,7 @@ class MiScaleHandler : ScaleDeviceHandler() {
 
     /** Prefer writing Current Time to the primary service; fallback to alternate if needed. */
     private fun writeCurrentTimePreferPrimary(primarySvc: UUID, alternateSvc: UUID) {
-        val c = Calendar.getInstance(TimeZone.getTimeZone("UTC"))
+        val c = Calendar.getInstance()
         val year = c.get(Calendar.YEAR)
         val payload = byteArrayOf(
             (year and 0xFF).toByte(), ((year shr 8) and 0xFF).toByte(),
@@ -455,7 +455,7 @@ class MiScaleHandler : ScaleDeviceHandler() {
     private fun parseMinuteDate(y: Int, m: Int, d: Int, h: Int, min: Int): Date? =
         runCatching {
             val sdf = SimpleDateFormat("yyyy/MM/dd/HH/mm", Locale.US)
-            sdf.timeZone = TimeZone.getTimeZone("UTC")
+            sdf.timeZone = TimeZone.getDefault()
             sdf.parse("$y/$m/$d/$h/$min")
         }.getOrNull()
 


oliexdev:
Before merging I'd like to settle one open question with a log, because I think
  both the old and the new behaviour may be masking the same underlying issue.

  The scale's clock is free-running and has no DST concept, which was the original
  motivation for cd4d5e16: after a DST change the scale keeps reporting winter wall
  time while we parse it as summer time. That reasoning only holds if our Current
  Time write actually reaches the scale — and that's exactly what we've never
  verified, because `writeTo()` is fire-and-forget and the success check in
  `writeCurrentTimePreferPrimary` can never return false.

  Could you (or another MIBFS/MIBCS user) send an openScale log?

  1. Settings → enable file logging
  2. Note the exact wall-clock time, then **step on the scale and take a measurement**
  3. *Afterwards*, open openScale and sync the device
  4. Send the log, plus the noted time and your timezone

  The order matters — the measurement has to happen before the sync, otherwise the
  timestamp doesn't come from the scale's own clock.

  Also: do you have Mi Fit / Zepp Life installed and paired with this scale? That's
  the most likely confounder — if it resets the clock to local time, UTC parsing can
  never work regardless of what we write.

  What the log will tell us, from the `2a2b` (Current Time) lines:
  - `write response … status=0` → the scale accepts our clock write, and the +2h has
    a different cause (most likely another app overwriting it)
  - `status=3/5/15` → we never set the clock at all, and this PR works only because
    Mi Fit happens to keep it in local time

  Either way it turns a guess into a measurement.

jvandenbroek:
Ran the requested test. Log + DB extract from a live weigh-in on the affected device (MIBFS scale) with the currently-shipped (unpatched) app:

- `2a2b` write at 23:03:57 local: payload `EA 07 07 15 15 03 39 00 00 01` → decodes to `2026-07-21 21:03:57` (UTC digits, 2h behind wall clock) — confirms the shipped app still writes UTC per this commit. **Write response: `status=SUCCESS`** at the actual GATT ATT layer (not just the app's derived boolean), so the write does reach the scale — the `writeTo()` fire-and-forget concern doesn't appear to be the cause here.
- Newest history record read back in the same sync: digits `21:03:34`, parsed as UTC → `23:03:34` local. **Matches the real measurement time — no offset** in this fresh test, because the clock-write happened seconds before the read (self-consistent UTC round trip).
- openScale DB confirms: new record timestamp `2026-07-21 23:03:00` — correct.

So a fresh weigh-in immediately following a sync does **not** reproduce the bug — it's internally self-consistent whether the convention is UTC or local, as long as write and read agree.

The actual failure mode only shows up for **stored history entries the scale accumulates between syncs** (visible in the same log dump — the scale buffers several older weigh-ins with their own embedded clock digits, read back in bulk on next connect). This device is also paired with a second companion app ("Notify & Fitness for Mi Band", `com.mc.miband1`) that independently connects to the same scale. That app's own historical weight log has matched true local time exactly for years — it was in fact used as the independent ground truth to originally diagnose this issue — consistent with it writing/reading the scale's clock in local-time convention. So the actual failure path is:

1. Notify & Fitness (or any other companion app) connects to the scale between openScale syncs and sets the clock using local-time digits (its own convention).
2. A weigh-in happens while the scale still holds those local digits.
3. openScale later connects, and reads that stored history entry — parsing local digits as UTC (post-regression code) → systematic +2h/+1h error, matching the flat DST-sized offset found in the DB. With the revert applied, the same digits parse as local → correct, regardless of which app last set the clock.

Net: openScale's own round trip is fine either way *in isolation*, but real devices in this class are commonly shared with other BLE companion apps that use local-time convention. Reverting to local time makes openScale's convention match the rest of that ecosystem, closing this failure path without needing to detect or block other apps from touching the scale.

oliexdev:
Thanks for your feedback I have merged it.

PR #1430, opened 2026-07-22
Add support for the FitTrack Dara 2.0 scale

### LINKED ISSUE #1016: FitTrack Dara 2.0
Scale name: **FitTrack Dara 2.0**
A link to the vendors app: https://play.google.com/store/apps/details?id=com.elink.fittrackhealth.pro

I would be happy to by your **openScale pro** if it supports FitTrack Dara 2.0 scale.
Please let me know.
Thanks.

richardlt:
Closes #1016.

## What
Adds `FitTrackDaraHandler` for the **FitTrack Dara 2.0** (vendor app "Hume by FitTrack",
`com.elink.fittrackhealth.pro`). It's an eLink/Icomon "MGB"-family scale on service `0xFFB0`,
so it shares framing with `MGBHandler` but advertises as `FitTrack` and uses a different
body-composition payload — hence its own handler.

It's registered ahead of `MGBHandler` (which also matches any `0xFFB0` device) so the specific
name match wins.

## Measurements
- Weight (with kg/lb/st conversion), body fat %, body water %, lean body mass, bone, protein %,
  visceral fat index.
- Time, user profile (sex/age/height) and unit are synced to the scale on connect.
- BMI and BMR are left to openScale's own derivation. Subcutaneous fat and metabolic age are
  sent by the scale but have no `ScaleMeasurement` field, so they're not stored.

## How it was verified
- Protocol reverse-engineered from a `btsnoop_hci.log` capture.
- Built and tested on a real FitTrack Dara 2.0. Weight is confirmed correct. The
  body-composition field mapping and per-field scaling were derived by comparing live readings
  against **older Hume-app records** for the same user — the Hume app no longer pairs with the
  Dara, so a simultaneous side-by-side wasn't possible.
- Unit tests cover the frame checksum and weight decode.

## Notes
- The vendor app performs an extra `AE/AD` challenge-response on connect; it is not required to
  receive weight or the body-composition dump, so it isn't replayed.
- If the dump doesn't arrive (e.g. a firmware variant), weight is still recorded.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/ScaleFactory.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/ScaleFactory.kt
@@ -21,6 +21,7 @@ import android.content.Context
 import com.health.openscale.core.bluetooth.scales.BeurerBF450Handler
 import com.health.openscale.core.bluetooth.scales.ScaleDeviceHandler
 import com.health.openscale.core.bluetooth.scales.AAAxHandler
+import com.health.openscale.core.bluetooth.scales.FitTrackDaraHandler
 import com.health.openscale.core.bluetooth.scales.ScaleupHandler
 import com.health.openscale.core.bluetooth.scales.ActiveEraBF06Handler
 import com.health.openscale.core.bluetooth.scales.CultSmartScaleProHandler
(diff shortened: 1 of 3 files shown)


oliexdev:
Thanks for your PR :+1:

PR #1432, opened 2026-07-26
Add Runstar R6 scale support

ldoench:
Adds a handler for the Runstar R6, reverse-engineered from an HCI snoop of the vendor app plus field tests against the device.

**Working:** live weight, heart rate, impedance, and syncing offline-stored history entries (weigh-ins taken while disconnected).
**Not implemented:** body composition (fat/water/muscle) — the scale only exposes raw impedance, no derived values.

The R6 shares its FFB0 frame family with the existing `RobiS9Handler` but is a different, incompatible protocol from the R5 despite the similar name — `RunstarR5Handler.supportFor()` previously matched any `RUNSTAR-*` name and would silently swallow R6 devices, so R5 now excludes `RUNSTAR-R6` explicitly and R6 is registered ahead of it.

Btsnoop capture attached.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/ScaleFactory.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/ScaleFactory.kt
@@ -134,8 +134,6 @@ class ScaleFactory @Inject constructor(
         MiScaleS400Handler(),
         XiaomiS800Handler(),
         MiScaleHandler(),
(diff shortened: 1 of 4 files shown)


ldoench:
[runstar_r6_btsnoop.log](https://github.com/user-attachments/files/30387581/runstar_r6_btsnoop.log)

oliexdev:
Thanks for your PR :+1:

PR #1441, opened 2026-08-04
Add Relaxmedic scale support (icomon protocol 107)

Caiofiga:
Adds a Handler for the Relaxmedic FG2015B-A scale, reverse-engineered by decompiling the vendor app, plus testing on the physical device.
**Working:** Weight; Body fat; Water; Muscle; Bone; Visceral fat; Protein; BMR; LBM.
**Not Implemented:** Offline history (Not supported);Segmented values.

It shares service `0xFFB0` and the 20-byte frame family with `RobiS9Handler` and `RunstarR6Handler`, but has fragmented frames, where byte 2 works as an index, with the result spanning 2 frames. 

Attached is the [btsnoop](https://github.com/user-attachments/files/30682805/relaxmedic-fg2015b-a.log) capture, filtered to this scale's traffic only.


oliexdev:
Thanks for your PR :+1:

PR #1445, opened 2026-08-06
Add support for Dr. Trust Smart 505 Scale (India)

joshferns:
Add 8-byte streaming protocol support to MGBHandler (Dr Trust Smart 505)

MGBHandler matches by name (swan/icomon/yg) or service 0xFFB0. The Dr Trust Smart 505 matches but sends 8-byte frames, while the handler only accepted 20-byte composite frames (if (data.size != 20) return) — so it connected, ran its init writes, and silently saved nothing.

This makes onNotification() dispatch on frame size and adds a path for the 8-byte variant. The 20-byte path is unchanged (extracted verbatim into a private method).

Protocol (8-byte variant)
Frame: AC 02 [b2] [b3] [b4] [b5] [flag] [chk], chk = (b2+b3+b4+b5+flag) & 0xFF. Flag byte: CE live weight, CA settled weight (u16 in b2:b3, 0.01 kg), CB impedance (FD 01 [hi] [lo], ohms), CC status. Completion is the CE→CA flag change, not a separate frame.

Implementation

Pure companion-object parsers (isValidStreamingFrame, parseFinalWeightRaw, parseLiveWeightRaw, parseImpedanceOhm), following VitafitVT701Handler's pattern.
Latches weight + impedance, publishes once both arrive, resets in a new onDisconnected().
Composition via StandardImpedanceLib — same library and sanity guard VitafitVT701Handler uses; raw impedance also stored. No new formulas.

Tests
New MGBHandlerStreamingTest.kt covering weight/impedance parsing, the flag distinctions, the FD-config-vs-impedance collision, and checksum/size rejection. Passes; existing VitafitVT701HandlerTest still passes.

Tested on a physical Dr Trust Smart 505: weigh-in saves correctly, and weight/BMI/bone match the vendor app. Composition is derived from raw impedance via StandardImpedanceLib, so it differs from the vendor's proprietary values (as with other impedance scales here) — happy to swap the estimator if preferred. I don't own the 20-byte variant, so only the 8-byte path is hardware-verified.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/MGBHandler.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/MGBHandler.kt
@@ -145,7 +145,22 @@ class MGBHandler : ScaleDeviceHandler() {
         if (characteristic != CHAR_CTRL) return
         when (data.size) {
             20 -> onCompositeFrame(data)
-            8 -> onStreamingFrame(data, user)
+            8 -> {
+                if (streamingPublished) return
+
+                parseFinalWeightRaw(data)?.let { raw ->
+                    streamingWeightRaw = raw
+                    logD("streaming final weight raw=$raw")
+                    tryPublishStreaming(user)
+                    return
+                }
+
+                parseImpedanceOhm(data)?.let { ohm ->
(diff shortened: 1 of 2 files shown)


oliexdev:
Thanks for your PR :+1:

Issue #547, opened 2020-02-02
Huawei Body Fat Scale AH100


dmitriytim:
datetime 02-02-2020 20:57
Weight  82,6
BMI 24,8
FAT 22,1
Visceral FAT 11,5
Basal metabolic rate 1831
water 54,1
Bone mineral 3,3
protein 19,9
Muscles 62,0
[btsnoop_hci_20200202205551.cfa.txt](https://github.com/oliexdev/openScale/files/4145257/btsnoop_hci_20200202205551.cfa.txt)
[2020-02-02.zip](https://github.com/oliexdev/openScale/files/4145260/2020-02-02.zip)

oliexdev:
Please use the correct template and do **all** the steps probably.

KerstinMaur:
Hey, I am interested in support for that Huawei AH-100 scale too as I might purchase a smartwatch where they bundle that scale at the moment.
Can you please provide @oliexdev with the provided information, @dmitriytim? 🙏

dmitriytim:
edited

fonshartendorp:
I'm also interested in added support for the AH-100, but have zero experience in reverse engineering. Is the correct template used now and are all steps done properly? I'd be happy to help

diegoalbertopp:
And what happened?

klosteinmann:
Oh here's the original. Sorry for duplication so does the scale properly work without workarrounds?

meigor:
I also want to contribute to the support of the Huawei AH100 (and Honor CH100 sem to be the same, my is a Huawei AH100 but bordcast the CH100 as model). 
i tried to get the log file with adb bugreport _NAME extract it and convert with btsnooze.py [btsnooze info](https://source.android.com/devices/bluetooth/verifying_debugging#debugging-with-bug-reports) but wireshark can't read the file.

So i tried to get some information with the nRF app and got the following log:

```
nRF Connect, 2021-03-23
CH100 (5C:CA:D3:2C:7F:E4)

D	17:46:16.744	gatt.setCharacteristicNotification(00002a9c-0000-1000-8000-00805f9b34fb, true)
D	17:46:16.745	gatt.setCharacteristicNotification(00002a9d-0000-1000-8000-00805f9b34fb, true)
D	17:46:16.746	gatt.setCharacteristicNotification(00002a99-0000-1000-8000-00805f9b34fb, true)
[...]

PR #1450, opened 2026-08-10
Huawei AH100/CH100: fix the user record sent to the scale

### LINKED ISSUE #1449: Huawei AH100/CH100: user record is sent with weight 0, scale re-requests it 127x per session
### Preflight Check (Required)

- [x] I tested the latest dev version
- [x] The issue still occurs in the dev version
- [x] I searched the existing open and closed issues
- [x] This issue has not already been reported or fixed

### Bug Description

`HuaweiAhCh100Handler` sends a user record with **weight 0** to the scale, and the
scale answers by re-requesting it in a tight loop: **127 `USER_CHANGED` (op 0x20)
round trips in 33 seconds**, each answered with a byte-identical encrypted
`USER_INFO` frame. The loop only stops once an actual measurement completes.

Cause is in `sendUserInfo`:

```kotlin
val w = (weightTenthKg ?: (user.initialWeight * 10f).toInt()).coerceAtLeast(0)
```

`weightTenthKg` comes from `lastMeasuredWeightTenthKg`, which is `-1` until the
first measurement of the session, so on connect it falls back to
`user.initialWeight` — and that is `0f` for a freshly created profile.

Decrypting the `USER_INFO` we send (XOR with the scale MAC, then AES-CTR with the
session magic key) confirms it:

```
11 22 33 44 55 10 01 | 29 | B9 | 00 | 00 00 | FF FF | 1C E2
auth token             41y  185cm pad  weight  resist  const
                                       = 0 kg
```

`HuaweiCH100SHandler.sendUserInfo` has the identical line and the same defect.

The measured values themselves are correct — this is not #1206 / #1280 resurfacing.
Weight, body fat, timestamp and weekday all decode exactly as the scale's own
display showed them.

### Steps to Reproduce

1. Fresh install, create a user profile without entering an initial weight.
2. Add a Huawei AH100 / CH100 (mine advertises as `CH100`, Chipsea CST34M97).
3. Connect and watch the log before stepping on the scale.
4. `op=0x20` / `CMD* 0x09` repeat roughly every 250 ms until a measurement lands.

### Debug Log (Required)

openScale 3.1.2 built from master `7a170e40`, Pixel 10 Pro, Android 17 (API 37).

```
10:01:03.750 D/HuaweiAhCh100Handler: ← op=0x20 len=1 7D
10:01:03.752 D/HuaweiAhCh100Handler: → CMD* 0x09 len=16 (encrypted)
10:01:03.838 D/GattScaleAdapter: → write to chr=…faa1 len=19 payload=[DC 10 09 BB 27 B1 6E 64 C1 A2 43 4F 99 D0 FD B5 F9 95 96]
10:01:04.136 D/GattScaleAdapter: ← received data chr=…faa2 len=4 payload=[BD 01 20 21]
10:01:04.138 D/HuaweiAhCh100Handler: ← op=0x20 len=1 7D
10:01:04.141 D/HuaweiAhCh100Handler: → CMD* 0x09 len=16 (encrypted)
   … 125 more identical cycles …
10:01:40.294 I/HuaweiAhCh100Handler: Measurement: 126.6 kg, fat=33.8%, impedance=306 Ω, userId=1
```

Note the `payload=` bytes are identical in every cycle.

With a fallback chain (last stored measurement → profile initial weight →
BMI-22 estimate from body height, never 0) the same hardware drops to **5 cycles**.
Verified on the device, not only in unit tests. Happy to open a PR.

martinriedel:
Closes #1449.

Two defects in the `USER_INFO` record this handler pushes to the scale. Both
confirmed on real hardware (advert name `CH100`, Chipsea CST34M97).

> **Two corrections to my own earlier claims, kept visible rather than edited out.**
>
> 1. I first claimed the weight fix collapsed the `USER_CHANGED` loop from 127
>    cycles to 5. Wrong: the cycle *count* mostly measures how long the user
>    takes to step on, since the loop runs until a measurement arrives.
> 2. I then claimed the length byte made the scale poll four times slower.
>    Also wrong. Re-measured on one phone, one scale, two runs six minutes
>    apart: 300 ms median either way.
>
> Both fixes stay. Neither is a behaviour fix — see below.

## 1. Never send a zero weight

`sendUserInfo` falls back to `user.initialWeight` when no measurement has been
taken yet in the session, and that field is `0f` on a fresh profile. Decrypted
from the wire on master:

```
11 22 33 44 55 xx xx | .. | .. | 00 | 00 00 | FF FF | 1C E2
auth token             age  cm   pad  weight  resist  const
                                      = 0 kg
```

`profileWeightKg()` — private to each of the two Huawei handlers — walks last
stored measurement → profile initial weight → BMI-22 estimate from body
height, and never returns 0. On hardware: the first ever connect sends the
BMI estimate, the next connect sends the previous measurement.

~~USER_CHANGED cycles: 127 → 5.~~ No measurable effect on scale behaviour.
This is a data-correctness fix.

## 2. Declare 14, not 16, in the length byte

openScale 2.5.4 built a 16-byte payload and declared 14:
[`BluetoothHuaweiAH100.java#L514-L524`](https://github.com/oliexdev/openScale/blob/v2.5.4/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothHuaweiAH100.java#L514-L524)
passes the literal `14` for `authCode(7)` plus 9 bytes, and
[`#L674-L684`](https://github.com/oliexdev/openScale/blob/v2.5.4/android_app/app/src/main/java/com/health/openscale/core/bluetooth/BluetoothHuaweiAH100.java#L674-L684)
writes it as `{0xDC, len + 0, cmd}`. The trailing `0x1C 0xE2` is transmitted
but not counted. The vendor app does the same. The 3.x port pulled the trailer
into the payload and moved the length byte to 16 with it.

`HuaweiCH100SHandler` agrees independently: it builds the record without the
trailer and declares its full 14 bytes.

~~Four times slower once the record is declared the way the vendor declares
it.~~ Measured on one phone, one scale, two runs six minutes apart:

| length byte | polls | duration | median interval |
|---|---|---|---|
| `0x0E` (14) | 139 | 36.3 s | 300 ms |
| `0x10` (16) | 259 | 72.0 s | 300 ms |

Identical, in every 10-second block of both runs. The earlier 1001 ms figure
came from a different phone and is most likely its BLE connection interval.
What the hardware does confirm is that the scale **accepts** the record at
`0x0E`: the measurement completes normally and matches the scale's display.

So this is a correctness fix too — it sends the record the way the vendor and
2.5.4 send it, nothing more.

## Not in scope: the body fat value

Worth stating explicitly, given this handler's history of "wrong values"
reports. **The measured values are correct** — weight, body fat, impedance,
timestamp and weekday all decode exactly as the scale's display shows them.
This is not #1206 / #1280 resurfacing.

Body fat arrives fully computed in the frame and does not respond to anything
we send. `StandardImpedanceLib` cannot do better: against Huawei's own numbers
it reads 10–14 points low on every sample I have, including the vendor-app
captures in #547.

## Tests

Most fixtures in `HuaweiAhCh100HandlerTest` were generated by a second
implementation of the *same assumed* layout, so they can only show that two
ports of one guess agree. A `Real hardware capture` section adds frames taken
off a scale and checked against its display.

The USER_INFO test in that file was worse than useless: it built a frame with
`explicitLen = 14` and then asserted the frame said 14 — green whatever
`sendUserInfo` did. Replaced by `HuaweiUserRecordWireTest`, which attaches both
handlers to a capturing transport, drives connect/wake/auth, and decrypts the
record off the wire. Its fixtures are synthetic; every expected value is
derived from the protocol.

Those tests were checked against mutations rather than assumed to work:
reverting the weight fallback and the length byte turns 5 of the 7 red, and the
2 that stay green are the ones that do not depend on those lines.

`HuaweiAhCh100HandlerTest` and `HuaweiUserRecordWireTest` pass. The suite as a
whole does not: `BackupRestoreUseCasesTest` has two failures, unrelated to this
branch and reproducible on master. An earlier version of this description
claimed the full suite passed; it did not.


oliexdev:
Thanks for your PR :+1: 

  I've pushed one change to your branch: the fallbackWeightKg / profileWeightKg pair is out of ScaleDeviceHandler again, and the logic now sits as a private profileWeightKg(user)
  in each of the two Huawei handlers. Behaviour is identical — lastMeasurementFor() is already protected on the base class, so nothing had to be added there.

  Could you re-run this against your CH100 before I merge? The user record path is exactly what changed, and I'd rather have it confirmed on the hardware than reason about it.

martinriedel:
Will check and reply

oliexdev:
Did you have time to test?

martinriedel:
Tested on hardware. New phone, fresh install, fresh profile with no initial weight, CH100. Two runs six minutes apart, identical except the length byte.

**Weight fix works.** On the first ever connect the record carries a plausible weight derived from body height, where master sends `00 00`. On the next connect it carries the measurement from the first run. Your inlined `profileWeightKg` produced both.

**The scale accepts the record at `0x0E`** — measurement completed normally and matched the scale's own display.

**I withdraw the poll-rate claim.**

| length byte | polls | duration | median |
|---|---|---|---|
| `0x0E` | 139 | 36.3 s | 300 ms |
| `0x10` | 259 | 72.0 s | 300 ms |

Same in every 10-second block of both runs, standing on the scale or not. The old 1001 ms was measured on a different phone — most likely its BLE connection interval, not the scale reacting to the record.

So neither change alters the scale's behaviour. What stands: weight 0 is wrong data, and 14 is what v2.5.4 declares — `BluetoothHuaweiAH100.java:524` passes the literal `14` for a payload it builds as 16 bytes.

Pushed: `HuaweiUserRecordWireTest` drives both handlers and decodes the record off the transport. The old test asserted a length it passed in itself; removed. Verified the new ones can fail — reverting each fix turns 5 of 7 red.

Also: "Full unit test suite passes" in the description was wrong. `BackupRestoreUseCasesTest` fails twice, on master too.

oliexdev:
Thanks, merged.

  One note: `BackupRestoreUseCasesTest` passes here, all 4 tests, on your branch and on master with
  `--rerun-tasks`. If it reproduces for you, please open a separate issue with the failure output —
  sounds environment-specific and I'd like to know what differs.

PR #1451, opened 2026-08-10
Complete the Turkish translation (154 -> 782 of 811 strings)

ahmetabdullahgultekin:
Turkish was the least complete of the maintained locales: **154 of 811 strings (19%)**, while German and Spanish sit at 95%, Russian at 87% and Italian at 84%. Four out of five strings a Turkish user sees are still English, which leaves Bluetooth setup, insights, backup and every error message unreadable for anyone who doesn't read English.

This adds the **628 missing translatable strings**.

- Existing translations are untouched (**0 deletions**), and their terminology is followed for consistency — e.g. the existing `Çizge` for chart is kept rather than switching to a synonym.
- Strings marked `translatable="false"` are excluded.
- Verified with `gradlew :app:processDebugResources` — builds successfully.
- Format specifiers (`%1$s`, `%d`, `%%`) were checked programmatically against the source strings and match one for one; no unescaped apostrophes or ampersands.

I'm a native Turkish speaker and run openScale daily with an OKOK/Chipsea broadcast scale, so I'll keep this locale up to date as new strings land.


ahmetabdullahgultekin:
Walked the app on an Android 16 emulator with the app locale set to Turkish, rather than trusting the build alone: Overview, Add user, the navigation drawer, Settings, Data management, and the Bluetooth permission screen. No truncation, no layout breakage, no strings landing in the wrong context.

Doing that surfaced three phrasings that were literal renderings rather than natural Turkish, so I've pushed a follow-up commit rewording them. One of those matters beyond style: `dialog_text_confirm_generic_verb` had the verb placeholder in a position that requires a Turkish case ending, which cannot be correct for an arbitrary verb — it now reads naturally whatever verb is substituted.

Unrelated to this PR, but noticed while testing: the activity-level dropdown renders `Sedentary` / `Mild` / … from the enum name (`UserDetailScreen.kt:417`), so it stays English in every locale, not just Turkish. Happy to send a separate small PR adding string resources for it if that's welcome — I didn't want to widen the scope of a translation PR.

oliexdev:
Thanks for your PR :+1:

PR #1456, opened 2026-08-11
Add the missing unit test for Wla25BodyComposition

ahmetabdullahgultekin:
`core/bluetooth/libs/` has a consistent convention: every lib ships with a matching test. Currently that is **17 libs, 16 tests** — `Wla25BodyComposition.kt` is the only one without. This adds it.

**Verified:** `gradlew :app:testDebugUnitTest --tests *Wla25*` → 14 tests, 0 failures.

### What it asserts and why

The class documents three rounding steps as *"load-bearing … each worth a tenth of a unit"* and pins the rounding mode with two exact values:

- `round1(26.35)` is **26.4** — half-to-even would give 26.3
- `round1(1.95)` is **1.9** — 1.95 has a float32 fraction of exactly 0.95, so the half-up test fails

Both are asserted through the public `compute()`, which rounds the incoming weight with `round1` before anything else and reports it in `Result.weightKg` — so `round1` stays private. Passing `heightCm = 100` makes BMI equal the weight, which keeps those cases readable.

The validity gate is covered slot by slot. Its asymmetry — slots 0 and 5 checked against `1.0`, the other eight against `100.0` — is what fixes the ordering of the ten impedances, so a test that only checked "all ≥ 100" would miss a reordering bug. There is an explicit case showing 20 Ω passes in a leading slot and fails in every other one.

The rest are on invariants the source states outright rather than on vendor-derived numbers:

- body fat clamped to 3–60 % (`BFR_MIN`/`BFR_MAX`)
- visceral fat truncated then clamped to 1–20
- fat-free mass = rounded weight − implied fat mass
- `muscleKg` consistent with `musclePercent`
- BMR from the Katch-McArdle constants

That distinction is deliberate: the KDoc says the constants were validated against the vendor library over 240 randomised inputs, but that evidence isn't in the repo, so I avoided baking in expected outputs I can't reproduce. These assertions stay valid even if the regression constants are ever retuned.

### Scope

One new file, no production code touched, no hardware needed. The only consumer is `RelaxmedicHandler.kt:200-215`, so nothing else can be affected.

### DIFF
--- a/android_app/app/src/test/java/com/health/openscale/core/bluetooth/libs/Wla25BodyCompositionTest.kt
+++ b/android_app/app/src/test/java/com/health/openscale/core/bluetooth/libs/Wla25BodyCompositionTest.kt
@@ -1,0 +1,177 @@
+/*
+ * openScale
+ * Copyright (C) 2026 openScale contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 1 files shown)


oliexdev:
Thanks for your PR :+1:

PR #1457, opened 2026-08-11
Fix: aggregation drops INT-typed measurements (heart rate vanishes from chart and statistics)

ahmetabdullahgultekin:
Aggregation averages the values correctly but writes `intValue = null`, while copying the measurement type unchanged — so the synthetic entry still reports `InputFieldType.INT`. Every consumer switches on `inputType` and reads `intValue`, so an INT-typed measurement reads as **absent** as soon as any aggregation level other than `NONE` is selected.

`MeasurementAggregationUseCase.kt:157-166` (before):

```kotlin
val syntheticVal = vwt.value.copy(
    id            = -1,
    measurementId = -1,
    floatValue    = avg,
    intValue      = null,   // type is still INT, so this is what consumers read
)
```

Downstream:

| consumer | effect |
|---|---|
| `MeasurementChart.kt:634-638, :673-677` | `yValue` resolves to null → the series disappears from the chart |
| `StatisticsScreen.kt:204-208` | min / max / avg / first / last / difference all become null |
| `MeasurementSmoothingUseCases.kt:119-123` | the type is dropped from smoothing entirely |

**User-visible symptom:** `HEART_RATE` is `InputFieldType.INT` (`Enums.kt:396`), as are INT custom types. Switching the Statistics screen to weekly aggregation renders the heart-rate card as `Min: -  Max: -  Avg: -` with an empty chart, while the same data at aggregation `NONE` renders correctly.

### The change

Populate `intValue` for INT-typed values, rounding rather than truncating:

```kotlin
intValue = if (vwt.type.inputType == InputFieldType.INT) avg.roundToInt() else null,
```

`floatValue` stays populated exactly as before, so consumers reading either field agree with each other. FLOAT-typed values are unaffected — there's a test asserting `intValue` stays null for them.

### Verification

Three tests added to `MeasurementAggregationUseCaseTest.kt`, which previously had no INT coverage at all:

- `aggregate_day_populatesIntValueForIntTypedMeasurements` — 60 and 70 aggregate to `intValue = 65`
- `aggregate_day_roundsIntValueRatherThanTruncating` — 60 and 65 give 63, not 62
- `aggregate_leavesIntValueNullForFloatTypedMeasurements` — guards the other direction

**Red/green confirmed, not assumed:** with the production change reverted and the tests kept, `gradlew :app:testDebugUnitTest --tests *AggregationUseCaseTest*` reports **7 tests, 2 failures**. With the change applied: **7 tests, 0 failures**.

`Fixtures.kt` gains an `intValueWithType` helper — the existing `valueWithType` only ever sets `floatValue`, so an INT fixture wasn't expressible. It's additive; nothing else changes.

No hardware needed for any of this.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/usecase/MeasurementAggregationUseCase.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/usecase/MeasurementAggregationUseCase.kt
@@ -32,6 +32,7 @@ import java.time.temporal.WeekFields
 import java.util.Locale
 import javax.inject.Inject
 import javax.inject.Singleton
+import kotlin.math.roundToInt
 
 /**
  * Use case that aggregates a list of [EnrichedMeasurement] into a smaller list of
@@ -160,7 +161,14 @@ class MeasurementAggregationUseCase @Inject constructor() {
                 id            = -1,
                 measurementId = -1,
                 floatValue    = avg,
-                intValue      = null,
+                // The synthetic value keeps the original type, so consumers still switch on
+                // InputFieldType.INT and read intValue. Leaving it null made INT-typed
+                // measurements read as absent everywhere downstream.
+                intValue      = if (vwt.type.inputType == InputFieldType.INT) {
+                    avg.roundToInt()
+                } else {
+                    null
+                },
             )
             vwt.copy(value = syntheticVal)
         }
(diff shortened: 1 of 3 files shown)


oliexdev:
Thanks for your PR :+1:

PR #1461, opened 2026-08-12
Support Terraillon X-LINE scale

pjestin:
Adding support for my scale, a Terraillon X-LINE. This model, from the Transtek family, is very similar to the Body Connect, which is already supported in the `BodyConnectHandler.kt`. A timestamp adaptation is needed though, as the scale returns a timestamp based on 2020-01-01, only when providing the body composition info.

I cannot test with the Body Connect scale, but the code should behave the same.

## Changes
* Add `0X-LINE*` to the supported devices for the `BodyConnectHandler.kt`
* Adapt the display name
* Make an adaptation on timestamps in `onBody`

## Tests
I tested the following:
* On the app, created a new user, matching the one that is already stored in the scale
* Paired the app with the scale
* Initiated a measurement with weight, water, fat, muscle and bone
* Made sure the history is retrieved
* Made sure the timstamps are correct

## btsnoop HCI log file
This was taken with the Manufacturer's app:
[btsnoop_hci.log](https://github.com/user-attachments/files/30969452/btsnoop_hci.log)

This one is more extensive, as it contains the pairing, etc:
[btsnoop_hci_pairing.log](https://github.com/user-attachments/files/30980172/btsnoop_hci_pairing.log)

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/BodyConnectHandler.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/BodyConnectHandler.kt
@@ -59,8 +59,8 @@ class BodyConnectHandler : ScaleDeviceHandler() {
 
     override fun supportFor(device: ScannedDeviceInfo): DeviceSupport? {
         val name = device.name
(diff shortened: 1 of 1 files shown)


oliexdev:
thanks for your PR :+1:

PR #1463, opened 2026-08-12
Add support for the Omron HBF-702T and its WLC-protocol siblings

ToastyTheBot:
# Add support for the Omron HBF-702T and its WLC-protocol siblings

## Summary

Adds a handler for the Omron HBF-702T body composition monitor, and with it six related models that
speak the same protocol: the HBF-222T in its Asia-Pacific, VIVA and BCM-500 trims, plus the
HBF-227T, HBF-228T and HBF-230T. They share one handler because they differ only in record size and
the granularity of a single field.

These are not streaming scales. Nothing useful arrives while you are standing on one — the weigh-in
is written into one of four on-device user slots, and the phone reads the stored records out
afterwards over a vendor GATT service.

The protocol was derived from OMRON connect's own device descriptors rather than from a packet
capture, and then checked against a physical HBF-702T: pairing, slot selection and record import all
work, and the imported values match the scale. The sibling models share the decoder but nobody has
run them yet. Details in Verification.

## The devices

Omron devices advertise as `BLEsmart_<group><model><mac>`, and those two 16-bit hex fields are the
device group and model ids that OMRON connect keys its own device table on. Scales live in group 1:

| Advertised prefix | Model | Record size |
|---|---|---|
| `BLEsmart_0001000C` | HBF-702T (Japan) | 48 B |
| `BLEsmart_00010011` | KRD-703T | 48 B |
| `BLEsmart_0001040C` | HBF-702T (Asia-Pacific) | 48 B |
| `BLEsmart_00010009` | HBF-227T | 32 B |
| `BLEsmart_0001000B` | HBF-228T | 32 B |
| `BLEsmart_0001000D` | HBF-230T | 32 B |
| `BLEsmart_00010408` | HBF-222T (Asia-Pacific) | 32 B |
| `BLEsmart_00010110` | BCM-500 | 32 B |
| `BLEsmart_00010208` | VIVA | 32 B |

Once the scale is bonded Android reports the GAP name instead of the advertised one, so
`supportFor()` also matches the plain model names (`HBF-702T`, `KRD-703T`, `HBF-228T` and so on).
Blood pressure monitors advertise in group 0 and are deliberately not matched — they store a
different record format.

## Protocol

Everything runs on service `ecbe3980-c9a2-11e1-b1bd-0002a5d5c51b`.

**Channels.** The protocol predates MTU negotiation, so a command is split into 16-byte pieces
across four write characteristics and the answer is reassembled from four notify characteristics in
the same order:

```
write   db5b55e0 / e0b8a060 / 0ae12b00 / 10e1ba60
notify  49123040 / 4d0bf320 / 5128ce60 / 560f1420
unlock  b305b680  (write + notify)
```

**Framing.** Byte 0 is the total frame length and the last byte is an XOR checksum over everything
before it, so an intact frame XORs to zero:

```
request   [len][cmd][flags][addrHi][addrLo][size][payload…][00][bcc]
response  [len][cmd|0x80][flags][addrHi][addrLo][size][payload…][result][bcc]
```

| Frame | Bytes |
|---|---|
| open session | `08 00 00 00 00 10 00 18` |
| read EEPROM | `08 01 00 <addrHi> <addrLo> <len≤0x20> 00 <bcc>` |
| close session | `08 0F 00 00 00 00 00 07` |

The flags byte is `0x00` for reads and `0xC0` for writes, which is why the top two address bits are
unusable.

**Unlock.** The device refuses EEPROM access until the phone proves it knows a 16-byte key, written
to `b305b680`. `01‖key` unlocks and is acknowledged with `81 00`. Programming a fresh key needs the
scale in pairing mode: `02` followed by sixteen zero bytes enters key-programming mode (`82 00`),
then `00‖key` stores it (`80 00`).

**Memory map.** Four user slots, each a ring buffer of 30 records. Every slot is allocated 31 record
slots' worth of address space, of which 30 are used:

| | Record | Slot bases |
|---|---|---|
| HBF-702T family | 48 B | `0x02C0` `0x0890` `0x0E60` `0x1430` |
| 32-byte siblings | 32 B | `0x02C0` `0x06A0` `0x0A80` `0x0E60` |

The settings block at `0x01A0` holds each slot's ring buffer write pointer in the low six bits of
bytes 0–3. The handler reads it to find the newest record per slot when it needs to show the user
what is on the scale.

**Fields.** A field reads `size` big-endian bytes at an offset into the record, then takes `bitSize`
bits starting at `startBit` counted from the LSB. Zero means "not measured" everywhere except the
weight, where it means the slot has never been written.

| Value | Offset | Bits | Scaling |
|---|---|---|---|
| weight | 26 | 4..15 | ×5, 10⁻² → kg (50 g steps) |
| body fat % | 2 | 6..15 | 10⁻¹ |
| visceral fat | 2 | 0..5 | half levels on the 702T, whole levels elsewhere |
| resting metabolism | 4 | 4..15 | kcal |
| skeletal muscle % | 6 | 6..15 | 10⁻¹ |
| BMI | 8 | 6..15 | 10⁻¹ |
| body age | 10 | 0..6 | years |
| timestamp | 7, 9, 11–13 | — | year+2000; hour straddles bytes 12/13 |

The 48-byte record extends this with segmental muscle and subcutaneous fat at offsets 28–40. Those
are decoded correctly but not published, since openScale has nowhere to put them.

Values map onto `ScaleMeasurement.weight`, `.fat`, `.muscle`, `.visceralFat` and `.bmr`. Visceral fat
goes in as a level rather than a percentage, matching what MiScale, Keep S3 and FitTrack Dara
already do with that field.

## Where the protocol came from

The APK ships the whole thing as data. `res/raw/device_category.zip` contains, per model, a
`DeviceConfig.sys` with the EEPROM memory map and a `VitalDataIndexes.json` describing every field
down to the bit, and `assets/Device/device_models.json` supplies the model ids above. Decompiling
was only needed twice: to confirm that `OGSCMemoryMapAnalyzer.h()` counts `startBit` from the least
significant bit, and to resolve Omron's measurement type codes (`0101`, `0103`, `010a`, …) to the
names in the table.

That last detail was worth pinning down. Day and hour both touch byte 12, and the two fields only
stop overlapping under an LSB-first reading, which independently confirms what the decompiled
extractor does.

## Design decisions

**Read-only apart from the unlock handshake.** OMRON connect resets the device's unread-record
counters and writes its clock after a sync; this handler does neither. New records are the ones
stamped later than the newest measurement openScale already holds for that user. Slower, but a
protocol mistake cannot corrupt what is stored on the scale. The counters and clock are writable and
documented in the same descriptors, so adding them later is not difficult if you want parity with
the vendor app.

**Slot binding through the existing CHOOSE_USER dialog.** On the first sync the handler reads the
newest record from each of the four slots and offers them as `Scale user 2 — 72.4 kg on 17/03/2024`,
then remembers the choice per app user in `DriverSettings`. The pairing key is stored the same way,
keyed by device address.

**Aborting on an unrecognised model.** The record layout differs per model and guessing it produces
plausible-looking nonsense, so an unresolvable device name ends the session with an error instead.

## Verification

Tested on a physical HBF-702T. Pairing, slot selection, readout and import all work, and the
imported weight, body fat, skeletal muscle, visceral fat and BMR match what the scale shows.

The pairing handshake was the part I expected to break, and it did not. That scale had already been
paired with OMRON connect, and openScale still programmed its own key once the Bluetooth button was
held into pairing mode — no factory reset, no unpairing from the vendor app first. So on this unit
at least, an existing OMRON connect pairing does not block a second key.

The slot dialog behaved too. It listed each of the four on-device slots with its newest measurement,
and after choosing one the link stayed up long enough to read the ring buffer in the same session.
That timing was my other worry, since the connection has to survive a user-facing dialog.

`./gradlew :app:testDebugUnitTest :app:assembleDebug` passes: 435 tests across 58 suites, no
failures or skips, 29 of them new.

The test vectors are records assembled field by field from Omron's bit-level description,
independently of the decoder under test. They deliberately exercise the fields that share a byte
with a neighbour: body fat sits beside visceral fat, BMR beside its classification, and the day/hour
pair straddles bytes 12 and 13. That is where an off-by-one in the bit positions would surface.
Frame tests assert the session-control frames against their literal byte sequences.

## Limitations

- Only the HBF-702T has been run against hardware. The HBF-222T, HBF-227T, HBF-228T, HBF-230T, VIVA
  and BCM-500 entries come from the same vendor descriptors and go through the same decoder, but
  nobody has tested them. Say the word and I will drop them from the model table so this ships with
  only what has been exercised.
- Segmental muscle and subcutaneous fat are decoded but dropped.
- Body age is decoded and dropped too — openScale has no field for it.
- The first sync holds the BLE link open across the CHOOSE_USER dialog, which worked on the tested
  unit. An 8-second watchdog aborts the session if the scale walks away mid-dialog; the user syncs
  again and the stored slot is used directly. If it turns out flaky on other units, the fallback is
  a slot picker in `DeviceConfigurationUi`.
- HBF-222T_LA appears in the vendor device configs but has no entry in the model table, so it is not
  claimed.

## Files

- `core/bluetooth/libs/OmronWlpFrame.kt` — framing, checksum, response reassembly
- `core/bluetooth/libs/OmronBodyCompositionLib.kt` — per-model geometry and the record decoder
- `core/bluetooth/scales/OmronWlcHandler.kt` — unlock/pair, transfer session, slot readout
- `core/bluetooth/ScaleFactory.kt` — registration
- `res/values/strings.xml` — 13 new strings
- three new test classes under `src/test/.../bluetooth/`


oliexdev:
Thanks for your PR :+1: I have only fold the Omron WLP framing into a single OmronLib file

PR #1464, opened 2026-08-13
Support for BMI calculations for Active Era BS-06 scales

NecroKote:
Continuation for #1091

PR contains full BIA calculation support for Active Era BS-06 scales reporting "used measurement algorithm No.7".
The calculation results matching displayed on the scales.

🥷 Also:
- switch to name-based scales matching (there are conflicting "generic" scales using same-ish protocol in 'MGBHandler')
- add support for HR reporting


oliexdev:
thanks for your PR :+1:

PR #1467, opened 2026-08-13
Add support for the Weight Gurus 0376 scale

psuedo-tty:
## What

Adds `WeightGurusA3Handler` for the Weight Gurus Bluetooth Smart Scale 0376 (Greater Goods /
DMD Brands), a Transtek A3 device on vendor service `0d005750-c36b-11e3-9c1a-0800200c9a66`. It
shares the command set with `BodyConnectHandler` (1BODY CONNECT, service `0x7892`) but encodes
its measurement frames differently, so it needs its own handler.

The scale advertises as `10376B` before pairing and `00376B<serial>` after.

## Measurements

**Working:** weight, impedance, body fat, body water, muscle, bone. Time sync, user profile
sync, and the history the scale queues while disconnected.

**Not implemented:** visceral fat and BMR. The append frame defines flags for both, but the
scale never sets them. Battery is the same case.

## Protocol notes

Three things the scale is strict about:

- It streams all eight user slots on connect, and the host must write nothing until the eighth
  frame lands. Answering earlier, for example on the first occupied slot, ends setup with `E1`.
- The `0x51` profile carries height as an IEEE-11073 SFLOAT in metres, not centimetres, so
  170 cm goes out as `0xD6A4`. The mantissa is signed and 12 bits wide, giving a ceiling of
  204.7 cm, above which the value would decode negative.
- `0x22` is followed by waiting. The scale closes the link itself once the pairing is committed
  to flash; dropping it from the host side interrupts that commit and also reports `E1`.

## Slots and body composition

Body composition is computed on the scale from the profile stored in whichever slot a reading
is attributed to, and slot selection is a physical button the host cannot influence. A slot
holding a stale height or age therefore yields wrong percentages for a correct weight. On this
device a slot carrying a stale profile (30 years, 160 cm) reported 24.6% body fat where the
corrected profile (40 years, 170 cm) reported 18.3% for the same person minutes later.

Writing `0x51` to that slot updates it, confirmed by the scale's own `0xC0` profile echo and by
the percentages moving. The handler therefore asks which of the eight slots the user is, listing
the names read off the scale, and writes the openScale profile there. The choice is persisted and
reused on later connections. Asking covers the cases guessing gets wrong: a new scale where every
slot is empty, and a second-hand one carrying a previous owner's profiles.

This reuses the existing `CHOOSE_USER` interaction, so no new dialog is introduced.

## Deduplication

`publishIfNew` matches on exact timestamp and leaves cross-session duplicates to the unique
`(userId, timestamp)` index on `Measurement`. A high-water mark against the newest stored
measurement loses data on this device: the scale keeps a separate queue per slot, drains only the
slot currently selected, and delivers newest first, so readings legitimately arrive older than
ones already stored. A delivered reading is dropped by the scale, so anything discarded is gone
for good.

## How it was verified

Built and tested against the physical scale.

- Weight decode matches the scale's display, checked against its own kg and lb readouts.
- Pairing completes without `E1`.
- Height encoding confirmed in both directions. The scale's `0xC0` echo uses the same
  SFLOAT-in-metres form it is sent.
- Backlog behaviour tested by weighing five times without syncing. All five arrived in one
  session, newest first.
- Per-slot queues tested by weighing alternately as two users and syncing under each. A sync
  drains only the selected slot and leaves the other slot's pending reading untouched.
- Profile correction tested by registering into a slot holding a stale profile and re-reading the
  `0xC0` echo.

27 unit tests cover the two IEEE-11073 float types, the pairing XOR, height encoding and
clamping, slot name decoding, and slot selection.

`./gradlew testDebugUnitTest` passes with 433 tests. Two failures in `BackupRestoreUseCasesTest`
are pre-existing on `master` and unrelated to this change.

## Notes

- Adds three strings for the slot list. No new measurement types and no schema change.
- The second commit adds a protocol reference under `docs/misc/`.


oliexdev:
Thanks for your PR :+1: I have only left out the md file.

PR #1468, opened 2026-08-14
1byone (classic): stop silently discarding valid weigh-ins

forrcaho:
The following was generated by Claude, as was the code in this PR. As a user, I can say that my scale was receiving a reading only about a third of the time before, and after this fix it works every time.

## Summary

My 1byone "Health Scale" needed about three attempts per reading. It turned out the connection was never the problem — **every** connection succeeded on the first try. The readings were arriving, being parsed, and then silently discarded by `OneByoneHandler`, with no log line and no user-visible error, so a working connection looked identical to a broken one.

There were two independent causes. Both are fixed here, with regression tests built from the captured frames.

## Cause 1: coalesced duplicate frames parsed as history

The scale sends its final measurement **twice**. Sometimes the copies arrive as two separate 11-byte notifications, and sometimes coalesced into one — the ATT payload caps at 20 bytes, so the buffer is a complete 11-byte frame followed by the first 9 bytes of its duplicate:

```
[00], [CF 92 0E 9C 27 1D 13 5F 00 00 B9 CF 92 0E 9C 27 1D 13 5F 00], [00 B9]
      └────────── complete frame ─────────┘└──── duplicate, truncated ────┘
```

`parseMeasurementFrame` decided history-vs-live on length alone:

```kotlin
val hasTimestamp = bytes.size >= 18
```

20 ≥ 18, so it read bytes 11–17 as a timestamp, got **year 53138, month 14, day 156, 39:29:19**, and the non-lenient `Calendar` threw into a `catch { return }` that dropped the reading.

The fix decides on the protocol's own XOR checksum at byte 10 plus a second `0xCF` marker at byte 11. A genuine history frame carries the year high byte (`0x07`) at byte 11 and measurement data at byte 10, so the two cases separate cleanly — history reads on the Eufy C1/P1/A1 models sharing this handler are unaffected.

## Cause 2: zero impedance threw away the weight

```kotlin
if (!impedancePresent || (isHistoric && !hasTimestamp)) return
```

The scale reports impedance `0` when it cannot run the bioimpedance measurement — socks, shoes, poor foot contact. The weight is still perfectly good, but the whole weigh-in was lost.

Now the weight is published and only the derived body composition is skipped, matching what `EufyP2Handler` and `EufyC20Handler` already do:

```kotlin
// EufyP2Handler.kt:110
if (reading.impedanceOhm > 0) impedance = reading.impedanceOhm.toDouble()
```

## Two smaller fixes in the same paths

- **Lock-status gate.** Live frames are now gated on byte 9 ∈ {`0x00`, `0x36`} — the vendor app's "locked" values — so in-progress readings are not recorded. This replaces the older `b9 != 1` heuristic with the protocol's actual rule.
- **Clock-ACK fallback.** `onConnected` gated both the history request *and* the "step on the scale" prompt on an `F1 00` ACK. This scale never sends one — the vendor app never even sends `F1` on this model — so the user got no feedback whatsoever. A 3 s fallback now prompts anyway. `waitAckClock` is deliberately left set so a late ACK still starts the history read.

## Verification

Confirmed on hardware — the same 103.00 kg weigh-in taken a minute apart:

| | Frame | Result |
|---|---|---|
| socks | `CF 00 00 3C 28 00 00 00 01 00 DA` | 103.00 kg, impedance 0 — **weight now saved** instead of lost |
| barefoot | `CF B6 0D 3C 28 B4 B5 99 01 00 F9 CF B6 …` | 103.00 kg, 351 Ω, full body composition — **coalesced frame now parsed** |

`OneByoneHandlerTest` covers both, plus the three earlier coalesced frames that were being dropped and a synthetic history frame confirming history detection still works. All frame vectors are verbatim from session logs, not synthesised.

Reverse-engineering was cross-checked against the decompiled vendor app ("New iWellness 4.0", `com.lefu.es.*`) — that's the source for the byte-9 lock values and for the fact that `F1` is never sent on this model.

## Deliberately not included

When a reading has no impedance, the user is still told nothing. A snackbar emitted at publish time is dismissed by `BleConnector`'s saved-measurement snackbar ~700 ms later, so it never really appears. Fixing that properly belongs in the save path rather than in a handler, so I've raised it separately rather than working around it here.

Two other things I looked at and left alone, both mentioned in case they're of interest:

- `MGBHandler.tryPublishStreaming` has the same "drop the weight if impedance never arrives" bug this PR fixes (`val impedanceOhm = streamingImpedanceOhm ?: return`), with no timeout fallback. I have no Dr. Trust hardware to test against, so I haven't touched it.
- The vendor app connects with `connectGatt(autoConnect = true)` and rescans every 10 s indefinitely, where openScale uses a direct connect after a 650 ms delay. That's a real robustness difference, but the logs show it costing nothing here, so it isn't part of this change.


oliexdev:
Two things before merging, both about the Eufy C1/P1/A1 that supportFor also matches and neither of us can test:

  1. The byte-9 whitelist. It narrows acceptance from "anything but 0x01" to only 0x00/0x36, and those values come from the 1byone vendor app — Eufy ships a different one. If a
  C1/P1/A1 uses another value, every live weigh-in there disappears silently: the same bug, relocated. supportFor already knows the model — could you apply the lock gate to
  the 1byone only and leave Eufy on the old b9 != 1?

  2. The rate limiter. Weight-only frames now reach DATE_TIME_THRESHOLD_MS and set lastSavedAt, where before they returned earlier. If a scale sends the settled weight first
  and the frame with impedance a second later, the second one is dropped and the user keeps a weight-only record. Starting the window only for frames carrying impedance would
  close that.

forrcaho:
After messing around with trying to get the weight with socks on to work correctly, we decided to drop that part and just have the reassembling frames part in this PR. As Claude puts it:

 Thanks — both points led somewhere I didn't expect. I've dropped the feature they were about rather than defending it.

## What changed

The second commit removes weight-only publishing entirely, along with the byte-9 lock gate that existed to make it safe. What's left is the coalescing fix and the clock-ACK fallback, both confirmed on hardware.

## Why

Your first point sent me to check what a weight-only measurement actually looks like once it's saved, and the answer is: not weight-only. `BleConnector` drops any value that isn't `> 0f`, so the six body-composition fields arrive at the database as *missing* rather than zero, and `applyValueInheritance` then fills them from the preceding measurement:

```
Carry-over: inherited 6 value(s) from measurement 8.
Carry-over: inherited 6 value(s) from measurement 10.
```

The second line is the problem — measurement 10 is itself a weight-only record, so the stale values propagate forward from weigh-in to weigh-in. The reading looks complete, and its body fat, water and bone mass are days old.

That's worse than the bug I set out to fix. A discarded reading is visibly absent; this one is quietly wrong.

The same capture also produced two records for one weigh-in: with socks on, the scale retries the bioimpedance measurement and re-sent the settled frame nine seconds later, well outside any plausible de-duplication window. So the hold-and-flush scheme I'd written for your second point wouldn't have collapsed it either.

## Where that leaves your two points

**The lock gate** is gone rather than scoped. Nothing guesses at Eufy's status byte values now, so the failure mode you identified can't occur. Eufy is byte-identical to master apart from the coalescing fix, which is protocol-level — it stops a duplicated frame being misread as history and doesn't depend on vendor semantics.

**The rate limiter** is untouched from master. It only became a problem because weight-only frames started reaching it; there are none now.

## What's left, and how well it's tested

- **Coalescing fix** — hardware-verified twice, including once on this exact build: a 20-byte coalesced frame parsed and published with full body composition (102.70 kg, 319 Ω).
- **Clock-ACK fallback** — hardware-verified; the prompt now appears on a scale that never sends `F1 00`.
- Eight unit tests over the framing decisions, all vectors verbatim from session logs.

Nothing speculative remains, and nothing in the change now depends on hardware I can't test.

## Follow-up

Weight-only measurements are worth supporting — my scale produces them whenever bioimpedance fails, and simply losing the weigh-in isn't great either. But the carry-over interaction isn't handler-specific: `EufyP2Handler`, `EufyC20Handler` and `VitafitVT701Handler`'s fallback all publish weight-only measurements today and will all be feeding it the same stale values. I'll raise that separately rather than work around it here.

oliexdev:
Thanks for digging into this one — the coalescing analysis is excellent, and the write-up made it
  easy to follow what the scale actually does. I've pushed a commit on top of your branch
  (`22639e4c`) rather than sending you round another review cycle. Here is what I changed and why.

  **I reverted the weight-only removal from `3b9d5d5b`.**

  The premise of that commit doesn't hold on our side: the carry-over in
  `MeasurementTransformationUseCase.applyValueInheritance` is deliberate, not a bug. Filling missing
  values from the previous measurement is what openScale is supposed to do when a scale reports
  nothing for them — it's the same behaviour the manual entry form relies on. So your first commit
  had it right, and publishing the weight while skipping only the body composition is exactly what
  `EufyP2Handler`, `EufyC20Handler` and `EtekcityFit8SHandler` already do.

  That means the byte-9 lock gate (`isFinalReading`, 0x00/0x36) is back too, along with the tests
  for it. Losing a valid weigh-in because the bioimpedance run failed is the worse outcome of the
  two.

  **I fixed a regression in `a3fcb9b3`.**

  `suppressStepOnPrompt()` ran for every `CF` frame, history frames included. On a scale that
  actually has stored history, the transfer therefore consumed the prompt, and the
  `promptForMeasurement()` call at the `F2 00` terminator became a no-op — so the user was never
  asked to step on. You can't see it on your unit, because it never sends the `F1` ACK and
  `historicMode` is never true for you, but it would hit the Eufy C1/P1/A1 models that share this
  handler. It's now gated:

      if (!historicMode) suppressStepOnPrompt()

  A history frame says nothing about whether someone is standing on the scale right now, and the
  prompt after the transfer is precisely what the user needs at that point.

  **I removed the rate limiter and replaced it for live frames.**

  You were right that a duplicate arriving nine seconds later is outside any plausible window — but
  the window was the wrong tool anyway. Note what it did to history reads:

      val nowMs = max(System.currentTimeMillis(), whenCal.timeInMillis)

  A history entry's timestamp is in the past, so `max()` always returned "now", and a history burst
  arrives within milliseconds. Everything after the first entry fell into the 3-second window and
  was dropped. That has been broken for a while, independent of your PR.

  `DATE_TIME_THRESHOLD_MS` and `lastSavedAt` are gone. In their place:

  - live frames: one measurement per connection (`publishedLive`, reset in `onDisconnected()`),
    which catches both the immediate duplicate and the nine-second repeat regardless of timing;
  - history entries: no limit at all — each is sent once and carries its own timestamp.

  `onDisconnected()` now also resets `historyCount`.

  I did consider holding a weight-only reading briefly in case a frame with impedance still follows,
  and decided against it: neither of your captures shows that sequence, and the scale settles once,
  so its bioimpedance run has either succeeded by then or it hasn't. If a log ever turns up showing
  "weight-only, then the same weigh-in with impedance", that's the moment to add a grace period —
  with a measured delay rather than a guessed one.

  **Could you re-test on your hardware?** The build is green and the unit tests pass here, but the
  paths that matter are ones I have no scale for. Specifically:

  1. barefoot — full body composition, one record, as before;
  2. in socks — the weight is saved again, without body composition, and only **one** record for the
     weigh-in even though the scale re-sends the settled frame;
  3. an in-progress frame (stepping on, still settling) is not recorded;
  4. if your scale can be made to hold history, that the "step on the scale" prompt still appears
     after the transfer.

  Once that checks out, please also bring the PR description in line with the final state — it still
  describes the two-cause version from your first commit, which is what made the branch hard to read
  at first. Then I'll rebase and merge.

jremy42:
Tested this PR on a real 1byone "Health Scale" and the fix works for me.

### Test device
- Scale: 1byone "Health Scale"
- Firmware: V2.4
- Phone: OnePlus / Android 16
- openScale: OSS debug build from this PR

Before applying the PR, openScale successfully connected to the scale and subscribed to FFF4, but no measurement was stored.

My scale sends the settled measurement twice. For example, I captured:

CF 3E 12 78 1E 5F 55 27 00 00 A8
CF 3E 12 78 1E 5F 55 27 00 00 A8

The two 11-byte frames were delivered as a 20-byte notification followed by a 2-byte notification.

### Results with this PR
- Barefoot measurement: works, weight is stored and body composition is calculated.
- Measurement with socks / no impedance: works, weight is still stored without a new body-composition measurement.
- I did not observe duplicate measurements being stored.

I did encounter two Kotlin compilation errors on the current PR branch:

1. `Unresolved reference 'impedance'`
2. `No 'set' operator method providing array access`

in `OneByoneHandler.kt`.

I fixed them locally by changing:

    if (impedancePresent) impedance = impedanceOhm.toDouble()

to:

    if (impedancePresent) {
        this[MeasurementType.IMPEDANCE] = Ohm(impedanceOhm)
    }

and by removing this line from the no-impedance `else` block:

    this[MeasurementType.IMPEDANCE] = Ohm(impedanceOhm.toFloat())

After these changes, the project builds successfully and the scale works correctly in both tests.

I pushed the compilation fix to my fork/branch if you'd like to cherry-pick it.

c0389b78 - Fix OneByoneHandler impedance compilation errors

oliexdev:
Thanks @jremy42 — the second device report is exactly what this needed, and your
  compile fix was right. I've pushed it along with two changes on top.

  **Scoped the 1byone rules to the 1byone.** The byte-9 lock whitelist
  (`{0x00, 0x36}`) and weight-only publishing both come from the 1byone vendor app
  and 1byone hardware. Applied to the whole family, an Eufy C1/P1/A1 whose firmware
  uses a third status value would lose every live weigh-in silently — the same bug,
  relocated. The model is now resolved at connect time, and the Eufy models keep the
  pre-existing rule (`b9 != 1` plus impedance required). They still get the
  coalescing fix, the history rate-limiter fix and the state reset, all protocol-level.

  **Per-connection state is reset in `onConnected()`** as well — handler instances
  live for the lifetime of the app.

  @forrcaho: your analysis of the coalescing was the hard part here and it holds up
  completely. I've taken the remaining items myself rather than leave the PR sitting.

PR #1472, opened 2026-08-15
Added support for Etekcity Fit 8S Scale.

basjam:
Model name: Etekcity Fit 8S.
Features working:
- Live Weight
- Body Composition (impedance)

Notes about the implementation:
- The reverse engineering was already done by ronnnnnnnnnnnnn in  [his python project](https://github.com/ronnnnnnnnnnnnn/etekcity_esf551_ble). This implementation takes information from the work in that project.
- The device operates in Broadcast mode only. No GATT connection.
- I used the built-in `StandardImpedanceLib` for derived information only when impedance is available.
- The device does not publish its name in advertisement. The only way to filter it is by Manufacturer ID in advertisement and Scale service UUID. As such:
    - The handler for `Etekcity Fit 8S` may interfere with advertisements from the same manufacturer. I made sure to put the handler below other `Etekcity` handlers in `modernKotlinHandlers`.
    - The device shows with `Unknown` name in Settings-Bluetooth screen. I'm not sure how to fix that. _(screenshot below)_
- Documentation and tests have not been updated. _(Let me know if they are required)_
- The device has a button to switch display units. It reports its current display unit in advertisement data. However, it always reports the actual weight in grams. There is no way to switch the on-device display units remotely. The code currently ignores the display unit flag.

Instead of the `btsnoop_hci.log`, I've attached a filtered WireShark capture of a couple of measurements. [scale-capture-72..zip](https://github.com/user-attachments/files/31092117/scale-capture-72.zip)  I filtered by BD_ADDR. Let me know if this is sufficient.

openScale-unknown-name

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/ScaleFactory.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/ScaleFactory.kt
@@ -39,6 +39,7 @@ import com.health.openscale.core.bluetooth.scales.EufyC20Handler
 import com.health.openscale.core.bluetooth.scales.EufyP2Handler
 import com.health.openscale.core.bluetooth.scales.EbelterBodyFatB2Handler
 import com.health.openscale.core.bluetooth.scales.EtekcityESF551Handler
+import com.health.openscale.core.bluetooth.scales.EtekcityFit8SHandler
 import com.health.openscale.core.bluetooth.scales.GattScaleAdapter
 import com.health.openscale.core.bluetooth.scales.HesleyHandler
 import com.health.openscale.core.bluetooth.scales.HoffenBbs8107Handler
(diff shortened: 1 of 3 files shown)


oliexdev:
Thanks a lot, I checked the matcher against the whole handler registry and made two small changes on top of your branch:
  
  1. onAdvertisement now looks up the 0x06D0 record directly instead of taking the first manufacturer record of at least 20 bytes — otherwise a second record  of that size in the same advertisement would be decoded as a weight. The length check moved into a PAYLOAD_SIZE constant.
  2. Dropped the redundant as? SparseArray<*> cast; scanRecord.manufacturerSpecificData is already a SparseArray<ByteArray>.

  On interfering with other scales: there is exactly one overlap, and your placement already handles it. ScaleupHandler claims any company id whose low byte is 0xD0/0xE0, and 0x06D0 ends in 0xD0, so both handlers answer for a Fit 8S — yours wins by standing earlier. The other direction is safe because you also require service 0xFFD0. I added tests for that to ScaleFactoryTest so a future reorder can't break it silently.

 The Unknown name is correct the supported device name is correct under the MAC address.

  Could you re-test the changed handler on your scale once more (live weight plus a measurement with impedance)? I only verified it against synthetic  advertisements.

basjam:
Hey @oliexdev, I tested the changes and it works. 

I tightened the check for payload size to prevent future collisions. All messages from scale are exact 20 bytes sized.

I tested again after tighter payload size check, and it works.

oliexdev:
Thanks for testing

PR #1476, opened 2026-08-16
feat: Add support for the ProfiCare PC-PW 3008 BT

tiagoagueda:
## Scale

**ProfiCare PC-PW 3008 BT** — body-analysis scale sold mainly in the German market. Vendor app: "Dr.Curve+" (`com.yilai.DrCurvePlus`).

## What this changes

The PC-PW 3008 BT runs the same Chipsea "WeChat scale" firmware as the Hoffen BBS-8107, so `HoffenBbs8107Handler` already decoded it correctly — the only thing missing was recognition. The handler's exact-name check becomes a small advertised-name → display-name map, so the scale is reported under its own product name instead of the Hoffen's.

**No protocol/parsing code changed.**

## Feature status

Verified on real hardware (Samsung Galaxy A55), matching the vendor app's readings:

- Weight
- Body composition: fat, water, muscle, bone, visceral fat

Capability flags are unchanged from the existing handler (`BODY_COMPOSITION`, `USER_SYNC`, `UNIT_CONFIG`).

## Evidence

The scale advertises:

```
02 01 06                    Flags
03 02 E7 FE                 16-bit service UUID 0xFEE7   (WeChat service)
0E 09 "PC-PW 3008 BT"       Complete Local Name
09 FF 31 01 6A2B0000A6BC    Manufacturer data, company 0x0131
```

I don't have a `btsnoop_hci.log` for this scale. Instead the protocol was confirmed by decompiling the vendor app: its `onWeixinFatScale` routine decodes exactly the offsets `parseFinalMeasurement` already uses — weight LE16 @3, fat @6, water @8, muscle @10, bone @14, visceral fat @17 — and its command set (`0xFA 0x85` user, `0xFA 0x83` unit, `0xFA 0x82` done; responses `0x01`/`0x02`/`0x03`) matches what the handler sends. The hardware test above then confirmed it end to end.

Happy to capture an HCI log if you'd like it for the record.

## Why matched by name

Matched on the advertised name rather than on service `0xFEE7`. That service is the generic WeChat BLE service used by plenty of unrelated hardware, so claiming it outright would list non-scales as supported devices. `HoffenBbs8107HandlerTest` covers that case explicitly.

## Tests

- `ScaleCatalog.kt` — one fixture line, so the registry regression test and the generated wiki row both follow automatically
- `scale_catalog_remarks.txt` — remark noting the shared firmware, since the table now has two rows served by one handler
- `HoffenBbs8107HandlerTest` — new; 6 tests covering both brands, case/whitespace tolerance, and negative matches (other 0xFEE7 devices, near-miss model names)

`./gradlew testDebugUnitTest compileDebugAndroidTestKotlin assembleDebug` passes, with two exceptions unrelated to this change: `BackupRestoreUseCasesTest.restoreDatabase_withValidBackupZip_restoresPreviousSnapshot` and `restoreDatabase_withLegacySingleFile_restoresAndMigrates` fail identically on unmodified `master` (61cb196) in my environment (Windows, JDK 21), so they look environmental rather than caused by this PR.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/HoffenBbs8107Handler.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/HoffenBbs8107Handler.kt
@@ -30,12 +30,17 @@ import java.util.UUID
 import kotlin.math.max
 
 /**
- * Handler for the Hoffen BBS-8107 scale.
+ * Handler for the Hoffen BBS-8107 and its rebrands.
  *
  * Protocol (single custom characteristic):
  *  - Service 0xFFB0, Characteristic 0xFFB2 (notify + write with response)
  *  - Packets start with 0xFA, then `cmdOrResp`, `len`, `payload`, `checksum`
  *  - Checksum = XOR of bytes 1..(n-2); some final measurement frames deliver checksum in a follow-up notify.
+ *
+ * This is the Chipsea "WeChat scale" protocol: the same firmware is sold under several brands,
+ * all advertising the WeChat service 0xFEE7 alongside the 0xFFB0 service they actually talk on.
+ * Verified against the stock "Dr.Curve+" app (`com.yilai.DrCurvePlus`), whose `onWeixinFatScale`
+ * routine decodes exactly the offsets [parseFinalMeasurement] uses.
  */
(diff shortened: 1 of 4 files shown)


oliexdev:
Thanks for your PR :+1:

PR #1480, opened 2026-08-20
Improvements to the X-LINE

pjestin:
The current implementation of the Body Connect / X-LINE has a basic user slot choice algorithm, where we essentially always pick the first non-empty slot. This has the downside of not supporting either a first-time setup, or multiple users.

What the manufacturer's app does in this case is the following:
* Receive all slot statuses from the device (8 in total) with the corresponding name
* Open a dialog for the user to pick a slot
* Send an ACK to the device with the slot number and the new name

It turns out this exact flow was recently implemented by @psuedo-tty in https://github.com/oliexdev/openScale/pull/1467 for another scale, the Weight Gurus 0376.
This PR borrows from it with an adaptation to incorporate this flow into the BodyScaleHandler class.

The only adaptation I had to make from WeightGurusA3Handler is that the maximum size of names is 16 instead of 18.

## Changes
* Change the `onSlotStatus` implementation to allow the user to pick a slot
* Copy the related flow from WeightGurusA3Handler, changing the max name size to 16
* Slight improvements on the profile writing, adding activity level and initial weight
* Add more logging
* Add a comprehensive set of unit tests

## Tests
On top of the unit tests I added, I tested the pairing flow with my X-LINE scale:
* The dialog appears with the slot and the associated names
* One of the choices matches the current user
* When choosing a slot, the device correctly registers the slot number and name
* The name is truncated when longer than 16 characters (note that the manufacturer's app fails at this point)

## Open questions
* This flow is likely to also work on the Body Connect but we can't be sure about it, since I cannot test with it. Maybe @LucasDumont could give it a try of he has time
* Should the slot choice flow be extracted to a common helper class?

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/BodyConnectHandler.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/BodyConnectHandler.kt
@@ -18,11 +18,14 @@
 package com.health.openscale.core.bluetooth.scales
 
 import com.health.openscale.R
+import com.health.openscale.core.bluetooth.BluetoothEvent.UserInteractionType
 import com.health.openscale.core.bluetooth.data.ScaleMeasurement
 import com.health.openscale.core.bluetooth.data.ScaleUser
+import com.health.openscale.core.data.ActivityLevel
 import com.health.openscale.core.service.ScannedDeviceInfo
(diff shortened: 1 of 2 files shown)


oliexdev:
Thanks for your PR.
No we don't need a common helper class for the slot choice extraction.

PR #1484, opened 2026-08-23
Add scoped Beurer BF1000 support

Lidara:
Claim the BF1000 in the standard Beurer/Sanitas handler and keep persisted measurements limited to the existing openScale fields.

Decode the BF1000 private segmental packets for logging/documentation only, with JVM test vectors from captured data.

Guard the BF1000-specific mixed user-list handling behind the BF1000 model path so existing Beurer/Sanitas models keep their previous parser behavior.


oliexdev:
Thanks for your PR and adjusting the source code

PR #1429, opened 2026-07-22
 Add selectable Scientific body-composition algorithm to Mi Scale

dkuku:
## Overview
Mi Scale v1/v2 previously derived body composition **only** from the reverse-engineered Mi Fit algorithm (`MiScaleLib`). This adds a per-scale setting letting users choose between:

- **Xiaomi (original app)** — the existing reverse-engineered path, reproduces Mi Fit / Zepp Life.
- **Scientific** — a new estimator ported from the [bodymiscale](https://github.com/dckiller51/bodymiscale) Home Assistant integration (GPL-3.0).

The picker lives in the scale's device configuration screen; the selected algorithm applies to new measurements.

## Changes
- **`BodyMiScaleLib`** — mono-frequency estimator. SCIENCE mode chains a hardware-calibrated LBM with peer-reviewed formulas (Siri 1956, Pace & Rathbun 1945, Wang 1999, Schofield/WHO) and additionally reports **protein** and **BMR**. The formulas are internally consistent (chained: fat←LBM, water/protein←fat/LBM, muscle←fat+bone).
- **`BodyMiScaleLibTest`** — locks SCIENCE mode to **byte-for-byte parity** with the upstream reference project via a real captured measurement, plus impedance-direction and Xiaomi-parity sanity checks.
- **`MiScaleHandler`** — radio picker to select the algorithm; measurement path branches on the setting.
- **`ScaleDeviceHandler`** — extracts a reusable `SettingRadioGroup` composable (settings-backed radio group) for device-configuration UIs.
- **`CREDITS`** — adds bodymiscale attribution.

## Testing
Full unit-test suite: **331 passed, 0 failed** (`:app:testDebugUnitTest`). Debug build installed and verified on device.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1488, opened 2026-08-25
Fix/female body composition

dkuku:
## Summary
- correct female lean-body-mass estimates using the bodymiscale calibration factor
- apply a sex-specific minimum body-fat percentage
- add regression coverage for female LBM and body-fat floors

## Scope
This correction applies only to Xiaomi scales when using the Scientific body-composition mode.

## Changelog reference
- https://github.com/dckiller51/bodymiscale/pull/415


oliexdev:
Thanks for your PR.  Merged and resolved the conflicts myself. Please rebase onto master before opening a PR next time — the branch was based on fe77bcbe, so BodyMiScaleLib.kt and the mi_algorithm_* strings showed up as new although #1429 had already added them.

PR #1492, opened 2026-08-26
Huawei Scale 3: add realtime body composition support

Stefanos1000:
### **## Summary**

_Adds realtime Huawei Scale 3 support through the existing Hagrid WSP handler._

The implementation includes:

- Huawei Hagrid WSP authentication using user-supplied CAK/C1/C2 credentials
- Profile synchronization
- Realtime weight measurements
- Huawei Scale 3 body composition calculation
- Body fat
- Water
- Muscle
- Lean body mass
- Bone mass
- Visceral fat
- BMR
- Protein
- Impedance
- Heart rate when available
- Multiple measurements within the same BLE connection
- Immediate publishing of completed realtime measurements
- Correct handling of incomplete body-composition measurements without carrying over stale values from the previous measurement

## Measurement behavior

Huawei Scale 3 reports different measurement stages independently.

The implementation therefore supports:

- weight-only measurements when the user steps off before impedance measurement completes
- body-composition measurements without heart rate when impedance completes but heart-rate measurement does not
- full measurements when all stages complete

Realtime measurements are published immediately when received. The scale measurement-status response is then used to complete the hardware measurement cycle and re-arm the same BLE connection for another measurement.

## Body composition

Huawei Scale 3 body-composition values are calculated from the low-frequency impedance measurement together with the user profile.

The calculation is implemented directly in Kotlin and does not depend on Huawei native libraries.

## Huawei Hagrid credentials

The Hagrid protocol requires three authentication values: `CAK`, `C1` and `C2`.

These credentials are intentionally not included or hardcoded in this contribution.

I currently do not have enough evidence to determine whether these values are device-specific, model-specific, or shared between Huawei Scale 3 devices. For that reason, credentials obtained from my own Huawei setup are not being published.

openScale therefore allows the user to provide these values locally.

### Obtaining the credentials

A short outline of the process used during development is:

1. Pair and configure the Huawei Scale 3 with Huawei Health normally.
2. Capture/debug the Hagrid communication or inspect the Huawei Health installation while the scale is configured.
3. Locate the three Hagrid authentication values:
   - `CAK`
   - `C1`
   - `C2`
4. Enter those values in the Huawei Hagrid settings in openScale.
5. openScale then uses them locally for Hagrid WSP authentication.

The credentials should not be posted in public logs or bug reports.

More detailed extraction documentation can be added separately if this implementation is accepted.

## Tests

Added and updated tests for:

- Huawei Scale 3 Hagrid WSP behavior
- realtime measurement publishing
- Scale 3-specific gender handling
- multi-cycle measurement behavior
- Huawei Scale 3 body-composition calculations
- incomplete BIA measurements
- prevention of stale body-composition and heart-rate value inheritance

The focused unit tests pass successfully, and the implementation was also tested with a physical Huawei Scale 3.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/HuaweiHagridWspHandler.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/scales/HuaweiHagridWspHandler.kt
@@ -119,11 +119,12 @@ class HuaweiHagridWspHandler(
     private var measurementStatusPollJob: Job? = null
     private var measurementStatusReady = false
     private var postStatusMeasurementReadsStarted = false
-    private var realtimeFinalPublished = false
-    // Scale 3: cache the last valid realtime measurement so it can be published
-    // on status=0 if it arrived before the status notification (the common case).
-    private var lastValidScale3Realtime: HuaweiHagridWspLib.HagridWeightMeasurement? = null
-    private var lastKnownUser: ScaleUser? = null
+    private var pendingRealtimeMeasurement: HuaweiHagridWspLib.HagridWeightMeasurement? = null
+    private var pendingRealtimeUser: ScaleUser? = null
+    private var measurementCycleCompleted = false
+    private var measurementPublished = false
+    private var rearmPollingJob: Job? = null
+    private var rearmPollOutstanding = false
 
     @Composable
     override fun DeviceConfigurationUi() {
@@ -237,6 +238,7 @@ class HuaweiHagridWspHandler(
         pendingRandB = null
         session = null
         stopMeasurementStatusPolling()
+        stopRearmPolling()
         configuredSecrets = loadConfiguredSecrets()
         handshakeState = if (configuredSecrets == null) {
             HandshakeState.PROBE_ONLY
@@ -252,10 +254,11 @@ class HuaweiHagridWspHandler(
         historyReadCount = 0
         publishedHistoryKeys.clear()
         postStatusMeasurementReadsStarted = false
+        pendingRealtimeMeasurement = null
+        pendingRealtimeUser = null
+        measurementCycleCompleted = false
+        measurementPublished = false
         measurementStatusReady = false
-        realtimeFinalPublished = false
-        lastValidScale3Realtime = null
-        lastKnownUser = null
 
         setNotifyIfPresent(svcUserData, chrRequestAuth)
         setNotifyIfPresent(svcUserData, chrAuthToken)
@@ -308,6 +311,7 @@ class HuaweiHagridWspHandler(
     override fun onDisconnected() {
         logI("Huawei Hagrid disconnected")
         stopMeasurementStatusPolling()
+        stopRearmPolling()
         accumulators.clear()
         pendingRandA = null
         pendingRandB = null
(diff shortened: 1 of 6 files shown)


oliexdev:
1. Please remove the MeasurementTransformationUseCase change.

  Value inheritance is intentional. applyValueInheritance is the app-wide carry-over rule and also feeds the manual entry form (MeasurementDetailScreen.kt:201). Many users maintain body fat, water, muscle and heart rate by hand — carrying the previous value forward keeps the series continuous instead of tearing a gap into graph and statistics. The removed heartRateId assertion in MeasurementValueInheritanceTest is the specification.

  For an interrupted BIA weigh-in the handler already does the right thing: calculate() returns null for missing impedance, so weight-only is published. 

  2. Fix the impedance handling in publishHagridMeasurement.

  - representativeLowOhm is already normalised by hagridImpedanceOhm() (HuaweiHagridWspLib.kt:139); the extra /10 makes impedanceLow unit-dependent on whether the composition succeeded, and shifts new Scale 3 values by factor 10 against existing data.
  - impedance = composition?.impedanceOhm ?: high overwrites the high-frequency value with the low-frequency one.
  - Add a handler-level test from raw frame to published impedance/impedanceLow; HuaweiScale3BodyCompositionTest bypasses this path.

  3. Smaller points.

  - Assert the computed fat instead of just deleting the fat == 18.5f assertion.
  - startRearmPolling: rearmPollOutstanding is only cleared on a status notification — if one is lost, the loop spins and the next measurement never starts. Needs a timeout.
  - Realtime now uses Date() instead of parsed.timestamp — add a comment why.

Stefanos1000:
Thanks for the detailed review. I’ve pushed a follow-up commit addressing the points you raised:

- Restored the original app-wide value inheritance behavior and its test.
- Kept the published high/low impedance values independent from the Scale 3 body-composition calculation.
- Added handler-level coverage from the realtime payload through to the published impedance / impedanceLow values.
- Restored a body-fat assertion using the calculated Scale 3 value.
- Added recovery for a lost rearm status notification so polling cannot remain permanently outstanding.
- Added a comment explaining why realtime measurements use the local receive time.

The focused HuaweiHagridWspHandlerTest and HuaweiScale3BodyCompositionTest suites pass successfully.

I also smoke-tested the updated build on a real Huawei Scale 3. Full BIA measurements, interrupted BIA measurements, and repeated measurements within the same BLE session behaved as expected. The published impedance fields now remain separate from the body-composition model input as intended.

Thanks again for pointing these out.

oliexdev:
Thanks — all five points check out. I also verified the other Hagrid families are unaffected:
  realtime publishing, the composition model and `startRearmPolling()` are all behind `isScale3`,
  and the gender encoding condition is unchanged.

  Two small changes on top before merging:

  **`impedanceLow` now stores the value the model was fed.** The same raw sample ended up as two
  numbers — `0x0DAC` stored as 3500 Ω, but 350 Ω passed to the composition. A stored impedance that
  doesn't match what the body composition was derived from can't be interpreted later, and 3500 Ω
  isn't a possible whole-body reading, so `hagridImpedanceOhm()`'s "below 4000 is already ohms" band
  is wrong for this family:

      val publishedLow = if (isScale3) scale3ModelImpedanceOhm ?: low else low

  Scale 3 only; the other families keep `representativeLowOhm` unchanged.

  **GPL headers.** `HuaweiScale3BodyComposition.kt` was cut off after the third paragraph and its
  test had none. Both are complete now.

  One open question: `impedance` (high-frequency) still uses the same heuristic. If your Scale 3
  reports that channel in 0.1-ohm units too, it's off by ten as well — but nothing cross-checks it
  and the test payload values are synthetic, so I didn't guess. A capture with raw high-frequency
  samples would settle it.

PR #1495, opened 2026-08-28
feat: Add support for AiLink broadcast body fat scales

manueltarouca:
## Scale

**AiLink / eLink broadcast body-fat scale family** — driven by the vendor "AiLink" app
(`com.pingwang.elink`) and sold under many private labels. The unit this was developed against
advertises as `EL1` and is retailed in Portugal as "BALANÇA BODY SMART C/ APP" (importer MCH S.A.,
model ZF3 / 7632673).

## Feature status

| Feature | Status |
|---|---|
| Weight | ✅ working, verified on hardware |
| Impedance | ✅ working, verified on hardware |
| Body composition | ✅ derived via `StandardImpedanceLib` |
| Live weight stream | ✅ consumed (only the final frame is recorded) |
| Time sync / user sync / history | ❌ n/a — the device is non-connectable |

Tested end-to-end on a Samsung SM-S928B (Android 15).

## How the scale works

It is **non-connectable**. Every attempt to open a GATT link times out; the scale instead repeats
the complete measurement inside an encrypted manufacturer record, so the handler runs in
`BROADCAST_ONLY` mode.

AiLink abuses the manufacturer-data field: the two bytes Android parses as a *company id* are
really the vendor's CID and VID, so `ScanRecord.getManufacturerSpecificData()` returns the record
already stripped of them.

```
service 0000F0A0

companyId = VID << 8 | CID
  [0]      PID
  [1..6]   MAC address, little-endian
  [7]      checksum: sum of [8..17] & 0xFF
  [8..17]  10-byte payload, TEA-encrypted
```

The payload is plain 32-round **TEA** over *only the first 8 bytes* (bytes 8–9 are passed through
untouched — the port keeps that quirk so frames match the vendor app byte for byte). The 128-bit
key is built inline from the device ids:

```
k0 = pid + 0x41493000      k2 = cid + 0x31783100
k1 = vid + 0x4C327900      k3 = 0x306C6531
```

Decrypted:

```
  [0]      rolling sequence counter
  [1]      status
  [2]      flags: bits 3..5 = weight unit, bits 1..2 = decimal places
  [3..4]   weight, big-endian; bit 7 of [3] = negative flag
  [5..6]   impedance in ohms, big-endian
  [7]      body-composition algorithm id
  [8..9]   not decrypted, meaning unverified — deliberately not decoded
```

Status values, named after the branches in the vendor app's own UI:
`0` measuring · `1` analysing · `2` impedance ready · `3` impedance failed · `4` data mode ·
`0xFF` complete.

Only the `0xFF` frame is published, once per session: the scale streams live weight while the user
settles, then latches the final frame and repeats it for as long as it stays awake.

## How it was reverse-engineered

No `btsnoop_hci.log` for this one, sorry — the scale is non-connectable, so there is no GATT
conversation to capture. The work was done from:

1. **Live BLE captures from a macOS host** (`bleak`), decoding advertisements off the air.
2. **The vendor APK**, decompiled with `jadx`: `BleValueBean.init` (advertisement framing),
   `MainModel.BroadCast` (checksum + payload slice), `BroadcastScaleDeviceData.dataCheck`
   (payload layout) and `BroadCastBodyFatMainActivity.getWeightData` (status semantics).
3. **The native library** `libAILinkBle-lib.so`, which ships with its C symbols intact
   (`encrypt_tea`, `decrypt_tea`, `get_ailink_key`, `decrypt_ailink`). Disassembling
   `Java_com_pinwang_ailinkble_AiLinkPwdUtil_decryptBroadcast` gave the key derivation above
   directly — nothing here is guessed.

## Capture evidence

Weight is redacted as `8X.X` and the MAC as `XX:XX:XX:XX:XX:XX` throughout; the encrypted payload
bytes are shown as a placeholder because this PR documents the key, which would otherwise make
them trivially decodable back to a real person's weight. Everything load-bearing — framing,
checksum, the status machine and the impedance behaviour — is intact.

Raw advertisement as received on Android:

```
03 03 A0 F0                          AD: complete 16-bit service UUIDs -> F0A0
04 09 45 4C 31                       AD: complete local name -> "EL1"
15 FF 01 03 01 XX XX XX XX XX XX BA <10 bytes encrypted>
      │  │  │  └── MAC, little-endian (redacted)
      │  │  └───── PID = 0x01
      │  └──────── CID = 0x01, VID = 0x03
      └─────────── AD type 0xFF, manufacturer specific data, len 0x15
```

Checksum check on one real frame: `sum(payload) & 0xFF == 0x5C`, matching byte [7]. ✅

A full measurement cycle, decoded live (barefoot):

```
seq= 27 status=  0 (measuring )  weight=8X.X kg  impedance=500
seq= 34 status=  0 (measuring )  weight=8X.X kg  impedance=500
seq= 35 status=  1 (analysing )  weight=8X.X kg  impedance=500
seq= 48 status=  1 (analysing )  weight=8X.X kg  impedance=500
seq= 49 status=  2 (imp-ready )  weight=8X.X kg  impedance=500
seq= 50 status=255 (COMPLETE  )  weight=8X.X kg  impedance=500
```

## The one subtlety: impedance is stale until the BIA phase ends

Worth calling out, because it is easy to misread. The impedance field **repeats the previous
measurement's value** until the scale's BIA phase finishes. Watching a single session it therefore
looks like a hardcoded constant — it is already `500` during `status 0`, before any measurement
has been taken.

It is not a constant. Controlled experiment on the same scale, minutes apart:

```
barefoot                          through footwear (BIA circuit broken)
seq= 49 status=2 imp-ready  500   seq= 75 status=  0 (measuring )  impedance=500   <- stale
seq= 50 status=255 COMPLETE 500   seq= 77 status=  3 (imp-FAILED)  impedance=  0   <- BIA ran, failed
                                  seq= 78 status=255 (COMPLETE  )  impedance=  0
```

The scale ran the same BIA phase, detected the broken circuit and reported failure. A device
emitting a hardcoded value could not distinguish those two cases.

Two rules follow, both implemented and documented in the handler:

- only ever read impedance from the `status 0xFF` frame;
- treat `0` as "the scale's BIA pass failed" — publish the weight but no body composition, rather
  than feeding a bogus `0` into `StandardImpedanceLib`.

## Notes on the implementation

- Protocol parsing lives in `AiLinkLib`, free of Android dependencies; `AiLinkBroadcastHandler` is
  the thin integration layer.
- Body composition is left to openScale's `StandardImpedanceLib` rather than porting the vendor's
  formulas.
- Bytes [8..9] of the payload are *not* decrypted by the vendor either, so their meaning is
  unverified and they are deliberately left undecoded.
- Matching is on service `0xF0A0` **plus** a manufacturer record that actually decrypts. The
  advertised name (`EL1`) is a meaningless vendor default shared by other eLink products, so it is
  not used. A checksum alone is not enough either — an all-zero record from another vendor
  satisfies it trivially — so the decrypted status byte must also be one the scale really sends.
- Unit codes other than kg (`1` jin, `6` lb, per the vendor's `UnitUtil`) were never observed on
  hardware, so the library refuses to convert them rather than guess.

## Testing

19 new unit tests (`AiLinkLibTest`, `AiLinkBroadcastHandlerTest`); full suite green (556 tests).

The library fixtures are **synthetic**: each was produced by running the protocol's own TEA
encryption over an invented reading, so they exercise the real key derivation, cipher, checksum
and parser while containing no personal health data — consistent with `ScaleCatalog`'s note that
fixtures are "built from the names and fingerprints documented in the handlers themselves, never
from personal captures".

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/ScaleFactory.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/ScaleFactory.kt
@@ -24,6 +24,7 @@ import com.health.openscale.core.bluetooth.scales.BeurerBF450Handler
 import com.health.openscale.core.bluetooth.scales.ScaleDeviceHandler
 import com.health.openscale.core.bluetooth.scales.AAAxHandler
 import com.health.openscale.core.bluetooth.scales.AfuB1Handler
+import com.health.openscale.core.bluetooth.scales.AiLinkBroadcastHandler
 import com.health.openscale.core.bluetooth.scales.FitTrackDaraHandler
 import com.health.openscale.core.bluetooth.scales.ScaleupHandler
 import com.health.openscale.core.bluetooth.scales.ActiveEraBF06Handler
@@ -155,6 +156,7 @@ class ScaleFactory @Inject constructor(
             SinocareHandler(),
             SenssunHandler(),
             RenphoHandler(),
+            AiLinkBroadcastHandler(),
             QNHandlerBroadcast(),
             QNHandler(),
             OneByoneHandler(),
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/libs/AiLinkLib.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/libs/AiLinkLib.kt
@@ -1,0 +1,320 @@
+/*
+ * openScale
+ * Copyright (C) 2026 openScale contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package com.health.openscale.core.bluetooth.libs
+
+/**
+ * Decoder for the **AiLink / eLink broadcast body-fat scale** protocol
+ * (vendor apps `com.pingwang.elink` / "AiLink", chipsets branded Aicare, Pingwang, eLink).
+ *
+ * These scales are *non-connectable*: they never accept a GATT link and instead repeat the whole
+ * measurement inside the manufacturer-specific data of an `ADV_NONCONN_IND` advertisement, with
+ * the interesting part encrypted.
+ *
+ * ## Provenance
+ * The protocol described below was determined for interoperability, by observing live
+ * advertisements from an "EL1" scale and by studying the behaviour of the vendor's own
+ * implementation (`BleValueBean.init`, `MainModel.BroadCast`, `BroadcastScaleDeviceData.dataCheck`
+ * and the native `libAILinkBle-lib.so`). What was recovered are facts about the wire format —
+ * field offsets, the checksum, the cipher and its key schedule — not source text. The
+ * implementation here is original: TEA itself is a published algorithm, and every routine below
+ * was written from the documented layout rather than transcribed from any decompiler output.
+ *
+ * ## Advertisement layout
+ * The scale advertises service `0000F0A0` and one manufacturer record. AiLink abuses the
+ * manufacturer field: the two bytes Android parses as the *company id* are really the vendor's
+ * CID and VID, so `ScanRecord.getManufacturerSpecificData(companyId)` hands back the record
+ * already stripped of them. Indices below are into that stripped array:
+ *
+ * ```
+ * companyId = VID << 8 | CID          (0x0301 -> CID 0x01, VID 0x03)
+ *   [0]      PID
+ *   [1..6]   MAC address, little-endian (reverse for display)
+ *   [7]      checksum: sum of [8..17] & 0xFF
+ *   [8..17]  10-byte payload, TEA-encrypted
+ * ```
+ *
+ * ## Payload encryption
+ * `AiLinkPwdUtil.decryptBroadcast` is plain 32-round [TEA](https://en.wikipedia.org/wiki/Tiny_Encryption_Algorithm)
+ * over **only the first 8 bytes** — bytes [8..9] of the payload are passed through untouched, and
+ * this decoder keeps that quirk so frames decode the same way the scale meant them. The 128-bit key
+ * is built inline from the device ids rather than from `get_ailink_key`'s MD5 path
+ * (see [broadcastKey]).
+ *
+ * ## Decrypted payload
+ * ```
+ *   [0]      rolling sequence counter
+ *   [1]      status (see STATUS_* below)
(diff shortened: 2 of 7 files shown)


oliexdev:
Thanks for this PR :+1: 

  Two things before I can merge:

  **1. Please rebase onto master.** `ScaleFactory.kt` and `ScaleCatalog.kt` conflict, but only in
  the import block and the handler list — master gained the HealthKeep, Picooc, Afu B1 and Hume
  Dara 2.0 handlers in the meantime. Nothing substantial.

  **2. A note on provenance.** The KDoc says "Ported from
  `Java_com_pinwang_ailinkble_AiLinkPwdUtil_decryptBroadcast`". I read that as "the key derivation
  was recovered from that function", which is fine — reverse engineering for interoperability is
  explicitly covered here, and openScale already ships the Xiaomi S400 decryption on the same
  footing. But since "ported" can also be read as "copied", could you reword it to something like
  "Derived from the observed behaviour of ..." and confirm in the PR description that the
  implementation is your own and that no decompiler output was copied in? Costs nothing and keeps
  the record clean.

  Minor, take it or leave it: the `val weightKg = frame.weightKg ?: return ...` in
  `onAdvertisement()` is redundant — `isUsable` already guarantees it is non-null.

manueltarouca:
Thanks for the quick review! All three addressed.

**1. Rebased onto master** (`6613db58`). The conflicts were exactly as you described — the import
block and the handler list in `ScaleFactory.kt` / `ScaleCatalog.kt`, nothing substantial. Rebased
rather than merged, so it's still a single commit. Full suite green afterwards (652 tests, up from
556 — master's new handlers came along with the rebase).

**2. Provenance.** Reworded to "Derived from the observed behaviour of ...", and I've added an
explicit `## Provenance` section to the `AiLinkLib` KDoc so the record lives with the code rather
than only in this thread.

To confirm here as well: **the implementation is my own and no decompiler output was copied in.**
Concretely, what that means for this PR:

- The protocol was determined for interoperability by observing live advertisements from the scale
  and by studying the behaviour of the vendor's implementation — jadx output for the Java side
  (`BleValueBean.init`, `MainModel.BroadCast`, `BroadcastScaleDeviceData.dataCheck`) and an
  `objdump` disassembly of `libAILinkBle-lib.so` for the encryption.
- What that yielded are *facts about the wire format*: field offsets, the checksum, which bytes are
  encrypted, and the key schedule (`k0 = pid + 0x41493000`, etc.). Those are the interface, not
  anyone's expression of it.
- TEA is a published algorithm with a public reference implementation, so the cipher itself owes
  nothing to the vendor.
- Every routine in `AiLinkLib` was written from the documented layout above. No decompiled Java or
  disassembly text was pasted in or mechanically translated; the structure, naming and the
  `Broadcast` model are unrelated to the vendor's.

Happy to reword further if you'd like it phrased differently.

**3. The redundant check** — good catch, and it pointed at a real duplication rather than just a
spare line. `isUsable` and `weightKg` were two half-answers to the same question, so I collapsed
them into one accessor on `Broadcast`:

```kotlin
val usableWeightKg: Float?
    get() = weightKg?.takeIf { !isNegative && rawWeight != 0 }

val isUsable: Boolean get() = usableWeightKg != null
```

The handler now unwraps once, and there's no second null check to fall out of sync:

```kotlin
val weightKg = frame.usableWeightKg
if (weightKg == null) { ...; return BroadcastAction.CONSUMED_KEEP_SCANNING }
```

oliexdev:
Thanks for your awesome PR :+1:

PR #1496, opened 2026-08-28
Fix ISO week aggregation consistency

### LINKED ISSUE #1454: WEEK aggregation: groups use the locale week, bounds use the ISO week
WEEK aggregation groups measurements with a **locale-dependent** week definition but computes the period boundaries with a **fixed ISO** one. When the two disagree, a group's own members fall outside the bounds reported for that group.

## The mismatch

Grouping and `periodKey` use the locale's first-day-of-week:

- `core/usecase/MeasurementAggregationUseCase.kt:86` — `WeekFields.of(Locale.getDefault())`
- `core/usecase/MeasurementAggregationUseCase.kt:99` — `date.get(wf.weekOfWeekBasedYear())`
- `core/data/Enums.kt:511` — `periodKey`, same locale lookup

But `periodBounds` is hardcoded to ISO:

- `core/data/Enums.kt:488` — `WEEK -> { val mon = date.with(DayOfWeek.MONDAY); mon to mon.plusWeeks(1) }`

## Demonstration

Run under `Locale.US` (first day of week = Sunday), with two measurements: **Sun 2025-04-06** and **Mon 2025-04-07**.

```
US first day of week : SUNDAY
Apr 6 (SUNDAY) week = 15  ┐ grouped together
Apr 7 (MONDAY) week = 15  ┘
periodBounds from minTs (Apr 6) = [2025-03-31, 2025-04-07)
does that range contain its own group member Apr 7? -> false
```

`minTs` is the Sunday, and `with(DayOfWeek.MONDAY)` walks *backwards* to the ISO week's Monday (Mar 31), so the reported period is the preceding ISO week — excluding the Monday that is in the same group, and covering Mar 31–Apr 5 which are not.

Consumers that filter by these bounds therefore select the wrong rows, e.g. `ui/screen/table/TableScreen.kt:278-279` and `:374-375`.

Under `Locale.GERMANY` (first day = Monday) the two definitions agree and the bug is invisible — which is likely why it has survived.

## Why it is not covered

`app/src/test/java/com/health/openscale/core/usecase/MeasurementAggregationUseCaseTest.kt` contains no occurrence of "week" at all — the existing tests cover empty input, `NONE`, and DAY grouping only. So neither the grouping nor the bounds are guarded for WEEK.

## Possible directions

Either make `periodBounds` locale-aware to match the grouping:

```kotlin
WEEK -> {
    val firstDay = WeekFields.of(Locale.getDefault()).firstDayOfWeek
    val start = date.with(TemporalAdjusters.previousOrSame(firstDay))
    start to start.plusWeeks(1)
}
```

…or make grouping ISO to match the bounds (`WeekFields.ISO`), which is more predictable across devices but changes what users in Sunday-first locales see. The first keeps current user-visible grouping behaviour; the second makes the data locale-independent. Your call which is intended — I couldn't tell from the code whether locale-sensitive weeks are a deliberate feature.

Happy to send a PR with either direction plus tests covering both locales, if you tell me which you prefer.

Fhatu12:
Fixes #1454.

## Summary
- use `WeekFields.ISO` for WEEK aggregation and `periodKey`
- keep WEEK bounds aligned to Monday-based ISO periods
- make WEEK labels use the same ISO week identity while retaining localised presentation
- keep aggregate table WEEK labels consistent with their stored period

## Tests
- US Sunday/Monday ISO boundary regression
- US/Germany locale-independence
- ISO week-based year boundary
- key/label/bounds consistency
- aggregate member/bounds consistency

## Validation
- `./gradlew testDebugUnitTest --tests '*AggregationLevelTest'` - passed
- `./gradlew testDebugUnitTest --tests '*MeasurementAggregationUseCaseTest'` - passed
- `./gradlew testDebugUnitTest` - passed
- `./gradlew compileDebugAndroidTestKotlin` - passed
- `./gradlew assembleDebug` - passed
- `git diff --check` - passed
- `git show --check HEAD` - passed

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/data/Enums.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/data/Enums.kt
@@ -508,7 +508,7 @@ enum class AggregationLevel(@param:StringRes val displayNameResId: Int) {
             NONE,
             DAY   -> date.toString()
             WEEK  -> {
-                val wf = WeekFields.of(Locale.getDefault())
+                val wf = WeekFields.ISO
                 "${date.get(wf.weekBasedYear())}-W${date.get(wf.weekOfWeekBasedYear())}"
             }
             MONTH -> "${date.year}-${date.monthValue}"
@@ -517,10 +517,10 @@ enum class AggregationLevel(@param:StringRes val displayNameResId: Int) {
     }
 
     /**
-     * Returns a human-readable, locale-sensitive label for the period containing [timestamp].
+     * Returns a human-readable label for the period containing [timestamp].
      *
-     * Intentionally separate from [periodKey] — labels are locale-dependent and must
-     * not be used as stable identifiers.
+     * Intentionally separate from [periodKey] — labels may use locale-sensitive
+     * presentation, but [WEEK] period identity remains ISO/device-independent.
      *
      * @param calendarWeekAbbrev Localised abbreviation for "calendar week" (e.g. "CW" / "KW").
      *                           Only used for [WEEK].
@@ -538,7 +538,7 @@ enum class AggregationLevel(@param:StringRes val displayNameResId: Int) {
                 DateTimeFormatter.ofLocalizedDate(FormatStyle.MEDIUM).withLocale(locale)
             )
             WEEK  -> {
-                val wf = WeekFields.of(locale)
+                val wf = WeekFields.ISO
                 "${date.get(wf.weekBasedYear())} – $calendarWeekAbbrev ${date.get(wf.weekOfWeekBasedYear())}"
             }
             MONTH -> date.format(DateTimeFormatter.ofPattern("MMMM yyyy", locale))
--- a/android_app/app/src/main/java/com/health/openscale/core/usecase/MeasurementAggregationUseCase.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/usecase/MeasurementAggregationUseCase.kt
@@ -29,7 +29,6 @@ import com.health.openscale.core.model.ValueWithDifference
 import java.time.Instant
 import java.time.ZoneId
 import java.time.temporal.WeekFields
-import java.util.Locale
 import javax.inject.Inject
 import javax.inject.Singleton
 import kotlin.math.roundToInt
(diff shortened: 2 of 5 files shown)


oliexdev:
Thanks — the analysis in #1454 was spot on and the invariant it describes is exactly right. I went a different way on the resolution, though, so I'm closing this in favour of b736d886.

  **Why not ISO.** The week rule now comes from the *system* locale, not from ISO. A US calendar starts on Sunday and that expectation is real; calendar-week numbers on the other hand are barely used there, so the grouping boundary matters more than the number. For DE/EU nothing changes — `WeekFields.of(Locale.GERMANY)` is  ISO.

  There is a trap that makes this less obvious than it looks: `SupportedLanguage.toLocale()` builds region-less locales (`de`, `fr`, `it`, …), and `WeekFields.of(Locale("de"))` is **Sunday**-based with a 1-day first week. On API 33+ `LocaleUtils.updateAppLocale` pushes that locale into `LocaleManager`, where it can also become `Locale.getDefault()`. So neither `effectiveLocale()`  nor `Locale.getDefault()` is a safe source — the rule is read from `Resources.getSystem().configuration.locales[0]` and falls back to ISO when no region is known.

  **Scope.** There were seven independent week implementations, not one, disagreeing in three ways — `periodBounds` hardcoded Monday, `periodKey`/`periodLabel` on `Locale.getDefault()`, `TableScreen` with two more copies, `MeasurementChartFilter` Monday buckets with locale labels, and `MeasurementChartLayers` using the `w` pattern letter with no locale at all.  `periodBounds`/`periodKey`/`periodLabel` now take the rule as a parameter and everything else calls into them.

  Your test cases carried over well — the Sunday/Monday boundary, the year boundary and the key/label/bounds consistency are all in there, parametrised over ISO and `Locale.US` so they no longer pass by accident on a Monday-first host. The one I added on top is `periodBounds_containEveryTimestampSharingThePeriodKey`: over 400 days, every level and both rules, each group member  must lie inside its own group's bounds. It fails against the old hardcoded Monday bound, which is what #1454 came down to.

Fhatu12:
Thanks for the detailed explanation. The system-locale distinction makes sense, especially the region-less locale behaviour on newer Android versions. Glad the boundary and consistency tests were useful. I also like the approach of passing one shared `WeekFields` rule through all of the aggregation and display paths. Thanks for taking the time to explain the final direction.

PR #1497, opened 2026-08-28
Added impedance support for Huawei AH/CH 100

D-Voi:
Enabled the impedance reading in the handler and added all the fields that derive from the impedance. Tested the changes in my phone with HUAWEI AH100 and they work. Please let me know if these changes make it into the play store app by emailing me in: voikiaetsi@gmail.com


oliexdev:
Thanks for wiring `StandardImpedanceLib` into the AH/CH100 handler.

  Before we look at the code: **do the derived values match reality on your device?**

  `StandardImpedanceLib`'s doc comment says the formulas only hold for whole-body impedance around 500 ± 100 Ω, and warns against using the class for scales reporting e.g. 250 or
  800 Ω. Your `1..3999` window suggests we don't yet know what the AH100 really sends.

  From two or three weigh-ins, could you post:

  - the raw impedance from the log line (`impedance=… Ω`)
  - what Huawei Health shows for fat, water, muscle and bone
  - what openScale shows with this patch

  If it's close, this is good to go. If it's far off, better to store nothing than wrong values — people track these over months.

  Two things for a later round:

  - The last commit reverted the `heightM > 0.0 && user.age > 0 && m.weightKg > 0f` guard from commits 2 and 3. That should come back — `ScaleUser.bodyHeight` defaults to `-1f`.
  - `fat` stays the scale's value while `lbm` comes from the library, so the two contradict each other. See `PicoocHandler.kt:341-352` for how to mix them.

  On your question from the first commit message: releases go to both F-Droid and Play Store, so this lands there once merged — GitHub will notify you, no email needed.

PR #1502, opened 2026-08-31
Fix: QNHandler for GE CS 10 G "Fit Plus": pairing, weight parsing, config

sebastian-ziegler:
Three independent bugs, all confirmed against a BTSnoop capture of the official "Fit Profile" app driving a real Fit Plus unit (a GE-branded Qing Niu / QN-Scale device):

1. supportFor() never claimed the device. Its pre-connect advertisement carries only the FFE0 service UUID -- AE00 and FFF0 are real GATT services but only visible after connecting, so neither the QN-family name list nor the AE00 relaxation matched "Fit Plus". Added "fit plus" as a recognized name.

2. Live-weight (0x10) frames were never recognized as valid. This unit reports weightScaleFactor=100 via the 0x12 handshake frame but still uses the ES-30M byte layout ([4]=state, [5,6]=weight), which was previously gated on weightScaleFactor==10 only. Added isLongFrameVariant, derived from the 0x12 frame's own structure (length byte == frame length) rather than the scale factor, and used it as an additional trigger for ES-30M-style parsing.

3. State==1 ("stabilizing") was treated as stable alongside state==2, but state==1 frames always carry R1=R2=0 in the capture -- treating it as final would publish a measurement with fabricated impedance and then discard the real state==2 reading via the hasPublishedForThisSession guard. Tightened to state==2 only.

Also, cross-referenced against the decompiled vendor Android SDK (CmdBuilder/QNDecoderImpl):

- Added the 0x1F stable-reading acknowledgment ([0x1F, 0x05, protocolType, 0x10, checksum]), which the vendor app sends unconditionally and openScale never sent at all.
- The 0x13 config command now carries the real user's height/age/gender instead of zeros, matching CmdBuilder.buildCmd(0x13, scaleType, [unit, lightInterval, height, age, gender]).

QNHandlerProtocolTest.kt is new and includes a regression test built from byte-for-byte frames taken from the BTSnoop capture (device address anonymized) covering bugs 2 and 3 together. QNHandlerMatchTest gains a matching test for the FFE0-only advertisement and has its now-incorrect "Fit Plus should not match" negative case corrected to a generic device name.

Verified end-to-end on the physical device: openScale now pairs and publishes real weight/impedance measurements.

I can provide a snoop capture file if needed.


oliexdev:
Thanks for the capture-backed write-up — the state 0/1/2 sequencing is clearly right, and the `Fit Plus` name match is fine as it stands.

  One change blocks this though: `isLongFrameVariant`.

  `byte[1]` is the frame length in this protocol, so `data[1] == data.size` is true for *every* well-formed frame, not just the 18-byte one. Your own classic-device test sets it:

  ```kotlin
  val scaleInfoFrame = ByteArray(15)
  scaleInfoFrame[1] = 15   // -> flag is true here too
  scaleInfoFrame[10] = 1   // -> weightScaleFactor 100
  ```
  That turns the layout gate into byte4Value <= 0x02 alone and removes the weightScaleFactor == 10 guard that keeps classic devices on the original layout. There, byte[4] is the weight's low byte — so on any QN/Yolanda/Renpho unit, a reading like 71.68 kg (0x1C00) is silently dropped and 71.70 kg (0x1C02) is parsed as weight from bytes[5,6], which are the stable flag and
  impedance. That's roughly 1 in 85 weigh-ins, with no error shown.

  The flag also isn't needed: your capture reports data[10] == 0x07, i.e. factor 10, so the ES-30M path already selects itself. Dropping the || isLongFrameVariant term keeps your hardware test at 85.35 kg.

  Could you:

  1. remove the flag, and add a test that a classic factor-100 frame with weight low byte 0x00/0x02 still decodes correctly;
  2. gate the state == 2 change, the 0x1F ack and the new 0x13 config bytes on seenProtocolType == 0xFF, unless you have captures from a classic QN/ES-30M unit — all three currently apply to every scale this handler serves;
  3. check the impedance: r1 = 4350 is stored raw, but real BIA is 300–800 Ω, so the device likely reports tenths and the derived composition is off.

  Master has moved on quite a bit meanwhile (measurement values are now written through typed keys, e.g. m[MeasurementType.WEIGHT] = Kg(x)), so the branch needs a merge.

Please verify your changes against your real scale. Thanks.

sebastian-ziegler:
Thanks for the detailed review @oliexdev  the isLongFrameVariant catch is right: byte[1] being the frame's own length is a property of every well-formed frame in this protocol (that's just how CmdBuilder.build works), not something specific to the 18-byte variant, so my check never discriminated anything. It was already removed entirely. It also turned out to be unnecessary in the first place: weightScaleFactor == 10.0f alone already selects the right layout for the captured Fit Plus family, because its 0x12 frame's byte[10] fails the existing == 1 check the same way a genuine factor-10 device's would (I'd conflated the vendor SDK's real & 1 check with this codebase's pre-existing == 1 one when I wrote the original patch). Added the regression test you asked for — a classic factor-100 frame with weight low byte 0x00 — which fails against the old code and passes now.

Scoped the other three changes (state==2-only, the 0x1F ack, real height/age/gender in 0x13) behind protocolType == 0xFF, with a companion test for each proving other protocolTypes keep their original behavior.

On the impedance: seems like you are right. I dug back into the decompiled vendor SDK — MeasureDecoder.h() (the function actually used for this 2-electrode path, as opposed to the 8-electrode g() which does apply a kRatio scale) combines the two bytes with no scaling at all, so 4350 is genuinely what the vendor's own code would produce too. But TrisaBodyAnalyzeLib's formula here is the well-documented one from the wider openScale project, and its coefficients are calibrated against impedance in the ~300–1000Ω range — and for what it's worth, the existing 0.3*(r1-400) transform's constants (410, 400) only really make sense against an input already in roughly that range, not the ~4000s we're seeing. So I've scoped impedance = r1 / 10.0f (replacing the ad-hoc transform) to the same protocolType == 0xFF family, leaving every other device's raw-r1 + existing-transform path untouched. I don't have the vendor's actual proprietary body-fat formula to confirm this against, so if you have a capture from another device (or know how that 0.3*(r1-400) was originally derived) that says otherwise, I'd rather match your evidence than mine.

oliexdev:
Thanks for the fix and your PR :+1:

PR #1507, opened 2026-09-02
Fix crash on every successful weigh-in: unwrap Kg in saved-measurement 

sebastian-ziegler:
While testing #1502 I noticed there was a bug introduced in master, here the description of it, already tested this in a local dev:

39a0c5c ("BLE handlers can contribute custom measurement types", #1481/#1484) changed BleConnector's "measurement saved" snackbar args from the old Float measurementData.weight to measurementData[MeasurementType.WEIGHT] ?: 0f. That subscript returns Kg? (the Key<Kg> type), not Float?, so whenever a measurement actually has a weight (i.e. on every successful weigh-in on every scale), the snackbar's format arg list carries a Kg object instead of a raw Float. Resources.getString then throws formatting it against the "%1$.1f kg" specifier in bluetooth_connector_measurement_saved:

  java.util.IllegalFormatConversionException: f != com.health.openscale.core.data.Kg
    at ... Resources.getString
    at ... AppNavigationKt$AppNavigation$2$1$1$2.emit

Fix: unwrap .value to get the Float the format string expects.

Added a regression test exercising the exact Resources.getString call that threw (constructing the full BleConnector needs a live ScaleCommunicator and DB facades, so the test mirrors just the argument-building step instead). Verified it fails against the pre-fix expression and passes against the fix.


oliexdev:
Thanks for the hint and your PR :+1:

PR #1376, opened 2026-05-22
Mi scale 2 dual impedance

oliexdev:

PR #1508, opened 2026-09-03
Xiaomi S400: use 50 kHz for estimation equations

miszobi:
## Problem

`S400BodyComposition` (added in #1367 / #1376, cc: @DanyPM ) feeds the **high-frequency**
impedance band to the Sun 2003 (TBW) and Janssen 2000 (SMM) equations. Both were
derived on **50 kHz** single-frequency BIA, and the S400's high-frequency band is
250 kHz.

For my setup this caused FFM and TBW values to be pushed out of range and suppressed.

## Change

`compute()` already computes `rL` (the corrected low-frequency band) for the
Hanai ECW step. This PR feeds `rL` to Sun 2003 and Janssen 2000 as well, instead
of `rH`.

## Testing

`./gradlew :app:testDebugUnitTest` — full suite green. The new
`usesLowFrequencyForPrediction` test fails without the one-line fix.

### DIFF
--- a/android_app/app/src/main/java/com/health/openscale/core/bluetooth/libs/S400BodyComposition.kt
+++ b/android_app/app/src/main/java/com/health/openscale/core/bluetooth/libs/S400BodyComposition.kt
@@ -38,6 +38,10 @@ private fun cbrt(x: Double): Double = x.pow(1.0 / 3.0)
  * `rLowRaw` (Ω, ~50 kHz). Heart rate, if present, is **not** an input to any
  * body-composition equation — pass through to the UI unmodified.
  *
+ * The single-frequency prediction equations (Sun 2003, Janssen 2000) were
+ * derived on 50 kHz BIA, so they are fed the corrected **low-frequency** band.
+ * The high-frequency band only informs the §2.1 Cole-Cole checks.
+ *
  * ## Validation (§1.1) — reject entire computation
  * `age 18-120` (Janssen/Cunningham not validated <18), `height 100-230`,
  * `weight 20-250`, `R_high/R_low 200-1500`, `BMI 12-60`. These are not hard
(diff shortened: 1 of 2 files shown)


oliexdev:
Before I go further into this one, a few questions — I'd like to understand the change before judging it.

  1. After this change, does `rHigh` still feed any prediction? From reading `compute()` it looks like it's left with the Cole-Cole swap, the `unreliableContact` check and the empirical bone formula,
  which is why `rH` can be deleted. If that's right, TBW and ECW both come from the same resistance, and `ICW = TBW - ECW` is the difference of two equations on one input rather than something taken from
  the gap between the bands. How do you see that?

  2. `ecwTbwRatio` moves from 0.42–0.45 to 0.46–0.48 across all three reference subjects. Still inside the 0.30–0.55 guard, but what's your read on those values?

  3. Body fat rises by roughly 5 points for each reference subject (A 14.1 → 19.5 %, B 41.5 → 46.2 %, C 34.5 → 40.3 %), so every S400 user would see a step in their chart. Is that the intended magnitude?

  4. How did you arrive at the new §7.1–7.3 expected values — by running the modified code, or checked against the specification document from #1367/#1376? I recomputed A and B by hand and they match the
  new implementation exactly (A: TBW 45.07, FFM 61.57, SMM 33.84; B: TBW 26.79), which is why I'm asking: the test docstring still describes them as "the spec's reference subjects".

  5. Can you say more about the case that triggered this — which values were suppressed, and what does the scale's own app report for the same weighing?

  @DanyPM, since the pipeline and the spec are yours, your view on 1 and 4 would help a lot.

DanyPM:
Hi! You're right, Sun 2003 and Janssen 2000 are both 50 kHz regressions and I fed them the 250 kHz band. My fault

1) Yes. After this change `rHigh` feeds only the label swap, the contact check and the `MI_LEGACY` bone regression, so `ICW = TBW - ECW` becomes the difference of two functions of one resistance. The second packet stops carrying information.

4) As @oliexdev says, the expected values match the implementation exactly, which is what your hand-recomputation shows ("the spec's reference subjects") but regenerated from the code they can't detect a wrong implementation. They should be recomputed by hand from the equations, KDoc updated in the same commit.

I'll have a look at the litterature to see how we can fix the issue you highlighted, but the current approach is not sufficient.

oliexdev:
Thanks @DanyPM, that helps a lot.

 Let's wait for your look at the literature before touching the equations, then.

 @miszobi, thanks for digging this up — the problem you hit is real, I'd just rather fix it properly than quickly.

joner391:
Wanted to add to this, not sure if its useful. But OpenScale body fat estimations differ from the Mi App by 6 percentile units (in my specific case at least).

PR #1517, opened 2026-09-07
fix: support alternate manufacturer ID for Sinocare CW286

albertomarturano:
## Summary

Fix Sinocare CW286 broadcast handling for devices that expose the Sinocare
manufacturer payload under manufacturer ID `0x0000` instead of `0xFF64`.

The device is already identified by the advertised name `Weight Scale`, but
`SinocareHandler.onAdvertisement()` previously only looked up manufacturer
data under `0xFF64`, causing these advertisements to be ignored.

The handler now tries `0xFF64` first and falls back to `0x0000`.

Device detection is intentionally unchanged, so `0x0000` alone is not used as
a Sinocare fingerprint.

## Testing

- Added a regression test using a sanitized advertisement captured from a
  physical Sinocare CW286.
- The captured frame decodes as 74.30 kg and reaches the existing stability
  threshold.
- Verified successfully on a physical CW286 using the debug APK.
- `assembleDebug` succeeds.


oliexdev:
Thanks for your PR :+1:

PR #1519, opened 2026-09-07
feat: Add support for FITINDEX FT-26R-W (QN broadcast, bit-0 stable flag)

### LINKED ISSUE #1518: Support FITINDEX FT-26R-W, and don't let the stabilize window drop stable broadcast frames
### Preflight

- [x] I tested the latest dev version (built from `7476f16`, the same commit as the current `dev-build` release)
- [x] The issue still occurs in the dev version
- [x] I searched the existing open and closed issues

---

**Model:** FITINDEX FT-26R-W — a weight-only (no BIA electrodes), non-connectable
QN/AABB broadcast scale. It is already claimed by `QNHandlerBroadcast`, but never
records a measurement.

Two separate causes, verified against captures from a physical unit:

**1. The stable flag is bit 0, not bit 5.** This model never sets
`FLAG_STABLE_BIT` (`0x20`). Its status byte runs `0x14` while the weight moves and
latches to `0x15` when the display settles, so every reading is discarded:

```
status=0x14  weight=70.35 kg   <- moving
status=0x15  weight=70.30 kg   <- settled (display read 155.0 lb = 70.31 kg)
```

**2. The stabilize window drops the frame that matters.** Fixing the bit alone
only worked about 1 time in 5. `BroadcastScaleAdapter`'s 1200 ms
`stabilizeWindowMs` check is purely time-based, and this scale advertises every
~130 ms but holds the stable status only a second or two before sleeping. Across
three on-device sessions the stable frame arrived 920 ms, 66 ms and 67 ms after
the previous forward — every one inside the window. Unlike a "still moving" frame,
that one does not come back.

This second part is not FITINDEX-specific: any broadcast-only scale that
advertises faster than the window and flags its settled reading in the
advertisement can lose measurements this way.

## Proposed fix

I'm opening this alongside a PR for review, since part of the fix touches shared
Bluetooth infrastructure and the contribution guidelines ask for an issue first:

1. `QNHandlerBroadcast`: accept bit 0 as well as bit 5. Safe to test together —
   the existing `WEIGHT_MIN_KG` check already rejects the idle advertisement,
   which is the only frame setting bit 0 without a real weight.
2. `ScaleDeviceHandler`: add `open fun isTimeCriticalAdvertisement(result:
   ScanResult): Boolean = false`.
3. `BroadcastScaleAdapter`: let a frame the handler marks time-critical bypass the
   stabilize window and de-dup.

Default `false` means no existing handler changes behaviour.

Shortening `stabilizeWindowMs` would forward many more frames for every broadcast
scale and still races; handling it inside the handler isn't possible, since the
adapter drops the frame before the handler is called. Happy to rework this in
whatever shape you prefer, or to split (1) from (2)+(3) into separate PRs.

The PR includes unit tests built from the captures: 3 of 3 weighings recorded on
device (Pixel 11 Pro), versus 0 of 5 on master.

lanrat:
Closes #1518.

**Model:** FITINDEX FT-26R-W — a weight-only, non-connectable QN/AABB broadcast
scale. No BIA electrodes, so weight only; BMI still comes from the app's profile
height as usual. Already claimed by `QNHandlerBroadcast`, but never recorded a
measurement.

## Summary

**1. Accept bit 0 as a stable flag** (`QNHandlerBroadcast`). This model never sets
bit 5 (`0x20`); its status byte runs `0x14` while the weight moves and latches to
`0x15` when the display settles. Testing both bits together is safe because the
existing `WEIGHT_MIN_KG` check already rejects the idle advertisement, the only
frame that sets bit 0 without a real weight.

**2. Let time-critical frames bypass the stabilize window**
(`ScaleDeviceHandler` + `BroadcastScaleAdapter`). The 1200 ms window is purely
time-based and drops whatever lands inside it, including the frame carrying the
final reading. This scale advertises every ~130 ms but holds the stable status
only briefly before sleeping, so that frame was usually discarded — and unlike a
"still moving" frame, it does not come back. Measured across three sessions, the
stable frame arrived 920 ms, 66 ms and 67 ms after the previous forward.

The new `isTimeCriticalAdvertisement()` hook defaults to `false`, so no existing
handler changes behaviour. This affects any broadcast-only scale that advertises
faster than the window and flags its settled reading in the advertisement, not
just this model. Happy to split this half into a separate PR if you'd prefer to
review it independently.

## Testing

- Added `QNHandlerBroadcastFT26RTest`: 8 regression tests using advertisements
  captured from a physical FT-26R-W, covering the wire format, the `0x14`→`0x15`
  transition, the absence of `0x20`, and idle-frame rejection.
- The captured frame decodes as 70.30 kg, matching the scale's own display
  (155.0 lb).
- Verified on a physical FT-26R-W using the debug APK: 3 of 3 weighings recorded,
  versus 0 of 5 on master.
- `./gradlew testDebugUnitTest` — 780 tests, no new failures.
  (`ScaleCatalogTest.every handler links to an existing source file` fails in my
  sandbox both with and without these changes; it resolves the repo root via
  `../..` and my build container mounts only `android_app`.)
- `assembleDebug` succeeds.

No `btsnoop_hci.log`: the scale is broadcast-only, so there is no GATT traffic to
capture and the vendor app never connects to it. The exact payloads are in the
tests, with the device address in bytes [2-7] replaced by the RFC 7042
documentation address (nothing parses those bytes).


oliexdev:
Thanks for the write-up, but I'm declining this.

  `QNHandlerBroadcastFT26RTest` never touches `QNHandlerBroadcast` — it reimplements
  the parsing and the stable check locally, so every test stays green if I revert
  `FLAG_STABLE_BIT_ALT`. That's no regression coverage.

  `QNHandlerBroadcast` claims the whole AABB/0xFFFF family (Renpho, ES-CS20M).
  Accepting bit 0 for all of them based on one FT-26R-W means that if bit 0 means
  something else elsewhere, the handler publishes a mid-weighing value and stops the
  scan — a wrong weight, written silently. `WEIGHT_MIN_KG` only catches the 0 kg
  idle frame.

  `isTimeCriticalAdvertisement()` is new public API on every handler, called from
  the adapter's hot path before `ensureAttached()`. And the de-dup bypass can't do
  anything here: your own test says byte 8 is a free-running counter, so de-dup
  never fires on this device.

  If you want to take this further: a QN-only PR, tests that drive the real handler
  through a real `ScanRecord`, a stable check tight enough to leave the other AABB
  models alone,

Issue #1333, opened 2026-03-16
Support for the Xiaomi Smart Scale S200


mirenbz:
Hi, 

I've tried following the directions from [How to support a new scale](https://github.com/oliexdev/openScale/wiki/How-to-support-a-new-scale) for a Xiaomi Smart Scale S200, but I'm quickly lost (as I have no idea of Android development, duh). I managed to get the log files so I could share them if someone is willing to help...

The Xiaomi Smart Scale S200 is a simple, quite cheap device that just logs weight (no body composition). Hope someone else is interested in getting this scale supported!

Thanks for the great work!

oliexdev:
If somebody as a working implementation please send a PR. Thanks.

PR #263, opened 2018-05-11
Logo Designs

jbeguna04:
Here are the designs in PNG files sir. Thank You

PR #1520, opened 2026-09-08
Add support for Xiaomi Smart Scale S200

jeanpijon:
Adds a handler for the Xiaomi Smart Scale S200 (MJTZC02YM), requested in #1333. Like the S400/S800, it is broadcast-only: it advertises AES-128-CCM encrypted MiBeacon v5 frames in service data 0xFE95 rather than accepting a GATT connection, so the handler runs in BROADCAST_ONLY mode.

Frame layout, and the decrypted measurement object:

    FC(2 LE) | PID(2 LE) | cnt(1) | [MAC(6) if FC&0x10] | [cap(1-2) if FC&0x20]  | cipher | extCnt(3) | MIC(4)
    nonce = macFrameOrder(6) || data[2..5) || extCnt(3)   AAD = 0x11   tag = 32 bit

    object 0x4e16, 9 bytes:
      [0]     profile id (Mi Home user slot, 1-based)
      [1..4]  weight, u32 LE, x0.01 kg
      [5..8]  timestamp, u32 LE, Unix seconds

Three product ids (0x45C9, 0x4C04, 0x4DCB) are all this same model under different SKUs/regions. Unlike the S400, the object carries only weight - no impedance, no heart rate - so this scale cannot report body composition, and unlike GATT scales it cannot stream a live/in-progress reading: an in-progress advertisement is either the objectless idle beacon or a zero-weight packet, both of which the lib discards, so there is no intermediate value to show even in principle. A bind key from Xiaomi Cloud is mandatory; nothing decrypts without one.

The frame control's capability bytes (FC & 0x20) are handled here, which the existing XiaomiS800Lib decryptor does not - a gap that only matters when a frame includes them, which this scale's own frames do not, but is cheap to handle correctly rather than carry the gap forward.

Protocol parsing lives in XiaomiS200Lib, free of Android dependencies and covered by unit tests. Its primary fixture is not synthetic: it is the `xiaomi-ble` project's own public upstream test vector (Bluetooth-Devices/xiaomi-ble, MIT licensed, test_Xiaomi_Scale_S200_MJTZC02YM), independently re-verified against pycryptodome before being committed, so it cannot hide a shared bug between this port and its source. The idle-beacon fixture is a real advertisement publicly posted in that project's issue #263. No personal health data appears in any fixture. The protocol itself is a clean-room reimplementation of what that library documents (obj4e16 / _parse_xiaomi) - no code is ported from it.

While working on this, I found what looks like a pre-existing bug in XiaomiS800Lib.parseWeightKg: it reads the same object 0x4e16's weight at byte offset 7, which against this scale's verified layout lands on the high half of the timestamp instead of the weight. I have not touched that code - its own tests use synthetic fixtures only, the handler's comments already note the BLE wiring is unverified on hardware, and I have no S800 to check against, so I did not want to guess at a fix for a scale I cannot test. Flagging it here in case it's useful; happy to open a separate issue instead if preferred.

Verified against real hardware with a real bind key: the idle beacon is correctly ignored, a real encrypted measurement decrypts to the expected weight matching the scale's own display, and repeated separate weigh-ins each land as their own measurement.


oliexdev:
Please extend `XiaomiS800Handler` instead of adding a second one. Both scales
  speak the same MiBeacon v5 framing and the same `0x4e16` object, and your decrypt
  routine is a strict superset of the S800's — it differs only by the capability
  bytes and the address fallback. So this collapses into: teach `XiaomiS800Lib`
  those two things, fix the weight offset to your verified layout, add the S200
  product ids and name patterns to `supportFor()`, and add a second `device()` row
  to the catalog. `XiaomiS200Lib` and `XiaomiS200Handler` then go away entirely.

  One handler covering several products is the established pattern here —
  `supportFor()` already returns a per-device `displayName` and capability set, so
  the S800 keeps its capabilities and the S200 reports `emptySet()`.

PR #1521, opened 2026-09-09
feat: Add support for FITINDEX FT-26R-W (QN broadcast, bit-0 stable flag)

lanrat:
Reworked from #1519 along the lines you outlined: QN-only, no shared
infrastructure touched, and tests that drive the real parsing code.

**Model:** FITINDEX FT-26R-W — a weight-only, non-connectable QN/AABB broadcast
scale. No BIA electrodes, so weight only. `QNHandlerBroadcast` already claims it,
but it never recorded a measurement.

## Summary

The handler treats bit 5 (`0x20`) of the status byte as "stable", and this model
never sets it. Its status runs `0x14` while the weight is moving and latches to
`0x15` once the display settles, so every reading was discarded:

```
status=0x14  weight=70.35 kg   <- moving
status=0x15  weight=70.30 kg   <- settled (display read 155.0 lb = 70.31 kg)
```

The wire format moves into `QnBroadcastAdv`, a plain Kotlin object with no Android
dependencies, which learns the bit-0 encoding.

On your concern about the rest of the AABB family: bit 0 is only honoured for
payloads that also carry the FT-26R's fixed signature at bytes [19-21], constant
across every frame captured from the unit while the counter, status and weight
varied around it. Any device that doesn't match keeps the original bit-5-only
rule, so an unrecognised AABB model can't be made worse by this. The signature
comes from one unit, so it only ever widens acceptance — there's a test asserting
a non-matching payload with bit 0 set is *not* treated as stable.

## Dropped from #1519

- `isTimeCriticalAdvertisement()` and the `BroadcastScaleAdapter` changes. You
  were right to push back, and for a better reason than I gave: re-measuring with
  the scale showed it holds the stable status for ~10 seconds, not the second or
  two I claimed, so roughly 7 stable frames already survive the 1200 ms window.
  The throttle was never the problem. Both files are untouched here.
- The de-dup bypass, which as you noted could never fire on this device.

## Testing

- `QNHandlerBroadcastFT26RTest`: 11 tests over advertisements captured from a
  physical FT-26R-W. They exercise `QnBroadcastAdv` and a real `ScanRecord`
  (`parseFromBytes`) rather than reimplementing the parsing — **reverting the
  bit-0 support fails two of them**, including the end-to-end one.
- Covers the wire format, the `0x14`→`0x15` transition, the absence of `0x20`,
  rejection of the zero-weight idle frame, and that bit 5 still works on
  non-FT-26R payloads.
- Verified on a physical FT-26R-W using the debug APK: every weighing the scale
  reported as settled was recorded.
- `./gradlew testDebugUnitTest` — 783 tests, no new failures.
  (`ScaleCatalogTest.every handler links to an existing source file` fails in my
  sandbox both with and without this change; it resolves the repo root via `../..`
  and my build container mounts only `android_app`.)
- `assembleDebug` succeeds.

## Known limitation

The scale sometimes ends a weighing without ever latching `0x15` — in one
capture it held a rock-steady 71.10 kg for 14 seconds, then stopped advertising
and went to sleep still reporting `0x14`. Nothing is recorded in that case, by
design: publishing a weight the scale hasn't marked final is exactly the silent
wrong-value risk worth avoiding. Stepping back on gets a normal reading. This is
noted in the catalog remark.

No `btsnoop_hci.log`: the scale is broadcast-only, so there is no GATT traffic to
capture and the vendor app never connects to it. The exact payloads are in the
tests, with the device address in bytes [2-7] replaced by the RFC 7042
documentation address (nothing parses those bytes).


oliexdev:
Three things:

  **Keep the parser in the handler file.** `libs/` is for reusable algorithms — that's why `AiLinkLib` is there.
  This one serves a single handler and was already inline. Follow
  `PicoocBroadcastHandler.kt`: same shape you wrote, `internal object
  PicoocMiniLiteAdv` with `parse(manufacturerId, payload): Frame?`, declared in the
  handler file. Your test already runs under Robolectric, so it reaches the parser
  inline just as well.

  **Trim the catalog remark.** That file feeds the user-facing scale table. Which
  status bit marks a settled reading, the company id, the AABB name — all code
  comments; the KDoc is the right place. Keep what a user can act on and which are not already in the table.

  **The `[19-21] = 51 0E 03` gate** is my real reservation. If those bytes are
  constant across the AABB family — a firmware or protocol version is an obvious
  candidate — the gate is a no-op and bit 0 is honoured for every AABB device,
  exactly what it's meant to prevent. If they're model-specific, the next FT-26R
  owner with different firmware silently gets nothing. We have no second capture to
  tell the two apart.

lanrat:
Thanks — first two done, and the third is a fair reservation I can't fully close.

**Parser back in the handler file.** You're right, `libs/` is the wrong home for it;
it's now `internal object QnBroadcastAdv` in `QNHandlerBroadcast.kt`, same shape as
`PicoocMiniLiteAdv`. The tests reach it inline as you said.

**Catalog remark trimmed** to the two things a user can act on: weight only (no
electrodes), and stay on until the display settles.

**On the `[19-21] = 51 0E 03` gate** — you've framed the problem exactly, and I
don't have the capture to resolve it. I went looking: the vendor APK's broadcast
parsing is in a native blob with nothing legible, so I can't tell a protocol
version from a model id either.

I did try replacing the gate with something self-evidencing — honour bit 0 only
after seeing it *clear* on an earlier weight-bearing frame from the same device,
since the FT-26R always sends several 0x14 frames before its first 0x15. It tests
well but has a race I'd rather not ship: a handler attached to an already-settled
scale never sees the transition and so never publishes. I reverted it.

So the gate stands as the conservative option, and its worst cases are the two you
named:

- If those bytes are family-wide, the gate is a no-op and bit 0 is honoured
  everywhere — the current behaviour of this PR for any AABB device, and the thing
  you want to avoid.
- If they're model-specific, another FT-26R with different firmware silently gets
  nothing — no worse than today, where it gets nothing regardless, but not a fix
  for them either.

Your call on which risk to take, and I'm happy to implement whichever: merge as
is, drop the gate and honour bit 0 for the whole AABB family, or hold this until
someone with a second AABB scale can post a capture. You have visibility into the
rest of that family and I have exactly one device.

One correction to the PR text while I'm here: I'd written that the scale holds its
stable state only briefly. Measuring it properly, it settles after about six
seconds and then holds 0x15 for ten. The practical consequence is the note now in
the catalog remark — step off early and nothing is recorded, which accounts for
every "missed" weighing I saw while testing. Confirmed by recording the
advertisements independently while the app ran: across that session the scale
never reported a finished measurement and openScale published nothing — the two
agreed exactly, with no dropped stable frame.

oliexdev:
Thanks for the rework :+1:

PR #1522, opened 2026-09-13
fix(bluetooth): accept bit-0 stable flag and decode BIA for QN/Renpho broadcast scales

iajmdc:
### Summary

Expands `QNHandlerBroadcast` to support full measurement recording and bioelectrical impedance analysis (BIA) body composition derivation for modern Yolanda / Qing Niu / Renpho broadcast-only scales (e.g., Renpho ES-26M-W, CS20x series, FITINDEX FT-26R).

### Background & Context

Recent Yolanda/Renpho smart scales advertise weight and impedance via non-connectable BLE advertisements (`ADV_NONCONN_IND`) under Manufacturer Specific Data (Company ID `0xFFFF`, `0xAA 0xBB` magic header).

1. **Settled Weight Flag (`0x15` / Bit 0):** While in motion, the scale broadcasts status `0x14` and transitions to status `0x15` (bit 0) upon latching a settled weight. Previously, bit 0 was gated exclusively on the FT-26R model signature (`[51 0E 03]`), causing other AABB models like the Renpho ES-26M-W to be discarded as non-stable.
2. **Bio-Impedance (BIA):** On BIA-capable models, primary resistance (in Ohms) is broadcast at bytes `[19-20]` (little-endian `uint16`).
3. **Nameless Advertisements:** Many AABB broadcast scales broadcast without a BLE Local Name, showing a blank string on the dashboard button.

### Changes

- **`QNHandlerBroadcast.kt`**:
  - Accept both bit 5 (`0x20`) and bit 0 (`0x01` / status `0x15`) as valid settled weight flags across all AABB broadcast packets.
  - Preserve `hasFt26rSignature()` to distinguish weight-only FT-26R units from BIA models so the fixed signature bytes `[51 0E 03]` are never misparsed as impedance.
  - Deterministically decode BIA resistance from bytes `[19-20]` (`uint16_le`) within plausible human impedance ranges (200–1500 Ω).
  - Derive body composition (`BODY_FAT`, `WATER`, `MUSCLE`, `BONE`) via `TrisaBodyAnalyzeLib` when impedance and user profile height are present.
  - Add `DeviceCapability.BODY_COMPOSITION` to `supportFor()`.
- **`BluetoothActionButton.kt`**:
  - Fall back to `savedDevice.determinedHandlerDisplayName` when `savedDevice.name` is null or blank.
- **`QNHandlerBroadcastFT26RTest.kt`**:
  - Update unit test assertions for universal bit-0 acceptance and add test coverage for Renpho ES-26M-W BIA impedance decoding.

### Verification

- [x] **Physical Hardware:** Verified against physical Renpho ES-26M-W scale on Android 12+ (successful real-time weight capture, display latching, and BIA body composition calculation).
- [x] **Automated Tests:** All 790 unit tests passing (`./gradlew testDebugUnitTest`).


oliexdev:
My main concern is that both changes drop the device gating and apply to every AABB broadcast scale:
 
  1. **Bit-0 stable flag for all devices.** In #1521 this was deliberately gated on the FT-26R signature because a false "stable" silently writes a wrong weight into the user's history. Two confirmed devices aren't enough for me to call it universal — for the bit-5 models (ES-CS20M variant 3) we have no captures showing that their bit 0 isn't used for something else. With the new  mask, status bytes like 0x01, 0x05, 0x21 or 0x35 also count as settled.

  2. **Impedance from bytes [19-20].** This is pure plausibility (`200..1500`), with no device fingerprint. On any other AABB model where those bytes are a counter or checksum, byte 20 only has to be 0x00..0x05 for the value to look valid — and then we silently store an invented impedance plus derived body fat / water / muscle / bone. That's exactly the failure mode the signature  gate was there to prevent.

  Could you gate the impedance decoding (and ideally the bit-0 flag) on a positive signature for the BIA models, the way `FT26R_SIGNATURE` does, instead of a value range? Since you have the ES-26M-W, real captured frames — moving → settled, with and without impedance — would give us that, and they'd replace the synthesized test frame, which currently only asserts that the parser does what the parser does.

  Two smaller things:
  - `BODY_COMPOSITION` is now declared for every AABB device, including the known weight-only FT-26R. `supportFor()` has the manufacturer data, so it could gate on the signature.
  - The handler publishes on the *first* stable frame and stops scanning. If impedance only arrives a few frames after the weight latches, we'd silently store weight only. Does the ES-26M-W already carry impedance in its first 0x15 frame?
  
  Please also drop the `BluetoothActionButton.kt` change — not in this PR and not as a separate one, I want to keep that file as it is.

iajmdc:
Thanks for the thorough review and guidance! You make completely valid points regarding defensive parsing and guarding against silent history corruption on unverified hardware.

I've updated the implementation to address all items:

### 1. Positive Signature Gating (`isBiaCapable`)
Rather than broadening bit-0/impedance parsing globally, we now explicitly gate both:
- **`FT-26R`**: Continues to use `FT26R_SIGNATURE = [51, 0E, 03]` at offset 19, remains weight-only, and accepts status `0x15`.
- **`Yolanda / Renpho BIA`**: Gated on positive structure via `isBiaCapable()`:
  - 24-byte payload layout with fixed vendor padding bytes `[12-14] == [FF, FF, FF]` and model vendor tag at byte 11 (`0x6A`).
  - Distinguishes settled status `0x23` / `0x15` and decodes BIA resistance.
- **Unknown / Generic AABB Models**: Strictly retain the original **bit-5-only (`0x20`) rule** and never attempt impedance decoding. Status bytes like `0x01`, `0x05`, or `0x21` will continue to be ignored.

### 2. Verified Hardware Timing & Impedance Format
From physical captures on the ES-26M-W:
- **Moving Frame (`status 0x02`)**:
  `aabb...02a06c1b 8813 038c12` (moving weight, status `0x02`)
- **Settled Frame (`status 0x23`)**:
  `aabb...23a0581b 8813 03b312` (70.00 kg, 0x1388 = 5000 ADC counts)

**Impedance Encoding:** Like the "Fit Plus" family in `QNHandler.kt`, raw ADC counts in bytes `[19-20]` are reported in tenths of an Ohm (`5000` → `500.0 Ω`).

**Timing:** The ES-26M-W carries the resistance in its settled frame. For robustness, if a BIA scale settles without resistance (e.g. user weighing with socks), the handler holds the scan for up to 2.5s before safely falling back to publishing weight-only.

### 3. Capability Gating in `supportFor()`
`supportFor()` now inspects the advertisement payload:
- FT-26R is declared with `setOf(DeviceCapability.LIVE_WEIGHT_STREAM)` only.
- BIA models (`isBiaCapable`) declare `setOf(LIVE_WEIGHT_STREAM, BODY_COMPOSITION)`.

### 4. Real Captured Test Vectors
Replaced the synthetic test vector in `QNHandlerBroadcastFT26RTest.kt` with actual hardware captures (moving, settled weight, and settled BIA at 500.0 Ω, using standard 70.00 kg / RFC 7042 vectors). All 790 unit tests pass.

### 5. Reverted `BluetoothActionButton.kt`
Reverted `BluetoothActionButton.kt` to match `master` byte-for-byte. The PR is now strictly confined to the 2 Bluetooth handler files (`QNHandlerBroadcast.kt` and `QNHandlerBroadcastFT26RTest.kt`).

oliexdev:
Thanks for the rework, but I don't think the gating does what the description says.

  **1. `isBiaCapable()` isn't a fingerprint.** It only checks `[12-14] == FF FF FF`, and that's in every AABB frame — including the FT-26R captures in the same test file (`moving1`, `settled1`, `idle`). Byte 11 `0x6A` is mentioned in your comment but never checked in code, and the FT-26R `idle` frame has `0x6A` there too. So the gate effectively means "AABB, ≥21 bytes, not  FT-26R": bit-0 stable and impedance decoding still apply to every unknown model, including the bit-5 ES-CS20M variant 3 I was worried about. The new test only exercises a frame with bytes 12-14 zeroed, which no real device sends.

  **2. The impedance range got wider, not narrower** — was `200..1500`, now plus `2000..15000 → /10`. Your own FT-26R signature bytes `51 0E` decode to 3665 → 366.5 Ω, a perfectly plausible impedance; only the signature check saves it. It collides the other way too: `FT26R_SIGNATURE` is `[19..21] = 51 0E 03` and byte 21 is `0x03` in both Renpho frames, so a real ES-26M-W reading  of 3665 raw would be taken for an FT-26R.

  **3. The Trisa conversion is scaled twice.** You describe the encoding as tenths of an ohm like the Fit Plus family, but that path in `QNHandler.publishQnMeasurement()` deliberately passes the ohm value straight to Trisa after `/10`. Here we do `/10` *and* `0.3f * (imp - 400f)`, so 5000 raw becomes 30 instead of 500 — roughly 5 percentage points of body fat.

  **4. The 2.5 s BIA window can lose a measurement.** A device that now counts as BIA-capable but broadcasts no impedance no longer publishes its first stable frame; it publishes only if another stable frame arrives >2.5 s later, and there's no timer — it's driven purely by further advertisements. On scan timeout `BroadcastScaleAdapter` calls `handleDisconnected()`, which nulls  `pendingMeasurement`, and the weight is dropped silently. That's a regression for devices that work today. I'd drop the window: you confirmed the ES-26M-W carries resistance in its settled frame, and `renphoMoving` already carries the same 5000 raw.

  **5. The captures.** `renphoSettledWeight` is byte-for-byte identical to `renphoSettledBia` apart from zeroed `[19-20]` — same sequence counter `0x0405` — so it's synthesized, not captured. Everything is also normalized to exactly 70.00 kg / 500.0 Ω, so the tests can't show whether the scaling is right.

  What would move this forward: a raw, unedited log of a full ES-26M-W session (moving frames, settled frame, idle/step-off), and a discriminator that actually separates the ES-26M-W from an FT-26R and from a bit-5 ES-CS20M — a model id or layout difference, not a byte that's constant across the family.

Quentin-Red-Turtle:
I own a Renpho ES-26M-W, one of the models this PR targets, and I captured it on real hardware. I think the findings settle the question, and unfortunately not in favour of the impedance decoding proposed here.

1. Bytes [19-20] are constant on this device.
I captured its advertisements two ways: from an app-level scanner (hundreds of frames, no deduplication, foreground service so no background throttling, `setLegacy(false)` + `PHY_LE_ALL_SUPPORTED` so extended advertising was not excluded), and from the Android Bluetooth stack itself via btsnoop, which cannot be filtered by `ScanRecord`.
Both agree: the device emits a single 31-byte advertisement, one AD structure (flags + manufacturer data, company ID `0xFFFF`). It never advertises as connectable (400+ observations), GATT connection attempts time out, and it is absent from a classic Bluetooth inquiry.
Bytes [19-20] hold `4C 15` in every frame — loaded, unloaded, barefoot, and through a garment.

2. Two completed measurements, two very different bodies, byte-identical frames.
I had two people with very different body mass each complete a full measurement cycle (status `0x23`). Comparing the stabilised frames, only five bytes differ: the weight and three counters. Everything else is identical, including `[12-14]`.

3. `[12-14] == FF FF FF` is not a BIA-capability signature.
It is a pairing/target field. In the official app it is read as `String.format("%02X%02X%02X", bArr[21], bArr[20], bArr[19])` and compared against the phone's address, with `"FFFFFF"` meaning "not bound to any phone". That is why it flips to a fixed value after a measurement, and why it is identical for different users — it identifies the phone, never the body.

4. The official app fabricates the impedance for this family.
I decompiled the Renpho app (`com.qingniu.renpho` 3.20.0). This scale is recognised as the `QS1` / `AABB` family (internal protocol id `124`), handled by `BroadcastQS1DecoderImpl`. Its frame parser contains:

```java
if (z11) {   // z11 = stable-measurement flag from the status byte
    baseBroadcastData.renphoOo0OoO0(new Random().nextInt(41) + 480);
}
```

That value is then passed as the third argument to the body-composition call — the same parameter that, for other scale families in this app, receives the resistance actually read from the frame (their decoder logs it as 阻抗, "impedance").
So the official app generates a random 480–520 Ω value for these scales. It would have no reason to do that if the frame carried a real one.

Conclusion: on this hardware, the `AABB` broadcast variant carries weight only. I would suggest not merging the impedance extraction for this family — it would produce body-composition values from bytes that never change.

Happy to share raw captures if useful.

iajmdc:
Thanks for the detailed review and for the additional investigation.

I did a few more captures on my ES-26M-W, including repeated attempts with normal contact, a sock/barrier, and wetted contact. Across the parsed AABB frames, bytes `[19-20]` stayed constant at `4C 15`, and byte `[21]` stayed constant at `03`; only the sequence/status/weight/trailing counter fields changed.

So I agree the current PR is not defensible: the proposed impedance decoding is very likely deriving body composition from a constant field rather than from measured/transmitted impedance. I’m going to close this PR rather than risk openScale storing misleading body-composition values.

Thanks again for catching this. If a future capture shows a separate protocol/frame carrying real impedance for this scale family, I can revisit it as a new, narrowly scoped PR.

PR #1523, opened 2026-09-13
feat: Renpho Elis Solar (R-A012/R-A020) body composition support

YoannDev90:
Body composition (body fat, water, skeletal muscle, bone mass, protein) for Renpho Elis Solar scales using the renpho-escs20m algorithm 0x04.

https://github.com/ronnnnnnnnnnnnn/renpho-escs20m/blob/master/src/renpho_escs20m/body_metrics.py


YoannDev90:
image

If needed you can add this on the wiki page

oliexdev:
Thanks for your PR. 
 
Could you show me what your scale actually advertises? Either an nRF Connect scan, or an openScale log file — but for the log it has to be from an **unpatched** build (release or master, without your handler change). The line to look for is `Modern handler '…' supports 'R-A012'`: if it says `ES-CS20M / ES-32MD`, then 0x1A10 is advertised; if it says `No suitable communicator
  found`, it isn't. The `Saving device snapshot: … uuids=N` line also tells me the count.
  
  This matters for the `ScaleFactory` reorder: `ESCS20MHandler` claims by service UUID, and the factory takes the first handler that matches. If the UUID is advertised, the ES-26BB-B is already being hijacked today, since `ESCS20MHandler` sits ahead of `RenphoES26BBHandler` on master. Either way, please pass the UUID into the catalog fixture (`device("R-A012", uuid16(0x1A10))`) —
  with `serviceUuids` empty, both orderings pass the test, so it doesn't lock anything in.
  
  A few other things:

  - **R-A020**: upstream lists ES-CS20MB1, ES-26BB-B, R-A012 and R-A016 under the 0x55aa basic flavor — R-A020 isn't mentioned anywhere, and R-A016 isn't added here. Upstream also warns that the marketed model name is unreliable and recommends checking the HVIN, which is exactly what `supportFor()` matches on. Which model did you actually test on? If R-A020 is unverified, I'd
  rather drop it than claim it by name.
  - **Body fat test**: `body fat formula matches renpho-escs20m algorithm 0x04` re-evaluates the same formula with the same coefficients, so it only checks the arithmetic. Since the impedance term is just `-500/R` (about one percentage point across the plausible range), the value is essentially a BMI+age regression. One real comparison — the Renpho app's reading next to
  openScale's for the same weigh-in — would be worth more than both formula tests.
  - **Doc vs code**: the new comment says "status low nibble: 0 settling, 1 final", but `parseLiveFrame` still checks `type == 0x01 || type == 0x11`. Please mask the nibble or drop that wording.
  - Minor: `bodyFatPercent` takes a `Boolean` for sex while the other four take a `Float`; missing newline at the end of the new test file. The commit message says "Replace TrisaBodyAnalyzeLib", but this handler never used it.

YoannDev90:
Here is the unpatched release build log (3.1.3, `7ebb55be`), cleaned up to remove duplicate EtekcityFit8SHandler lines:

```
14:01:29.335 BluetoothScannerMgr: Starting device scan for 20000 ms.
14:01:36.454 GattScaleAdapter: Found CF:E9:35:14:20:F7 → stop scan + connect
14:01:38.406 GattScaleAdapter: Services discovered for CF:E9:35:14:20:F7
14:01:38.407 DebugGattHandler: attach()
14:01:38.407 DebugGattHandler: handleConnected(userId=1, height=183.0, age=18)
14:01:38.407 DebugGattHandler: Connected in Debug mode. Dumping full GATT services/characteristics…
14:01:38.408 DebugGattHandler: === GATT Service Dump BEGIN ===
14:01:38.408 DebugGattHandler: Service 0x1800
14:01:38.408 DebugGattHandler:   └─ Char 0x2a00 props=READ
14:01:38.409 DebugGattHandler: Service 0x1a10
14:01:38.409 DebugGattHandler:   └─ Char 0x2a11 props=WRITE
14:01:38.409 DebugGattHandler:   └─ Char 0x2a10 props=NOTIFY
14:01:38.409 DebugGattHandler: Service 0x180a
14:01:38.409 DebugGattHandler:   └─ Char 0x2a29 props=READ
14:01:38.410 DebugGattHandler:   └─ Char 0x2a24 props=READ
14:01:38.410 DebugGattHandler:   └─ Char 0x2a25 props=READ
14:01:38.410 DebugGattHandler:   └─ Char 0x2a27 props=READ
14:01:38.411 DebugGattHandler:   └─ Char 0x2a28 props=READ
14:01:38.411 DebugGattHandler:   └─ Char 0x2a26 props=READ
14:01:38.411 DebugGattHandler:   └─ Char 0x2a23 props=READ
14:01:38.411 DebugGattHandler:   └─ Char 0x2a50 props=READ
14:01:38.411 DebugGattHandler: Service 0x180f
14:01:38.412 DebugGattHandler:   └─ Char 0x2a19 props=READ|NOTIFY
14:01:38.412 DebugGattHandler: Service 0xfe59
14:01:38.412 DebugGattHandler:   └─ Char 8ec90001-f315-4f60-9fb8-838830daea50 props=WRITE|NOTIFY
14:01:38.412 DebugGattHandler:   └─ Char 8ec90002-f315-4f60-9fb8-838830daea50 props=WRITE_NR
14:01:38.413 DebugGattHandler: === GATT Service Dump END ===
14:01:38.452 DebugGattHandler: NOTIFY chr=0x2a00  payload=[52 2D 41 30 32 30]  ascii=R-A020
14:01:38.584 DebugGattHandler: NOTIFY chr=0x2a29  payload=[4C 65 46 75 20 53 63 61 6C 65]  ascii=LeFu Scale
14:01:38.682 DebugGattHandler: NOTIFY chr=0x2a24  payload=[33 2E 39 34 38 56 2C 30 2E 39 38 37 56 2C 31 30 30 25]  ascii=3.948V,0.987V,100%
14:01:38.765 DebugGattHandler: NOTIFY chr=0x2a26  payload=[32 31 30 37 31 32]  ascii=210712
14:01:38.863 DebugGattHandler: NOTIFY chr=0x2a28  payload=[76 32 2E 37]  ascii=v2.7
14:01:38.967 DebugGattHandler: NOTIFY chr=0x2a19  payload=[64]  ascii=d
14:01:45.809 UserSettingsRepository: Bluetooth developer mode disabled.
14:01:47.400 GattScaleAdapter: Disconnected CF:E9:35:14:20:F7: CONNECTION_TERMINATED_BY_LOCAL_HOST
14:01:47.402 DebugGattHandler: handleDisconnected()
14:01:47.402 DebugGattHandler: detach()
14:01:47.734 ScaleHandlerFactory: createCommunicator: Searching for communicator for 'R-A020' (CF:E9:35:14:20:F7). Handler hint: 'null'
14:01:47.737 ScaleHandlerFactory: No suitable communicator found for device (name: 'R-A020', address: 'CF:E9:35:14:20:F7', handler hint: 'null').
14:01:47.770 BluetoothScannerMgr: Stopping device scan. Triggered by timeout: false
14:02:13.133 BluetoothFacade: Saved device 'R-A020' is not recognized as supported anymore.
14:02:13.134 BleConnector: Device R-A020 is NOT supported according to ScannedInfo. Connection ABORTED.
```

Key findings:
- R-A020 **does** advertise service `0x1a10` (confirmed by GATT dump)
- Manufacturer is **"LeFu Scale"** (Renpho protocol family)
- On the unpatched release build, `ScaleHandlerFactory` finds **no suitable communicator** for R-A020
- `ESCS20MHandler` does **not** claim R-A020 despite the UUID match (its `supportFor()` doesn't match the name)

YoannDev90:
> R-A020: upstream lists ES-CS20MB1, ES-26BB-B, R-A012 and R-A016 under the 0x55aa basic flavor — R-A020 isn't mentioned anywhere, and R-A016 isn't added here. Upstream also warns that the marketed model name is unreliable and recommends checking the HVIN, which is exactly what supportFor() matches on. Which model did you actually test on? If R-A020 is unverified, I'd rather drop it than claim it by name.

Actually, I only have the R-A020

YoannDev90:
> Body fat test: body fat formula matches renpho-escs20m algorithm 0x04 re-evaluates the same formula with the same coefficients, so it only checks the arithmetic. Since the impedance term is just -500/R (about one percentage point across the plausible range), the value is essentially a BMI+age regression. One real comparison — the Renpho app's reading next to openScale's for the same weigh-in — would be worth more than both formula tests.

Here's what we tried before settling on this formula:

1. **TrisaBodyAnalyzeLib** (same approach as ESCS20MHandler for ES-CS20M): gave ~2× the official app's body fat on the same weigh-in (e.g. ~10% vs ~5%).
2. **Decompiled the official Renpho APK** (v4.3.0): the body fat calculation lives in `libICBodyFatAlgorithms.so` (native C/C++), which can't be decompiled from Java. The Java interface (`ICBodyFatAlgorithmsParams`) takes age, height, sex, weight, 5 impedance values and `algType` — but our scale only provides 1 impedance (4-electrode), so the 5-impedance path doesn't apply.
3. **renpho-escs20m** ([body_metrics.py](https://github.com/ronnnnnnnnnnnnn/renpho-escs20m/blob/master/src/renpho_escs20m/body_metrics.py)): community-reverse-engineered the exact algorithm 0x04 (non-athlete) as a BMI+age+impedance linear regression. The README confirms it's validated against the official app for 0x55aa basic-flavor scales (R-A012, ES-26BB-B, etc.) — [see compatibility table](https://github.com/ronnnnnnnnnnnnn/renpho-escs20m#device-compatibility).

I'd rather not share personal health measurements in a public issue, but tests on this scale showed <1% difference between the official app's body fat reading and the value computed with this algorithm.

oliexdev:
Thanks for the log — but I read it the other way round.

  The GATT dump is taken *after* connecting and says nothing about what's in the advertisement.  ScannedDeviceInfo.serviceUuids` is filled from `scanRecord.serviceUuids` (`BleScanner.kt:253`), and  ESCS20MHandler.supportFor()` claims on `hasSvc || nameMatch`. So if the R-A020 advertised 0x1A10, master would have handed it to `ESCS20MHandler` — instead the log says "No suitable communicator found". That means 0x1A10 is **not** advertised, only present on the device once connected.

  Two things follow:

  - `device("R-A020", uuid16(0x1A10))` in the catalog asserts something your own log contradicts. Please drop the UUID from both fixtures.
  - With no advertised UUID, the `ScaleFactory` reorder is a no-op for these scales — `RenphoES26BBHandler` only matches exact names, so ordering never comes into play. It would only matter for the ES-26BB-B, and nobody has verified whether that one advertises 0x1A10. I'd rather leave `ScaleFactory` alone in this PR than reorder on a guess. (Note the two are coupled right now:  with the UUID in the fixture the catalog test *requires* the reorder, so both unverified claims prop each other up.)

Last one, minor: dropping `BATTERY_LEVEL` and renaming `battery` to `flag` changes a shipping device you can't test. I'm fine with it if upstream's `protocol.py` really has nothing at that offset — just confirm that's where it comes from.

PR #1526, opened 2026-09-17
fix(bluetooth): accept Hagrid auth challenge with trailing status byte

stpinkie:
## Summary

Huawei Scale 3 on fw 2.0.0.70 sends the auth challenge as `randA[16] || status[1]`
(17 bytes). `handleAuthChallenge` required exactly 16 bytes, dropped the
challenge, and never sent the auth token — the scale disconnected ~25 s later.

- accept challenges of ≥16 bytes; use `copyOfRange(0, 16)` as randA (the trailing
  byte is a status field, not part of the HMAC input)
- `isValidAuthResponsePayload` gets the same prefix tolerance in case other
  firmwares append a trailer to the response (defensive — only the challenge
  trailer was observed)

Verified on-device: a HUAWEI Scale 3 (fw 2.0.0.70) completes auth (token
verified, work key accepted, READY) and publishes measurements with body
composition. Same driver family as #1492.

## Evidence (sanitized)

    → write  request-auth  [DB 03 00 C1 40]
    ← notify request-auth  [BD 12 10 …15B…] [BD 05 11 2F 00 04 E2]
    HuaweiHagridWspHandler: Ignoring invalid Hagrid auth challenge length=17
    (scale disconnects ~25 s later: REMOTE_USER_TERMINATED_CONNECTION)

The randA interpretation was verified cryptographically against a captured
vendor-app session: HMAC(randA‖randB, HMAC(randA‖randB, CAK‖"1123")) over the
first 16 challenge bytes reproduces both captured auth tokens exactly.

## Test plan

- [x] Regression test replays a 17-byte challenge through the real WSP frame
      reassembly path and asserts the token uses randA[0:16]
- [x] Lib test: exact-length, +1 trailer, and truncated response cases
- [x] CI green (fork): testDebugUnitTest, compileDebugAndroidTestKotlin,
      assembleDebug — https://github.com/stpinkie/openScale/actions/runs/35119274310
- [x] Real hardware weigh-in on HUAWEI Scale 3 / fw 2.0.0.70 via dev-build APK

## Known observation (not addressed here)

The scale answered set-user-info with status=1 (profile rejected) while still
streaming realtime data. Likely related to the huid field — the bound value
differs from the synthesized `openscale:<mac>` fallback. Worth a separate look.


oliexdev:
Thanks for your PR :+1:

PR #1530, opened 2026-09-19
feat: add measurement comparison feature

Sayandevxyz:
## Summary

Adds a **Measurement Comparison** feature that lets users compare two body-composition measurements side-by-side, showing absolute delta, percentage change, and weekly rate of change per metric.

## Entry Points

- **Table screen**: Select exactly 2 raw (non-aggregated) rows and a compare icon appears in the contextual top bar
- **Measurement Detail screen**: A compare icon in the top bar compares the current entry vs the user's previous measurement

## Changes

| File | Type | Description |
|---|---|---|
| `core/model/MeasurementComparison.kt` | New | Domain model holding delta, percentage, and weekly rate |
| `core/usecase/MeasurementComparisonUseCase.kt` | New | Pure Kotlin business logic -- zero external dependencies |
| `ui/screen/dialog/MeasurementComparisonDialog.kt` | New | Material 3 dialog with swap direction, trend icons, weekly rate |
| `core/usecase/MeasurementComparisonUseCaseTest.kt` | New | 5 JVM unit tests covering all edge cases |
| `ui/screen/table/TableScreen.kt` | Modified | Comparison entry point in selection mode (2 raw rows selected) |
| `ui/screen/overview/MeasurementDetailScreen.kt` | Modified | Comparison entry point in detail view |
| `res/values/strings.xml` | Modified | 14 new localized strings |

## Design Decisions

- **Zero new dependencies** -- uses only existing Kotlin stdlib, Compose M3, and Hilt APIs already in the project
- **Privacy-first** -- 100% local computation, no data leaves the device
- **No breaking changes** -- purely additive; all existing screens and flows are unchanged
- **Follows project conventions** -- `@Singleton @Inject` Hilt pattern, `LaunchedEffect` VM wiring, `LocaleUtils.formatNumber` for all numeric output

## Testing

- 5 JVM unit tests in `MeasurementComparisonUseCaseTest` covering:
  - Basic positive/negative delta
  - Zero baseline (avoids division by zero)
  - Weekly rate calculation
  - Missing values (skipped metrics)
  - Swap direction


oliexdev:
Thanks for this — the idea is a good one and it is now on master, though not as this patch. I ended up building it on top of what the app already has rather than
  merging, so let me be transparent about why.

  Three things in the PR needed rework beyond polish:

  * `MeasurementComparisonUseCase()` is constructed directly inside two composables, so the `@Singleton @Inject` never applies and a new instance is made per tap.
  Screens here go through a ViewModel and a facade; every other use case hangs off `MeasurementFacade`.
  * "Compare with previous" uses `lastMeasurementOfSelectedUser`, which is the user's *newest* entry — the flow that preloads a new measurement's fields. Open an
  older entry and it compares against the latest one while labelling it "previous".
  * The swap button rebuilds its numbers by trimming signs off already formatted strings. That breaks for the stone unit ("+3 st 5 lb") and for any locale that
  places the sign differently.

  Smaller ones: `"/wk"`, `" %"` and `"Metric"` are built in code rather than `strings.xml`, so they never reach Weblate, and `TrendCalculator` already exists as an
  injectable service with an epsilon.
 
  What landed instead needs no new use case and no new model: `SharedViewModel.drillDownFlow` gained an aggregation level and loads both entries through the pipeline
  the table drill-down already uses, and `calculateStatisticsForType` turns each metric into its two values and their difference. The screen is a table — one column
  per side plus the change — and it works for aggregated rows too, so you can compare one month's mean against another's.

  Your domain layer was the strongest part of the PR, and its shape carried over. Closing this in favour of master; thanks again for pushing the feature.

Sayandevxyz:
Thanks for the detailed feedback and for explaining the architectural changes. I’m glad the domain-layer approach was useful and that the feature made it into master.

I’ll take a look at the new implementation and the surrounding architecture so I can align future contributions with the existing ViewModel/facade pipeline. If there’s any follow-up improvement or testing work around this feature that would be useful, I’d be happy to contribute.

PR #1535, opened 2026-09-21
Runstar R6: derive body composition from the reported impedance

ldoench:
The Runstar R6 reports whole-body impedance in its 0xA3 result frame, but
the handler added in #1432 only stored the raw value. Since the IMPEDANCE
measurement type is internal and disabled, and DerivedValuesCalculator
does not read it, users saw weight, BMI and heart rate and nothing else.

This feeds the impedance into StandardImpedanceLib to derive body fat,
water, skeletal muscle, bone mass, LBM and BMR, following the pattern
VitafitVT701Handler and EtekcityESF551Handler already use for their
weight-plus-impedance-only scales.

Measured impedance on a physical R6 is 533-605 Ohm across eight weigh-ins,
within the range StandardImpedanceLib documents its formulas for. Readings
outside 200-1200 Ohm, and users with no body height set, are skipped with
a warning rather than producing made-up values; the raw impedance is
stored regardless.

Also fixes two robustness gaps in the same handler:

- 0xA4 history timestamps were used unchecked, so an unset scale clock
  produced measurements dated 1970. Now guarded like AfuB1Handler and
  YunmaiHandler, falling back to the current time rather than dropping
  the entry — acking retires it on the scale, so a rejected reading
  cannot be recovered.
- The documented 0xA3 status byte was never read. A non-zero status is
  now logged rather than used to discard the weigh-in; its values have
  never been observed in any capture.

Verification:

- New RunstarR6HandlerTest covers the 0xA3 decoder and the checksum
  against hand-built frames, including malformed, wrong-type,
  wrong-length and non-zero-status input.
- One test decodes a frame captured verbatim from a physical R6
  (`F3 08 00 A3 00 01 03 2E 4B 02 45 ...` -> 66.35 kg, 75 bpm, 581 Ohm).
- Verified end-to-end on the device: body composition now appears in the
  app after a weigh-in.


oliexdev:
Thanks for your PR :+1:

